PR #1127, opened 2023-09-01
Nft and qr modals

isstuev:


ui/address/details/__screenshots__/AddressQrCode.pw.tsx_dark-color-mode_default-view-mobile-dark-mode-1.png
tom2drum, line 1: pull the latest changes from the main branch, so there should be a smaller icon for address and a new icon for contract (separate for unverified and verified)

ui/shared/nft/NftMedia.tsx
```diff
@@ -9,24 +9,26 @@ import useFetch from 'lib/hooks/useFetch';
 
 import NftFallback from './NftFallback';
 import NftHtml from './NftHtml';
+import NftHtmlWithFullscreen from './NftHtmlWithFullscreen';
 import NftImage from './NftImage';
+import NftImageWithFullscreen from './NftImageWithFullscreen';
 import NftVideo from './NftVideo';
+import NftVideoWithFullscreen from './NftVideoWithFullscreen';
```
tom2drum, line 16: Do we really need 3 separate components for this? I think it is better to configure the NftMedia that is in the fullscreen modal by simply passing the necessary styles to it. Now, we have a lot of duplication in media components.

tom2drum (changes requested):
1. [image](https://review-nft-and-qr-modals.k8s-dev.blockscout.com/token/0xf68F81069AD049a11006bbffEa06D8dE664A5522/instance/200) in Safari in modal looks skewed

<details><summary>Screenshot</summary>
<p>
image
</p>
</details>

ui/shared/nft/NftMedia.tsx
isstuev: we need. you know I'm not a fan of sophisticated abstractions. I can move common props to utils, but not a unite component

tom2drum: I assume that passing style props to already existing components is not something sophisticated.

What I mean is apart from the "content" variable here, we can have a "modal" variable that will render the same media components but with custom style props:

```js
const modal = (() => {
    switch (type) {
      case 'video':
        return (
          <NftMediaFullscreenModal>
            <NftVideo { ...styleProps }/>
          </NftMediaFullscreenModal>
        );
     etc...
    }
})();
```

It is just reusing the existing components with different props. I don't get why it is over-complicated.

tom2drum (approved):

ui/shared/nft/NftMedia.tsx
isstuev: I don't see what would be better in that case, honestly.

isstuev: it doesn't work for the html type, fullscreen and small has different structures, so it's not "the same media component"

PR #1144, opened 2023-09-05
Alternative websocket API PORT error

ekawibisono:
This is a screenshot when Error adding ENV API PORT
image

This is an alternative if you use the HOST IP Address API, then the PORT must be added to the API code too and no errors occur on the websocket

image

### DIFF
--- a/configs/app/api.ts
+++ b/configs/app/api.ts
@@ -16,7 +16,7 @@ const socketSchema = getEnvValue(process.env.NEXT_PUBLIC_API_WEBSOCKET_PROTOCOL)
 const api = Object.freeze({
   host: apiHost,
   endpoint: apiEndpoint,
-  socket: `${ socketSchema }://${ apiHost }`,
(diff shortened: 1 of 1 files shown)


configs/app/api.ts
```diff
@@ -16,7 +16,7 @@ const socketSchema = getEnvValue(process.env.NEXT_PUBLIC_API_WEBSOCKET_PROTOCOL)
 const api = Object.freeze({
   host: apiHost,
   endpoint: apiEndpoint,
-  socket: `${ socketSchema }://${ apiHost }`,
+  socket: `${ socketSchema }://${ apiHost }:${ apiPort}`,
```
isstuev, line 19: what if there is no api port variable?

isstuev (changes requested):

isstuev:
I'm closing this PR because 
https://github.com/blockscout/frontend/issues/1158

Issue #1137, opened 2023-09-05
ERC-xxx token balances should have the same level of rounding from API as for native coin


vbaranov:
### Description

Especially, sensitive for rollups with native coin balance = bound erc-20 token balance

![Screenshot 2023-09-05 at 12 28 15Z](https://github.com/blockscout/frontend/assets/4341812/8fb364e5-4cd4-4d04-a385-8f14f5d8b9bd)

### Link to the page

https://eth.blockscout.com

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### App version

v1.10.0

### Operating system

None

PR #1145, opened 2023-09-06
fix tokens rounding

isstuev:
resolves https://github.com/blockscout/frontend/issues/1137


ui/address/tokenSelect/TokenSelectItem.tsx
```diff
@@ -19,7 +19,7 @@ const TokenSelectItem = ({ data }: Props) => {
     switch (data.token.type) {
       case 'ERC-20': {
         const tokenDecimals = Number(data.token.decimals) || 18;
-        const text = `${ BigNumber(data.value).dividedBy(10 ** tokenDecimals).toFormat(2) } ${ data.token.symbol || '' }`;
+        const text = `${ BigNumber(data.value).dividedBy(10 ** tokenDecimals).dp(8).toFormat() } ${ data.token.symbol || '' }`;
```
tom2drum, line 22: probably, the test suite should be modified as well to show the case with 8+ digits after the comma

tom2drum (approved):

ui/address/tokenSelect/TokenSelectItem.tsx
tom2drum: I think it is more about we chose the correct way to display these values and it doesn't affect anything in the layout

Issue #1117, opened 2023-08-25
tx revert reason can be utf-8, not hex


isstuev:
https://puppyscan.shib.io/tx/0xde9bea7f595dab90bc83da62ff3ff9c6851d855e1b6276938b32c197c96ec8e2?tab=index

![Image](https://user-images.githubusercontent.com/108066576/263308646-8f8663ea-da25-42c8-8ca8-f0facb0901ab.png)

PR #1148, opened 2023-09-06
support no-hex revert reason

isstuev:
resolves https://github.com/blockscout/frontend/issues/1117


ui/tx/details/TxRevertReason.tsx
```diff
@@ -12,7 +12,14 @@ const TxRevertReason = (props: Props) => {
   const bgColor = useColorModeValue('blackAlpha.50', 'whiteAlpha.50');
 
   if ('raw' in props) {
+    const hexRegexp = /^(?:0x)?[\da-fA-F]+$/;
```
tom2drum, line 15: maybe it is better to store this RegExp here `lib/regexp.ts` in order to reuse it later in other parts of the code base

tom2drum (approved):

PR #1161, opened 2023-09-09
Add 'false' as acceptable parameter for the blockchain parameters: NEXT_PUBLIC_IS_TESTNET

jasonzysun:


jasonzysun:
Some users (such as me) may think that when I deploy a main web browser, I can set `NEXT_PUBLIC_IS_TestNET` to `'false'` (corresponding to true), so I believe that in addition to `'true'` and `''`, `'false'` should also be accepted.

jasonzysun:
`NEXT_PUBLIC_IS_TestNET=false` will cause error when the frontend start, which is unreasonable.

jasonzysun:
Well, that's OK.

Issue #1162, opened 2023-09-10
Set NEXT_PUBLIC_API_SPEC_URL default


ulyanas:
Set the default value of `NEXT_PUBLIC_API_SPEC_URL` to https://raw.githubusercontent.com/blockscout/blockscout-api-v2-swagger/main/swagger.yaml

The reason being is that API is on by default, but the docs are missing that

PR #1166, opened 2023-09-11
default spec url

isstuev:
resolves https://github.com/blockscout/frontend/issues/1162


tom2drum (approved):
So, there is no way to disable this feature now.. I am not sure is it ok or not.

isstuev:
> So, there is no way to disable this feature now.. I am not sure is it ok or not.

I asked Nikita, he said it was the point

PR #1184, opened 2023-09-14
validate ENV values at run-time

tom2drum:
resolves #1156

PR #1202, opened 2023-09-21
New feature: Add custom ad banner

jasonzysun:
In addition to the three methods, adbutler, coinzilla and slise, provided in the adBanner, a custom adBanner method has been added. The method of obtaining configuration files refers to NEXT_PUBLIC_MarkeTPLACE_CONFIG_URL and advertising display method refer to slise.
I did not use the two testing methods mentioned in the Contribution guide, but I have applied this configuration in Qitmeer's block browser and can access [qng.qitmeer.io](https://qng.qitmeer.io) for viewing.


docs/ENVS.md
```diff
@@ -228,9 +228,19 @@ This feature is **enabled by default** with the `slise` ads provider. To switch
...
+| NEXT_PUBLIC_AD_BANNER_PROVIDER | `slise` \| `adbutler` \| `coinzilla` \| `custom` \| `none` | Ads provider  | - | `slise` | `coinzilla` |
 | NEXT_PUBLIC_AD_ADBUTLER_CONFIG_DESKTOP | `{ id: string; width: string; height: string }` | Placement config for desktop Adbutler banner | - | - | `{'id':'123456','width':'728','height':'90'}` |
 | NEXT_PUBLIC_AD_ADBUTLER_CONFIG_MOBILE | `{ id: string; width: number; height: number }` | Placement config for mobile Adbutler banner | - | - | `{'id':'654321','width':'300','height':'100'}` |
+| NEXT_PUBLIC_AD_CUSTOM_CONFIG_URL | `string` | URL of configuration file (.json format only) which contains list of custom banners that will be shown in the network menu. See below list of available properties for particular banner | - | - | `https://example.com/ad_custom_config.json` |
+
+#### Custom banners configuration properties
+
+| Variable | Type | Description | Compulsoriness  | Default value | Example value |
+| --- | --- | --- | --- | --- | --- |
+| text | `string` | Tooltip text displayed when the mouse is moved over the banner. | Required | - | `Anyblock` |
```
tom2drum, line 240: From my point of view, it should not be required. Some ad banners can exist without any tooltip, I think.

docs/ENVS.md
```diff
@@ -228,9 +228,19 @@ This feature is **enabled by default** with the `slise` ads provider. To switch
...
+| url | `string` | Link that opens when clicking on the banner. | Required | - | `https://example.com` |
+| desktop | `string` | Banner image (both .png, .jpg, and .gif are acceptable) used when the screen width is greater than 62em (usually 99px). | Required | - | `https://example.com/configs/ad-custom-banners/desktop/example.gif` |
```
tom2drum, line 242: > used when the screen width is greater than 62em (usually 99px)

This is not true. We use the `1000px` value as a breaking point between desktop / mobile view. See `theme/foundations/breakpoints.ts`

nextjs/nextjs-routes.d.ts
```diff
@@ -21,8 +21,8 @@ declare module "nextjs-routes" {
     | StaticRoute<"/api/media-type">
     | StaticRoute<"/api/proxy">
     | StaticRoute<"/api-docs">
-    | DynamicRoute<"/apps/[id]", { "id": string }>
     | StaticRoute<"/apps">
+    | DynamicRoute<"/apps/[id]", { "id": string }>
```
tom2drum, line 25: Is it an accidental change?

types/client/adCustomConfig.ts
```diff
@@ -0,0 +1,6 @@
+export type AdCustomConfig = {
```
tom2drum, line 1: Well, now I think it is better to move all ad types in one file - `types/client/ad`. I will appreciate if you do that.

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+const CustomAdBanner = ({ className }: { className?: string }) => {
+  const customBanners = useCustomBanners().data || [];
+  const [ currentBannerIndex, setCurrentBannerIndex ] = useState(0);
+  useEffect(() => {
+    if (customBanners.length === 0) {
+      return;
+    }
+    const timer = setInterval(() => {
+      setCurrentBannerIndex((prevIndex) => (prevIndex + 1) % customBanners.length);
+    }, 60000);
```
tom2drum, line 15: I would suggest putting it in the config as well, so other users can change this interval.

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,48 @@
+import { Flex, chakra, Tooltip, Image } from '@chakra-ui/react';
+import React, { useState, useEffect } from 'react';
+
+import useCustomBanners from 'ui/shared/ad/useCustomBanners';
+
+const CustomAdBanner = ({ className }: { className?: string }) => {
+  const customBanners = useCustomBanners().data || [];
```
tom2drum, line 7: As a side note, we don't need to prepend variables with `custom`-word, since this code is already inside the `CustomAdBanner.tsx` file. So, it is a little bit wordy now.

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  const currentBanner = customBanners[currentBannerIndex];
+
+  return (
+    <>
+      <Flex className={ className } h="90px" display={{ base: 'none', lg: 'flex' }}>
+        <Tooltip label={ currentBanner.text } aria-label={ currentBanner.text }>
+          <a href={ currentBanner.url } target="_blank" rel="noopener noreferrer">
+            <Image src={ currentBanner.desktop } alt={ currentBanner.text } height="100%" width="auto" borderRadius="10px"/>
```
tom2drum, line 33: let's use one of our standard values for `border-radius` here - `theme/foundations/borders.ts`

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+          </a>
+        </Tooltip>
+      </Flex>
+      <Flex className={ className } h="90px" display={{ base: 'flex', lg: 'none' }}>
+        <Tooltip label={ currentBanner.text } aria-label={ currentBanner.text }>
+          <a href={ currentBanner.url } target="_blank" rel="noopener noreferrer">
+            <Image src={ currentBanner.mobile } alt={ currentBanner.text } height="100%" width="auto" borderRadius="10px"/>
+          </a>
+        </Tooltip>
+      </Flex>
```
tom2drum, line 43: I think there is no need to render both desktop and mobile banners simultaneously. Can it be just an if-statement for the image source?

image

ui/shared/ad/useCustomBanners.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const apiFetch = useApiFetch();
+  const { isLoading, data } = useQuery<unknown, ResourceError<unknown>, Array<AdCustomConfig>>(
+    [ 'custom-configs' ],
+    async() => apiFetch(configUrl),
+    {
+      enabled: feature.isEnabled && feature.provider === 'custom',
+      staleTime: Infinity,
+    });
+
+  return React.useMemo(() => ({
```
tom2drum, line 22: I assume that result from `'@tanstack/react-query'` is already memoized. So, this is unnecessary. You can just return `useQuery` call result from here. And probably this hook doesn't deserve to be in a separate file, imho.

tom2drum (changes requested):
Can you also add the variable of the config URL to the list of downloaded assets, here - `deploy/scripts/download_assets.sh`? Then use helper `getExternalAssetFilePath` in the app config file to get the public path to the file in application code.

docs/ENVS.md
jasonzysun: Yes, I think you're write and fix it.

docs/ENVS.md
jasonzysun: I'll fix this.

nextjs/nextjs-routes.d.ts
jasonzysun: This seems to be just a sequence adjustment, automatically completed by the system, and not something I can operate. Do I need to withdraw this modification?

types/client/adCustomConfig.ts
jasonzysun: Ok, I'll do that.

ui/shared/ad/CustomAdBanner.tsx
jasonzysun: Ok, let me change this.

ui/shared/ad/CustomAdBanner.tsx
jasonzysun: Ok, I'll rename it.

ui/shared/ad/CustomAdBanner.tsx
jasonzysun: Can you tell me how to use this? How to replace borderRadius="10px"?

ui/shared/ad/CustomAdBanner.tsx
jasonzysun: I'll try to fix this.

ui/shared/ad/useCustomBanners.tsx
jasonzysun: Okay, I will optimize this section.

jasonzysun:
OK, I'll add the variable of the config URL to the list of downloaded assets, and use helper getExternalAssetFilePath in the app config file to get the public path to the file in application code. I've done that at first commit but remove it at second because at that time I wanted this part of the advertisement to be automatically adjusted with modifications to external files, rather than fixed at startup. Now I think we can add a separate environment variable to control the use of this feature.

jasonzysun:
I'll add another env variables like NEXT_PUBLIC_DISABLE_DOWNLOAD_AT_RUM_TIME to realize that.

ui/shared/ad/CustomAdBanner.tsx
jasonzysun: Well, I've found it.
  return (
    <Flex className={ className } h="90px">
      <Tooltip label={ currentBanner.text } aria-label={ currentBanner.text }>
        <a href={ currentBanner.url } target="_blank" rel="noopener noreferrer">
          <Image src={ isMobile ? currentBanner.mobile : currentBanner.desktop } alt={ currentBanner.text } height="100%" width="auto" borderRadius="md"/>
        </a>
      </Tooltip>
    </Flex>
  );

jasonzysun:
@tom2drum Please check again.

nextjs/nextjs-routes.d.ts
tom2drum: I would say yes, since if I run dev build on my machine, I don't see any changes in this file. Weird that it was changed for you.

deploy/tools/envs-validator/schema.ts
```diff
@@ -103,12 +102,40 @@ const adButlerConfigSchema = yup
       .required(),
   });
 
+const adCustomConfigSchema = yup
+  .object()
+  .shape({
+    banners: yup
+      .array()
+      .of(
+        yup.object().shape({
```
tom2drum, line 111: for type safety, it is better to put this schema in a separate variable and specify the type for it as `yup.ObjectSchema<AdCustomBannerConfig>` or add the type annotation for the whole `adCustomConfigSchema`

see `marketplaceAppSchema` as an example

deploy/tools/envs-validator/schema.ts
```diff
@@ -103,12 +102,40 @@ const adButlerConfigSchema = yup
...
+          text: yup.string(),
+          url: yup.string().url().required(),
+          desktop: yup.string().url().required(),
+          mobile: yup.string().url().required(),
+        }),
+      )
+      .required(),
+    interval: yup.number(),
```
tom2drum, line 119: I would add that it should be a positive integer as well

docs/ENVS.md
```diff
@@ -228,9 +228,27 @@ This feature is **enabled by default** with the `slise` ads provider. To switch
...
+&nbsp;
+
+#### Custom banners configuration properties
+
+| Variable | Type | Description | Compulsoriness | Default value | Example value |
+| --- | --- | --- | --- | --- | --- |
+| text | `string` | Tooltip text displayed when the mouse is moved over the banner. | - | - | - |
+| url | `string` | Link that opens when clicking on the banner. | Required | - | `https://example.com` |
+| desktop | `string` | Banner image (both .png, .jpg, and .gif are acceptable) used when the screen width is greater than 1000px. | Required | - | `https://example.com/configs/ad-custom-banners/desktop/example.gif` |
+| mobile | `string` | Banner image (both .png, .jpg, and .gif are acceptable) used when the screen width is less than 1000px. | Required | - | `https://example.com/configs/ad-custom-banners/mobile/example.gif` |
```
tom2drum, line 251: I am not sure, but probably from the end-user perspective it will be more clear if these properties are more self-decriptive, like `desktop_image_url` and `mobile_image_url`. Not really necessary though.

What do you think?

types/client/ad.ts
tom2drum, line 1: great! thanks 👍

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { useQuery } from '@tanstack/react-query';
+import React, { useState, useEffect } from 'react';
+
+import type { AdCustomConfig } from 'types/client/ad';
+
+import config from 'configs/app';
+import type { ResourceError } from 'lib/api/resources';
+import useFetch from 'lib/hooks/useFetch';
+
+import useIsMobile from '../../../lib/hooks/useIsMobile';
```
tom2drum, line 11: better to use aliased import - `lib/hooks/useIsMobile`

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const CustomAdBanner = ({ className }: { className?: string }) => {
+  const isMobile = useIsMobile();
+
+  const feature = config.features.adsBanner;
+  const configUrl = (feature.isEnabled && feature.provider === 'custom') ? feature.configUrl : '';
+
+  const apiFetch = useFetch();
+  const { data: adConfig } = useQuery<unknown, ResourceError<unknown>, AdCustomConfig>(
+    [ 'custom-configs' ],
```
tom2drum, line 21: This key is a bit generic now. Let it be more specific like `ad-banner-custom-config` for example

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,61 @@
...
+    async() => apiFetch(configUrl),
+    {
+      enabled: feature.isEnabled && feature.provider === 'custom',
+      staleTime: Infinity,
+    });
+  const interval = adConfig?.interval || 60000;
```
tom2drum, line 27:
```js
import { MINUTE } from 'lib/consts';
const interval = adConfig?.interval || MINUTE;
```

just for the sake of code readability

tom2drum (approved):
Great work! Leave some comments about the code style and types. Feel free to fix them.

tom2drum:
> I'll add another env variables like NEXT_PUBLIC_DISABLE_DOWNLOAD_AT_RUM_TIME to realize that in other pr.

I don't really think that we need this kind of setting. In general, the app should not break if the remote config is suddenly changed. Configuration should be persistent for the running instance. That's why we download all these files at run-time. If you need another configuration for your app, then re-run the container with new ENVs and you're good to go. It should not be any different than for other (non-json like) ENVs.

nextjs/nextjs-routes.d.ts
jasonzysun: The beginning of the file reads:
// THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
// This file will be automatically regenerated when your Next.js server is running.

Are you sure I need to manually change it back?

deploy/tools/envs-validator/schema.ts
jasonzysun: Okay

deploy/tools/envs-validator/schema.ts
jasonzysun: Yes, you're right

docs/ENVS.md
jasonzysun: Well, good idea

ui/shared/ad/CustomAdBanner.tsx
jasonzysun: ok

ui/shared/ad/CustomAdBanner.tsx
jasonzysun: great

ui/shared/ad/CustomAdBanner.tsx
jasonzysun: Yes, your reminder was great. I should learn more about some existing configurations in lib.

jasonzysun:
> > I'll add another env variables like NEXT_PUBLIC_DISABLE_DOWNLOAD_AT_RUM_TIME to realize that in other pr.
> 
> I don't really think that we need this kind of setting. In general, the app should not break if the remote config is suddenly changed. Configuration should be persistent for the running instance. That's why we download all these files at run-time. If you need another configuration for your app, then re-run the container with new ENVs and you're good to go. It should not be any different than for other (non-json like) ENVs.

Why not leave this as an optional configuration for deployment users to decide?

nextjs/nextjs-routes.d.ts
tom2drum: Well if you don't revert these changes, I guess next time when someone runs dev build it could cause the opposite change in this file.

But ok, let's leave it as it is and see what happens.

jasonzysun:
> > I'll add another env variables like NEXT_PUBLIC_DISABLE_DOWNLOAD_AT_RUM_TIME to realize that in other pr.
> 
> I don't really think that we need this kind of setting. In general, the app should not break if the remote config is suddenly changed. Configuration should be persistent for the running instance. That's why we download all these files at run-time. If you need another configuration for your app, then re-run the container with new ENVs and you're good to go. It should not be any different than for other (non-json like) ENVs.

Check this issue #1206  , that's anothe reason why I think this is necessary.

tom2drum:
> Why not leave this as an optional configuration for deployment users to decide?

It is fallible and potentially can break app integrity. I don't see any reason why we should make an exception for some config properties and allow them to be dynamic and not persistent. Such behavior should not be accepted on the general level of the app, imho. But users who really need this for some reason are always free to create a fork and make amends.

tom2drum:
> Check this issue https://github.com/blockscout/frontend/issues/1206 , that's anothe reason why I think this is necessary.

I will take a look later when I have time. But I don't think it is going to change the approach.

jasonzysun:
> > Why not leave this as an optional configuration for deployment users to decide?
> 
> It is fallible and potentially can break app integrity. I don't see any reason why we should make an exception for some config properties and allow them to be dynamic and not persistent. Such behavior should not be accepted on the general level of the app, imho. But users who really need this for some reason are always free to create a fork and make amends.

I fully understand your considerations. If you think this is not a feature that most users need, I will only implement it on my own branch.

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const timer = setInterval(() => {
+      setCurrentBannerIndex((prevIndex) => (prevIndex + 1) % banners.length);
+    }, interval);
+
+    return () => {
+      clearInterval(timer);
+    };
+  }, [ interval, banners.length ]);
+
+  if (banners.length === 0) {
```
isstuev, line 44: this part causes a layout shift, because before you fetch config it shows nothing. maybe it's better to require an array of banners to be not empty (for the empty case we have NEXT_PUBLIC_AD_BANNER_PROVIDER=none) and reserve a place for the banner

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,61 @@
...
+    [ 'custom-configs' ],
+    async() => apiFetch(configUrl),
+    {
+      enabled: feature.isEnabled && feature.provider === 'custom',
+      staleTime: Infinity,
+    });
+  const interval = adConfig?.interval || 60000;
+  const banners = adConfig?.banners || [];
+
+  const [ currentBannerIndex, setCurrentBannerIndex ] = useState(0);
```
isstuev, line 30: maybe it would be better in general to start with a random one and then change banners randomly to provide more homogenous distribution?

isstuev (changes requested):

ui/shared/ad/CustomAdBanner.tsx
jasonzysun: Perhaps it's just right not to occupy a position when the array is empty? The effect is equivalent to NEXT_PUBLIC_AD_BANNER_PROVIDER=none

ui/shared/ad/CustomAdBanner.tsx
jasonzysun: That's a good suggestion, but considering that users may want a fixed order, I think it's possible to start randomly and provide a parameter to control whether the next one will be played sequentially or randomly

ui/shared/ad/CustomAdBanner.tsx
isstuev: What I tried to say, empty array is equal to no ad, so there shouldn’t be an empty config for a custom banner. In that case you can remove this condition, and reserve ad place while fetching a config. Otherwise (like in your demo) you can see, that page renders without an ad, then ad appears and it causes a layout shift (latest blocks and txs goes down), which is bad for UX

jasonzysun: Well, I will reserve the height of the advertisement before obtaining the custom configuration file.

jasonzysun:
@tom2drum Please check the latest update. I've develop at https://testnet-qng.qitmeer.io/

docs/ENVS.md
```diff
@@ -228,9 +228,29 @@ This feature is **enabled by default** with the `slise` ads provider. To switch
...
 | NEXT_PUBLIC_AD_ADBUTLER_CONFIG_MOBILE | `{ id: string; width: number; height: number }` | Placement config for mobile Adbutler banner | - | - | `{'id':'654321','width':'300','height':'100'}` |
+| NEXT_PUBLIC_AD_CUSTOM_CONFIG_URL | `string` | URL of configuration file (.json format only) which contains settings and list of custom banners that will be shown in the home page and token detail page. See below list of available properties for particular banner | - | - | `https://example.com/ad_custom_config.json` |
+
+#### Configuration properties
+| Variable | Type | Description | Compulsoriness | Default value | Example value |
+| --- | --- | --- | --- | --- | --- |
+| banners | `array` | List of banners with their properties. Refer to the "Custom banners configuration properties" section below. | Required | - | See below |
+| interval | `number` | Duration (in milliseconds) for how long each banner will be displayed. | - | 60000 | `6000` |
+| randomStart | `boolean` | Set to true to randomly start playing advertisements from any position in the array | - | `false` | `true` |
+| randomNextAd | `boolean` | Set to ture to randomly play advertisements | - | `false` | `true` |
```
tom2drum, line 263:
```suggestion
| randomNextAd | `boolean` | Set to true to randomly play advertisements | - | `false` | `true` |
```

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,73 @@
...
+import useIsMobile from 'lib/hooks/useIsMobile';
+
+const CustomAdBanner = ({ className }: { className?: string }) => {
+  const isMobile = useIsMobile();
+
+  const feature = config.features.adsBanner;
+  const configUrl = (feature.isEnabled && feature.provider === 'custom') ? feature.configUrl : '';
+
+  const apiFetch = useFetch();
+  const { data: adConfig } = useQuery<unknown, ResourceError<unknown>, AdCustomConfig>(
```
tom2drum, line 20: I think we need to manage the following states of the component as well:
- loading - while the config is loading it is good to show a skeleton
- error - if the request fails then render nothing (`null`)

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,73 @@
...
+    );
+  }
+
+  const currentBanner = banners[currentBannerIndex];
+
+  return (
+    <Flex className={ className } h="90px">
+      <Tooltip label={ currentBanner.text } aria-label={ currentBanner.text }>
+        <a href={ currentBanner.url } target="_blank" rel="noopener noreferrer">
+          <Image src={ isMobile ? currentBanner.mobileImageUrl : currentBanner.desktopImageUrl }
```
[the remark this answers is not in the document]
jasonzysun: Yes, I will implement it in conjunction with the `loading` state.

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const [ currentBannerIndex, setCurrentBannerIndex ] = useState(
+    randomStart ? Math.floor(Math.random() * banners.length) : 0,
+  );
+  useEffect(() => {
+    if (banners.length === 0) {
+      return;
+    }
+    const timer = setInterval(() => {
+      if (randomNextAd) {
+        setCurrentBannerIndex(Math.floor(Math.random() * banners.length));
```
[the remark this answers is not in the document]
jasonzysun: The current plan does not take into account this continuous occurrence. I will provide a new plan to address the issue of random playback order.

jasonzysun:
@tom2drum please check latest version

ui/shared/ad/CustomAdBanner.tsx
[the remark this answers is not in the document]
jasonzysun: I achieved random playback by shuffling the array order.

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const currentBanner = banners[currentBannerIndex];
+
+  return (
+    <Flex className={ className } h="90px">
+      <Tooltip label={ currentBanner.text } aria-label={ currentBanner.text }>
+        <a href={ currentBanner.url } target="_blank" rel="noopener noreferrer">
+          <Image src={ isMobile ? currentBanner.mobileImageUrl : currentBanner.desktopImageUrl }
+            alt={ currentBanner.text } height="100%" width="auto" borderRadius="md"
+            fallback={ <Skeleton width={ isMobile ? '270px' : '728px' } height="90px"/> }
```
tom2drum, line 76:
```suggestion
            fallback={ <Skeleton width="100%" height="100%"/> }
```

will this work? the less we rely on absolute values in multiple places, the easier it will maintain in future

ui/shared/ad/CustomAdBanner.tsx
```diff
@@ -0,0 +1,84 @@
...
+    {
+      enabled: feature.isEnabled && feature.provider === 'custom',
+      staleTime: Infinity,
+    });
+
+  const interval = adConfig?.interval || MINUTE;
+  const baseBanners = adConfig?.banners || [];
+  const randomStart = adConfig?.randomStart || false;
+  const randomNextAd = adConfig?.randomNextAd || false;
+  const banners = randomNextAd ? shuffle(baseBanners) : baseBanners;
```
tom2drum, line 33: I would probably do a memoization here, otherwise whenever component re-renders, the new sequence of banners will be generated, so it kind of affects randomness a little bit

deploy/tools/envs-validator/schema.ts
```diff
@@ -132,12 +131,45 @@ const adButlerConfigSchema = yup
...
+const adCustomBannerConfigSchema: yup.ObjectSchema<AdCustomBannerConfig> = yup
+  .object()
+  .shape({
+    text: yup.string(),
+    url: yup.string().test(urlTest),
+    desktopImageUrl: yup.string().test(urlTest).required(),
+    mobileImageUrl: yup.string().test(urlTest).required(),
+  });
+
+const adCustomConfigSchema = yup
```
tom2drum, line 143: Have you checked the schema validation localy as it is described in [docs](https://github.com/blockscout/frontend/blob/main/docs/CONTRIBUTING.md#adding-new-env-variable)?

I guess we are missing `.json()` here, so the yup can parse the string to the actual JSON.

Also in `deploy/tools/envs-validator/index.ts` you need to replace ENV value with the content of the json-file in order to make the validator work.

```typescript
appEnvs.NEXT_PUBLIC_AD_CUSTOM_CONFIG_URL = await getExternalJsonContent(
  './public/assets/ad_custom_config.json',
  appEnvs.NEXT_PUBLIC_AD_CUSTOM_CONFIG_URL,
) || '[]';
```

Can you please temporarily add this config ENV to the `.env.main` preset file to demonstrate that everything works as expected? I will run the validator in our CI then. After that, you can comment it out. I know it is not the best workflow, but we are trying to improve it.

ui/shared/ad/CustomAdBanner.tsx
jasonzysun: Considering that the actual advertising image ratio used by users may not strictly follow the provided ratio, the image width is set to "auto" mode. If the width of the skeleton screen is 100% instead of a fixed size, the width will become 0 when used.

fascinated:
Do I understand correctly that this approach preloads a set of ads from a dynamic JSON manifest once on blockscout app server start? Is there any way to refresh the set of loaded assets in this approach?

jasonzysun:
> Do I understand correctly that this approach preloads a set of ads from a dynamic JSON manifest once on blockscout app server start? Is there any way to refresh the set of loaded assets in this approach?

Yes, your understanding is correct. Due to the design of #1184 , the external customized configuration files referenced by the front-end will be downloaded to the Docker during startup. Therefore, if the customized advertising content in the configuration files is modified externally after startup, the current version of the front-end UI will not respond to changes in a timely manner, and the browser needs to be manually restarted.

fascinated:
@jasonzysun got it, thank you!

I know you've done a lot of work on this, so I must ask - is there a reason you didn't choose an approach like specifying an external iframe (or some external JS that would insert the iframe into the ad slot) url via an ENV value?

This general practice is very common with display advertising systems + does allow you to manage the ad content, frequency, etc completely without making further changes to the blockscout/docker/etc instance.

jasonzysun:
> @jasonzysun got it, thank you!
> 
> I know you've done a lot of work on this, so I must ask - is there a reason you didn't choose an approach like specifying an external iframe (or some external JS that would insert the iframe into the ad slot) url via an ENV value?
> 
> This general practice is very common with display advertising systems + does allow you to manage the ad content, frequency, etc completely without making further changes to the blockscout/docker/etc instance.

In fact, the purpose of this custom feature is not to introduce a new advertising system, but to provide a simple feature that allows users to implement their own advertising with just one configuration file when they have copy and images.

deploy/tools/envs-validator/schema.ts
tom2drum: @jasonzysun I have just merged some changes in the validation script, now it should be easier to add tests for the new variables. I've described all the necessary steps in the contribution guide. Check it out and let me know if you have any questions.

jasonzysun: That's great to hear. I haven't been able to address this section of the content lately, but I'll take care of it in the coming days.

PR #1221, opened 2023-09-27
support alpha-n versions

isstuev:


tom2drum:
it will be fixed here - https://github.com/blockscout/frontend/pull/1218/files

PR #1299, opened 2023-10-22
Add app base path

frisitano:
This PR introduces the `NEXT_PUBLIC_APP_BASE_PATH` environment variable which specifies the base path at which the frontend service will be served on. This feature will allow for multiple frontend instances to be hosted on the same domain. For example we can have two frontends with the following configuration:

```
NEXT_PUBLIC_APP_HOST: www.example.com 
NEXT_PUBLIC_APP_BASE_PATH: /base-path-1
baseUrl: www.example.come/base-path-1

NEXT_PUBLIC_APP_HOST: www.example.com 
NEXT_PUBLIC_APP_BASE_PATH: /base-path-2
baseUrl: www.example.come/base-path-2
```

This change should be backwards compatible as we provide a default value for `NEXT_PUBLIC_APP_BASE_PATH` (``).  

I am currently receiving an error from pre-commit hooks saying that `process.env.*` should only be used in `app.ts` however I can't import the typescript module into the javascript next config.  Maybe we could import it after it's compiled to js or alternatively add a rule to allow it for this specific case?

### DIFF
--- a/ui/snippets/networkMenu/NetworkLogo.tsx
+++ b/ui/snippets/networkMenu/NetworkLogo.tsx
@@ -1,6 +1,7 @@
 import { Icon, Box, Image, useColorModeValue, Skeleton } from '@chakra-ui/react';
 import React from 'react';
 
+import NextLink from 'next/link';
 import { route } from 'nextjs-routes';
 
 import config from 'configs/app';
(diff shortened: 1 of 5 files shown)


tom2drum:
> I can't import the typescript module into the javascript next config

Yes, you are right. Please apply the following diff to eslint config.

```
diff --git a/.eslintrc.js b/.eslintrc.js
index aa89d508..14def4ad 100644
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -305,7 +305,7 @@ module.exports = {
       },
     },
     {
-      files: [ 'configs/**/*.js', 'configs/**/*.ts', '*.config.ts', 'playwright/**/*.ts', 'deploy/tools/**' ],
+      files: [ 'configs/**/*.js', 'configs/**/*.ts', '*.config.ts', 'next.config.js', 'playwright/**/*.ts', 'deploy/tools/**' ],
       rules: {
         // for configs allow to consume env variables from process.env directly
         'no-restricted-properties': [ 0 ],
```

ui/snippets/networkMenu/NetworkLogo.tsx
```diff
@@ -50,39 +51,41 @@ const NetworkLogo = ({ isCollapsed, onClick }: Props) => {
...
-        w="auto"
-        h="100%"
-        src={ iconSrc }
-        alt={ `${ config.chain.name } network logo` }
-        fallback={ <LogoFallback isCollapsed={ isCollapsed } isSmall/> }
-        display={{ base: 'none', lg: isCollapsed === false ? 'none' : 'block', xl: isCollapsed ? 'block' : 'none' }}
-        style={ iconStyle }
-      />
-    </Box>
+    <NextLink href={ '/' } passHref legacyBehavior>
```
tom2drum, line 54: I think, this change is not related to the feature itself. I would leave it as it was, and clicking on the logo will hard-reload the page. So a user will be able to reset the app state if necessary.

tom2drum:
Thanks for contributing to this project, but unfortunately, it seems like this feature is not possible in terms of the current app setup.

The thing is all ENV variables are run-time variables, so the adopters can fully customize the app behavior when they run its instance from a distributed Docker image. However, `basePath` property in the Next.js config cannot be changed at run-time: Next.js builds production bundles based on that. I believe it is so, and I was not able to find any workaround solution in the official Next.js documentation.

We are aware of some limitations of the current tech stack but are not ready yet to get users any alternatives apart from using sub-domains in this case.

@frisitano, I also want to ask you, what is your real-life scenario? Why do you need to have several BS front-end apps in one domain?

frisitano:
> The thing is all ENV variables are run-time variables, so the adopters can fully customize the app behavior when they run its instance from a distributed Docker image. However, `basePath` property in the Next.js config cannot be changed at run-time: Next.js builds production bundles based on that. I believe it is so, and I was not able to find any workaround solution in the official Next.js documentation.

Ahh ok understood - I hadn't realised!  The official documentation states that is the case.

> We are aware of some limitations of the current tech stack but are not ready yet to get users any alternatives apart from using sub-domains in this case.

Understood, thanks for clarifying.  If I manage to find a work around I will maintain a separate fork. 

> @frisitano, I also want to ask you, what is your real-life scenario? Why do you need to have several BS front-end apps in one domain?

I am toying with the idea of a system that can automatically deploy instances without the requirement for the dns configuration of a new domain which can take some time to propagate.

frisitano:
I think the only viable solution of achieving this feature is to run the next build at runtime of the docker container.  I don't expect this feature to be supported by blockscout as it's somewhat hacky.  I will put together a `Dockerfile` to do this and maintain it on a fork. I just wanted to ask if you envision any obvious issues (other than a long startup time) with this pattern?

tom2drum:
> I just wanted to ask if you envision any obvious issues (other than a long startup time) with this pattern?

Apart from that, I guess the container size will be much bigger since you will have to download and store all dependencies in it.

I don't expect that we will do the same thing for the main app package, so please feel free to implement it on your own private fork solely.

Should we close this PR then?

Issue #1245, opened 2023-10-02
Address > NFT improvements


ulyanas:
Waiting for https://github.com/blockscout/blockscout/issues/8231
[PR](https://github.com/blockscout/blockscout/pull/8634)

**1st iteration:** 

-  without filter by collection
-  two tabs: ERC-20 and NFTs

![Image](https://github.com/blockscout/frontend/assets/103045386/b782578b-a9b0-4df9-adaa-ee49bb486345)

-  "Collection" view by default
- "View in token collection" goes to collection page filtered by address
-  you can filter by token type (ERC-721, ERC-1155)
-  Preview of 9 items and "View all NFTs" link to collection page filtered by address

![Image](https://github.com/blockscout/frontend/assets/103045386/6756effa-cc29-4357-bd8a-696397174d6c)

![Image](https://github.com/blockscout/frontend/assets/103045386/0d796c35-de57-49b8-bfde-626394ffbc85)

![Image](https://github.com/blockscout/frontend/assets/103045386/8d7325c8-18be-4518-8095-1265d774983f)

-  ERC-1155 with qty

![Image](https://github.com/blockscout/frontend/assets/103045386/6e85863b-c3b8-4434-a9c3-08a78227bfed)

**Figma** 
https://www.notion.so/blockscout/Design-For-Collections-view-7b7ac288884a4dbda78101363af8683f?pvs=4

https://www.figma.com/file/4In0X8UADoZaTfZ34HaZ3K/Blockscout_2.0?type=design&node-id=14498-45626&mode=design

isstuev:
add redirects from old tabs urls

isstuev:
token balances ws???

NikitaSavik:
Waiting for [backend PR](https://github.com/blockscout/blockscout/pull/8634) review

ulyanas:
Update: waiting for the backend release

PR #1320, opened 2023-11-01
Nft address

isstuev:
resolves https://github.com/blockscout/frontend/issues/1245


ui/address/AddressTokens.tsx
```diff
@@ -58,111 +61,94 @@ const AddressTokens = () => {
...
+    },
   ];
 
+  const nftDisplayTypeRadio = (
+    <RadioButtonGroup<TNftDisplayType>
+      onChange={ handleNFTsDisplayTypeChange }
+      defaultValue={ nftDisplayType }
+      name="type"
+      options={ [
+        { title: 'By collection', value: 'collection', icon: collectionIcon, onlyIcon: isMobile },
```
tom2drum, line 128: Why don't we have `onlyIcon` props for the component itself instead of for each option? Could one of the options be with just an icon and another with text and an icon?

ui/shared/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,97 @@
...
+  onlyIcon: false | undefined;
+} | {
+  title: string;
+  icon: React.FC<React.SVGAttributes<SVGElement>>;
+  onlyIcon: true;
+}
+
+type RadioButtonProps = UseRadioProps & RadioItemProps;
+
+const RadioButton = (props: RadioButtonProps) => {
```
tom2drum, line 17: I think that will be great if we have some visual tests for `RadioButtonGroup` component

tom2drum (approved):

ui/address/AddressTokens.tsx
isstuev: maybe it can, but I chose this way because of typescript

ui/address/tokens/NFTItemContainer.tsx
```diff
@@ -0,0 +1,27 @@
+import { Box, useColorModeValue, chakra } from '@chakra-ui/react';
+import React from 'react';
+
+type Props = {
+  children: React.ReactNode;
+  className?: string;
+};
+
+const NFTItem = ({ children, className }: Props) => {
```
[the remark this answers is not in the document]
isstuev: I divided it from NFTItem because I needed it for "view all" card

ui/address/AddressTokens.tsx
tom2drum: I don't really get why Typescript forced you to have separate flags for each option instead of just one prop for the whole component.

PR #1390, opened 2023-12-04
fix(dockerfile): nit

sambacha:
speed up build and correctly set users in alpine image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,15 @@
 # *****************************
 # *** STAGE 1: Dependencies ***
 # *****************************
-FROM node:18-alpine AS deps
+FROM node:18-alpine3.18 AS deps
 # Check https://github.com/nodejs/docker-node/tree/b4117f9333da4138b03a546ec926ef50a31506c3#nodealpine to understand why libc6-compat might be needed.
-RUN apk add --no-cache libc6-compat
+RUN apk add --no-cache libc6-compat git
 
 ### APP
 # Install dependencies
 WORKDIR /app
 COPY package.json yarn.lock ./
-RUN apk add git
+
 RUN yarn --frozen-lockfile --ignore-optional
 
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -77,17 +77,20 @@ RUN cd ./deploy/tools/envs-validator && yarn build
...
 # Uncomment the following line in case you want to disable telemetry during runtime.
-# ENV NEXT_TELEMETRY_DISABLED 1
+ENV NEXT_TELEMETRY_DISABLED 1
+
+RUN deluser --remove-home node && delgroup node
 
-RUN addgroup --system --gid 1001 nodejs
-RUN adduser --system --uid 1001 nextjs
+RUN addgroup nonroot --gid 1100 && \
+  adduser nonroot --ingroup nonroot --uid 1100 --home /home/nonroot --disabled-password
```
tom2drum, line 93: Can you explain please why this approach is better than the current one which was taken from Next.js docs?

Do we really need to delete the `node` user?

Dockerfile
```diff
@@ -77,17 +77,20 @@ RUN cd ./deploy/tools/envs-validator && yarn build
...
+FROM node:18-alpine3.18 AS runner
 RUN apk add --no-cache --upgrade bash curl jq unzip
 
 ### APP
 WORKDIR /app
 
+ENV NODE_ENV production
 # Uncomment the following line in case you want to disable telemetry during runtime.
-# ENV NEXT_TELEMETRY_DISABLED 1
+ENV NEXT_TELEMETRY_DISABLED 1
```
tom2drum, line 88: Is this change necessary? Do you believe that Next.js telemetry can significantly slow down the build process?

tom2drum:
@sambacha, there has been no activity in this pull request for a month. Are you able to continue working on it, or should we close it?

PR #1398, opened 2023-12-06
Noves integration

juanlenoves:
Here's our code for integrating Noves data with Blockscout. 

Our goal is to significantly improve user experience by displaying Noves-translated transaction types and human-readable descriptions, on both the account and transaction detail pages.

We've been working on this in close collaboration with the Blockscout team, and are very excited for what's to come :)

Happy to make any revisions as needed.


configs/app/features/noves.ts
```diff
@@ -0,0 +1,23 @@
+import type { Feature } from './types';
+
+import { getEnvValue } from '../utils';
+
+const novesEnabled = getEnvValue('NEXT_PUBLIC_NOVES_ENABLED') === 'true';
+
+const title = 'Noves API';
+
+const config: Feature<{ isEnabled: true }> = (() => {
```
tom2drum, line 9: Recently, we introduced our own transaction interpretation service. So, please consider to extend the feature file `configs/app/features/txInterpretation.ts` by adding new value to `NEXT_PUBLIC_TRANSACTION_INTERPRETATION_PROVIDER` env (e.g "noves")

ui/tx/Noves/components/NovesActionCard.tsx
```diff
@@ -0,0 +1,114 @@
...
+import NovesTokensCard from './NovesTokensCard';
+
+interface Props {
+  item: NovesFlowViewItem;
+}
+
+const NovesActionCard: FC<Props> = ({ item }) => {
+  const popoverBg = useColorModeValue('gray.700', 'gray.300');
+
+  const getTokenData = (action: NovesAction) => {
```
tom2drum, line 19: I believe this doesn't need to be a function. It can be simplified to a value using the React.useMemo hook.

ui/tx/Noves/components/NovesActionCard.tsx
```diff
@@ -0,0 +1,114 @@
...
+          <Text fontWeight="500">
+            { item.action.amount }
+          </Text>
+          <TokenEntity
+            token={ getTokenData(item.action) }
+            noCopy
+            noSymbol
+            noLink
+            fontWeight="500"
+            color="link"
```
tom2drum, line 48: Why can it be a link to the token page? Consider using the `onlySymbol` prop instead of `noSymbol`

ui/tx/Noves/components/NovesActionCard.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+          </PopoverTrigger>
+          <PopoverContent
+            bg={ popoverBg }
+            shadow="lg"
+            width="fit-content"
+            zIndex="modal"
+          >
+            <PopoverArrow bg={ popoverBg }/>
+            <Box p={ 2 }>
```
tom2drum, line 99: Can we set padding for `PopoverContent` instead?

ui/tx/Noves/components/NovesTokensCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const NovesTokensCard: FC<Props> = ({ token, amount }) => {
+  const textColor = useColorModeValue('white', 'blackAlpha.900');
+
+  if (!token) {
+    return null;
+  }
+
+  const showTokenName = token.symbol !== token.name;
+  const showTokenAddress = Boolean(token.address.match(/^0x[a-fA-F\d]{40}$/));
```
tom2drum, line 22: Are there cases where the token address is not in the form of a hash?

ui/tx/Noves/components/NovesTokensCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <Box color={ textColor } display="flex" flexDir="column" alignItems="center" gap={ 1 }>
+      <Text as="p" color="inherit" fontWeight="600">
```
tom2drum, line 26: It is the <p> tag by default, no?

ui/tx/Noves/components/NovesTokensCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+import type { NovesNft, NovesToken } from 'types/novesApi';
+
+import CopyToClipboard from 'ui/shared/CopyToClipboard';
+
+interface Props {
+  amount?: string;
+  token: NovesToken | NovesNft | undefined;
+}
+
+const NovesTokensCard: FC<Props> = ({ token, amount }) => {
```
tom2drum, line 14: I think a better name could be `NovesTokenTooltipContent` or something similar.

ui/tx/Noves/components/NovesActionCard.tsx
```diff
@@ -0,0 +1,114 @@
...
+import TokenEntity from 'ui/shared/entities/token/TokenEntity';
+
+import type { NovesAction, NovesFlowViewItem } from '../utils/NovesGenerateFlowViewData';
+import NovesTokensCard from './NovesTokensCard';
+
+interface Props {
+  item: NovesFlowViewItem;
+}
+
+const NovesActionCard: FC<Props> = ({ item }) => {
```
tom2drum, line 16: Maybe the `NovesActionSnippet` name will be more accurate.

ui/tx/Noves/utils/NovesGenerateFlowViewData.ts
```diff
@@ -0,0 +1,102 @@
...
+function getLeftActor(item: NovesSentReceived, perspectiveAddress: string) {
+  if (item.to.address && item.to.address.toLowerCase() === perspectiveAddress) {
+    return { address: item.to.address, name: item.to.name };
+  }
+
+  return { address: item.from.address, name: item.from.name };
+}
+
+function getFlowDirection(item: NovesSentReceived, perspectiveAddress: string): 'toLeft' | 'toRight' {
+  // return "toLeft" or "toRight"
```
tom2drum, line 85: I think this is implied by the return type of the function.

ui/tx/Noves/utils/NovesGenerateFlowViewData.ts
```diff
@@ -0,0 +1,102 @@
...
+    name: string | null;
+  };
+  leftActor: {
+    address: string;
+    name: string | null;
+  };
+  accountAddress: string;
+}
+
+export function NovesGenerateFlowViewData(data: NovesResponseData): Array<NovesFlowViewItem> {
```
tom2drum, line 26: Please add unit tests for all the helper functions.

ui/tx/Noves/utils/NovesGetSplittedDescription.ts
```diff
@@ -0,0 +1,180 @@
...
+    symbolList,
+    idList,
+    names: mappedNames,
+    symbols: mappedSymbols,
+  };
+}
+
+export const NovesGetSplittedDescription = (translateData: NovesResponseData) => {
+  const description = translateData.classificationData.description;
+  const removeEndDot = description.endsWith('.') ? description.slice(0, description.length - 1) : description;
```
tom2drum, line 78:
```suggestion
  const removedEndDot = description.endsWith('.') ? description.slice(0, description.length - 1) : description;
```

ui/tx/Noves/utils/NovesGetSplittedDescription.ts
```diff
@@ -0,0 +1,180 @@
...
+  return {
+    nameList,
+    symbolList,
+    idList,
+    names: mappedNames,
+    symbols: mappedSymbols,
+  };
+}
+
+export const NovesGetSplittedDescription = (translateData: NovesResponseData) => {
```
tom2drum, line 76: Personally, I find it challenging to reason about this function. The logic is complex and heavy. Do you think it's possible to split this function into separate, independent, and clear functions?

tom2drum (changes requested):

nextjs/nextjs-routes.d.ts
```diff
@@ -19,10 +19,13 @@ declare module "nextjs-routes" {
     | StaticRoute<"/api/csrf">
     | StaticRoute<"/api/healthz">
     | StaticRoute<"/api/media-type">
+    | StaticRoute<"/api/noves/describe">
+    | StaticRoute<"/api/noves/history">
+    | StaticRoute<"/api/noves/translate">
```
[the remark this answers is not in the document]
juanlenoves: This wasn't changed because we didn't have the new URLs from the proxy at the time this initial refactoring was done. But yes the plan was always to replace it with the new custom proxy. It will be changed.

ui/txs/TxsListItem.tsx
```diff
@@ -31,24 +32,27 @@ type Props = {
...
-const TxsListItem = ({ tx, isLoading, showBlockInfo, currentAddress, enableTimeIncrement }: Props) => {
+const TxsListItem = ({ tx, isLoading, showBlockInfo, currentAddress, enableTimeIncrement, translate }: Props) => {
   const dataTo = tx.to ? tx.to : tx.created_contract;
 
   const isOut = Boolean(currentAddress && currentAddress === tx.from.hash);
   const isIn = Boolean(currentAddress && currentAddress === tx.to?.hash);
 
   const timeAgo = useTimeAgoIncrement(tx.timestamp, enableTimeIncrement);
 
+  const { data: describeData } = NovesUseFetchDescribe(translate ? tx.hash : null);
```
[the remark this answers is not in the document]
juanlenoves: This functionality was discussed with the rest of the Blockscout team prior to submitting the PR, and it was agreed that it would be included in this page. The design mockups include it too. 

All of our APIs are prepared to handle this level of concurrency (we have layers of caching, etc in place). This calls a separate endpoint than the one called by the transaction detail page (`describeTx` instead of `tx`) because the `describe` endpoint has been optimized to be hit in a highly concurrent way.

So on this, we're going to include the functionality as originally agreed upon.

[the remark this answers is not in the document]
juanlenoves: I should also note that we designed this so that making the calls in parallel wouldn't affect first-time loading of the page. The UI will update as it gets the transaction types from the Noves API, but this shouldn't affect performance for loading the main view.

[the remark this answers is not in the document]
tom2drum: Today, Nikita and I discussed this topic. We believe that a batch request is the most sensible solution. Our suggestion is to cover the tx status label with a skeleton until the data is loaded. If the request fails, we should display a regular tx type.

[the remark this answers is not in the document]
juanlenoves: Alright, we can set up a POST endpoint on our side that receives a batch of transaction hashes and returns the descriptions for all of them at once. Then the number of requests made from the browser is significantly reduced. Let's say we do this for 10 txs at a time? Thoughts?

Also, what would be the skeleton while the data is loaded? In the current version it displays the regular tx type and then replaces it with the Noves type when it arrives.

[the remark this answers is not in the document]
tom2drum: Our standard page size is 50, so you can send all 50 transaction hashes together.

Regarding the skeleton, we suggest rendering the list item with all transaction information except for the transaction type, which will be covered with the skeleton. To achieve this, you can pass the `isLoading` prop to the `TxType` component, which will handle this state.

tom2drum:
Please note that our designer has updated the [mock-ups](https://www.figma.com/file/CEgxqWOzVulwfTUHhs0gUC/%F0%9F%94%A5-RESEARCH%2FDRAFTs?type=design&node-id=704%3A5846&mode=dev) for the assets flow table.

We believe there is no need to truncate the value string. The column width should be sufficient to display the entire value. If required, we can truncate the token symbol instead. Therefore, we can utilize a regular tooltip for the TokenEntity component here instead of creating a custom one.

ui/tx/TxSubHeading.tsx
```diff
@@ -3,45 +3,72 @@ import React from 'react';
...
+
   return (
     <Box display={{ base: 'block', lg: 'flex' }} alignItems="center" w="100%">
-      { hasInterpretation && (
-        <Flex mr={{ base: 0, lg: 6 }} flexWrap="wrap" alignItems="center">
-          <TxInterpretation
-            summary={ txInterpretationQuery.data?.data.summaries[0] }
-            isLoading={ txInterpretationQuery.isPlaceholderData }
-            fontSize="lg"
+      { hasInterpretation && hasNovesInterpretation ?
```
tom2drum, line 53: I believe this change is not entirely accurate. If there is no interpretation, the `TxInterpretation` component will be displayed.

ui/txs/TxsContent.tsx
```diff
@@ -62,7 +66,12 @@ const TxsContent = ({
     setSorting(value);
   }, [ sort, setSorting ]);
 
-  const content = items ? (
+  const translateEnabled = feature.isEnabled && feature.provider === 'noves';
```
tom2drum, line 69: Should we move this logic inside the `useDescribeTxs` hool?

ui/txs/TxsContent.tsx
```diff
@@ -62,7 +66,12 @@ const TxsContent = ({
...
+
+  // Same array of transactions as "items" with "translate" field added
+  const itemsWithTranslate = useDescribeTxs(items, translateEnabled);
```
tom2drum, line 72: The `DataListDisplay` should receive the same array, I suppose.

ui/txs/noves/useDescribeTxs.tsx
```diff
@@ -0,0 +1,89 @@
...
+export interface TransactionWithTranslate extends Transaction {
+  translate?: DescribeTxs;
+}
+
+export default function useDescribeTxs(items: Array<Transaction> | undefined, translateEnabled: boolean) {
+  const txsHash = items?.map(i => i.hash);
+  const txsChunk = _.chunk(txsHash || [], 10);
+
+  const txsParams = Array(5).fill(undefined).map((_, i) => txsChunk[i]);
+  const txsQuerys = useFetchTxs(txsParams, translateEnabled);
```
tom2drum, line 26:
```suggestion
  const txsQueries = useFetchTxs(txsParams, translateEnabled);
```

ui/txs/noves/useDescribeTxs.tsx
```diff
@@ -0,0 +1,89 @@
...
+
```
tom2drum, line 27: I think if the translation is not enabled, we should do an early return of unchanged data

ui/txs/noves/useDescribeTxs.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const isLoading = txsQuerys.some(query => query.isLoading || query.isPlaceholderData);
+  const queryData = txsQuerys.map(query => query.data ? query.data : []).flat();
+
+  const data: Array<TransactionWithTranslate> | undefined = items?.map(tx => {
```
tom2drum, line 31: To prevent unnecessary re-rendering, it's advisable to memoize the result.

ui/txs/noves/useDescribeTxs.tsx
```diff
@@ -0,0 +1,89 @@
...
+  txHash: string;
+  type?: string;
+  description?: string;
+  isLoading: boolean;
+  enabled: boolean;
+}
+
+export interface TransactionWithTranslate extends Transaction {
+  translate?: DescribeTxs;
+}
```
tom2drum, line 19: Since the field is optional, I think it is safe to add it directely to `Transaction` type. But let's rename it to `transalation` (nouns are better than verbs for property names)

ui/txs/noves/useDescribeTxs.tsx
```diff
@@ -0,0 +1,89 @@
+import type { UseQueryResult } from '@tanstack/react-query';
+import _ from 'lodash';
+
+import type { NovesDescribeTxsResponse } from 'types/api/noves';
+import type { Transaction } from 'types/api/transaction';
+
+import type { ResourceError } from 'lib/api/resources';
+
+export interface DescribeTxs {
+  txHash: string;
```
tom2drum, line 10: I assume this is redundant since `hash` property is already present in the `Transaction` type.

ui/txs/noves/useDescribeTxs.tsx
```diff
@@ -0,0 +1,89 @@
...
+  type?: string;
+  description?: string;
+  isLoading: boolean;
+  enabled: boolean;
```
tom2drum, line 14: I think this is redundant too, since absence of `translate` prop means the feature is disabled.

ui/txs/noves/useDescribeTxs.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+import type { NovesDescribeTxsResponse } from 'types/api/noves';
+import type { Transaction } from 'types/api/transaction';
+
+import type { ResourceError } from 'lib/api/resources';
+
+export interface DescribeTxs {
+  txHash: string;
+  type?: string;
+  description?: string;
```
tom2drum, line 12: Maybe it is better to move these properties under `data` fields, so the structure will be similar to `QueryResult` type.

```
data: { type: string; description: string } | null;
isLoading: boolean;
```

ui/txs/noves/useDescribeTxs.tsx
```diff
@@ -0,0 +1,89 @@
...
+  enabled: boolean;
+}
+
+export interface TransactionWithTranslate extends Transaction {
+  translate?: DescribeTxs;
+}
+
+export default function useDescribeTxs(items: Array<Transaction> | undefined, translateEnabled: boolean) {
+  const txsHash = items?.map(i => i.hash);
+  const txsChunk = _.chunk(txsHash || [], 10);
```
tom2drum, line 23: The uniqueness of each item should also be checked, as there may be multiple rows with the same transaction hash on a single page.

ui/txs/noves/useDescribeTxs.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const txsParams = Array(5).fill(undefined).map((_, i) => txsChunk[i]);
```
tom2drum, line 25: Shouldn't we pass the chunks directly into the `useFetchTxs` hook? I don't really understand the reason for formatting it in this manner.

ui/tx/TxAssetFlows.tsx
```diff
@@ -0,0 +1,122 @@
...
+    queryOptions: {
+      enabled: Boolean(props.hash),
+      placeholderData: NOVES_TRANSLATE,
+    },
+  });
+
+  const [ page, setPage ] = useState<number>(1);
+
+  const ViewData = queryData ? generateFlowViewData(queryData) : [];
+  const chunkedViewData = _.chunk(ViewData, 10);
```
tom2drum, line 35: If pagination is not available for this resource on the server side, I question whether we should implement it on the client side.

By the way, our standard page size is 50.

ui/tx/assetFlows/components/NovesTokenTransferSnippet.tsx
```diff
@@ -0,0 +1,46 @@
+import { Flex } from '@chakra-ui/react';
+import React from 'react';
+
+import NftEntity from 'ui/shared/entities/nft/NftEntity';
+import TokenEntity from 'ui/shared/entities/token/TokenEntity';
+
+interface Token {
```
tom2drum, line 7: Shouldn't this type be derived from the `TokenInfo` then, if it is used as a prop for `TokenEntity` component?

ui/tx/assetFlows/components/NovesActionSnippet.tsx
```diff
@@ -0,0 +1,115 @@
...
+import type { NovesFlowViewItem } from '../utils/generateFlowViewData';
+import NovesTokenTooltipContent from './NovesTokenTooltipContent';
+
+interface Props {
+  item: NovesFlowViewItem;
+  isLoaded: boolean;
+}
+
+const NovesActionSnippet: FC<Props> = ({ item, isLoaded }) => {
+  const popoverBg = useColorModeValue('gray.700', 'gray.300');
```
tom2drum, line 17: Has our designer agreed to use a custom background for the tooltip?

tom2drum (changes requested):

ui/tx/TxAssetFlows.tsx
juanlenoves: I understand the page size is 50 on your side, but on our side we're more comfortable sticking to the original 10 (for our API call).

ui/tx/assetFlows/components/NovesActionSnippet.tsx
juanlenoves: I was under the impression that yes

ui/tx/Noves/NovesTxAssetFlows.tsx
```diff
@@ -0,0 +1,160 @@
...
+                  <Tr>
+                    <Th fontWeight="bold" fontFamily="inherit">
+                    Actions
+                    </Th>
+                    <Th fontWeight="bold" fontFamily="inherit" >
+                    From/To
+                    </Th>
+                  </Tr>
+                </Thead>
+                <Tbody>
```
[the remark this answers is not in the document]
juanlenoves: So basically:

```
<Table variant="simple" size="sm">
```

?

ui/tx/Noves/NovesTxAssetFlows.tsx
```diff
@@ -0,0 +1,160 @@
...
+        </Box>
+      </div>
+    </Box>
+  );
+
+  return (
+    <DataListDisplay
+      isError={ isError }
+      items={ data }
+      emptyText="There are no transactions."
```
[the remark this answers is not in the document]
juanlenoves: Changed wording in upcoming commit

ui/tx/Noves/components/NovesTokensCard.tsx
juanlenoves: Not really on our side. When the asset is the native coin the address won't be a hash, but there's no real "address" concept at that point. So it makes sense not to display the address for "ETH" because that doesn't really exist.

ui/shared/Noves/NovesFromToComponent.tsx
```diff
@@ -0,0 +1,68 @@
...
+          textAlign="center"
+        >
+          { data.text }
+        </TagLabel>
+        <TagLabel visibility="hidden" whiteSpace="nowrap" px={ 2 } >Received from</TagLabel>
+      </Tag>
+
+      <Box ml="2" textOverflow="ellipsis" overflow="hidden" whiteSpace="nowrap">
+        <AddressEntity
+          address={ addressObj }
```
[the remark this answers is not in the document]
juanlenoves: Changed in upcoming commit

ui/txs/noves/useDescribeTxs.tsx
juanlenoves: Changed in upcoming commit

ui/tx/TxAssetFlows.tsx
NahuelNoves: We extract all the "transfers" items in this tab from `api/v2/proxy/noves-fi/transactions/${th_hash}` so it doesn't need pagination but it can have a lot of transfers, for example https://goerli.etherscan.io/tx/0x3fd8edc3cc1c891d2a07d31f20c45561d942f7b7d3487b6f4c708dc9bd0c6d4e

So i chunk the transfers to display 10 per "page"

ui/shared/Noves/NovesFromToComponent.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    <Box display="flex">
+      <Tag
+        colorScheme={ isSent ? 'yellow' : 'green' }
+        px={ 0 }
+        pos="relative"
+        minW="max-content"
+      >
+        <TagLabel
+          position="absolute"
```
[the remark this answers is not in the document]
NahuelNoves: Removed this TagLabel

ui/shared/Noves/NovesFromToComponent.tsx
```diff
@@ -0,0 +1,68 @@
...
+          w="full"
+          textAlign="center"
+        >
+          { data.text }
+        </TagLabel>
+        <TagLabel visibility="hidden" whiteSpace="nowrap" px={ 2 } >Received from</TagLabel>
```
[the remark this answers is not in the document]
NahuelNoves: Removed the hidden Prop.

We had a hidden label, and another absolute positioned with a fixed text to assure the width was the same

ui/tx/Noves/components/NovesActionCard.tsx
NahuelNoves: It may be cases were we don't have the token address, so we check if it's available

juanlenoves:
I believe all items from previous submissions have been addressed with the latest push. Please correct me if I'm wrong, but we should be very close or basically done here (and I hope that's a true statement 🙏)

ui/tx/TxAssetFlows.tsx
tom2drum: Can we set the page size to 50 then? So it will be consistent with the other lists.

juanlenoves: Done

lib/api/resources.ts
```diff
@@ -580,6 +587,23 @@ export const RESOURCES = {
...
+    pathParams: [ 'hash' as const ],
+  },
+  noves_address_history: {
+    path: '/api/v2/proxy/noves-fi/addresses/:address/transactions',
+    pathParams: [ 'address' as const ],
+    filterFields: [],
+  },
+  noves_describe_txs: {
+    path: '/api/v2/proxy/noves-fi/transactions',
+    queryParams: [
```
tom2drum, line 602: We typically do not describe all query parameters for resources due to their potential abundance. Is it necessary? I cannot find any other instance in the code where it is utilized.

ui/tx/Noves/NovesTxAssetFlows.tsx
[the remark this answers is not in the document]
tom2drum: yes, but the header of the table should be sticky.

refer to `ui/txs/TxsTable.tsx` as an example

ui/tx/assetFlows/components/NovesActionSnippet.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const token = {
+      name: name,
+      symbol: symbol?.toLowerCase() === name?.toLowerCase() ? undefined : symbol,
+      address: action.nft?.address || action.token?.address,
+    };
+
+    return token;
+  }, [ item.action ]);
+
+  const validTokenAddress = token.address ? Boolean(token.address.match(/^0x[a-fA-F\d]{40}$/)) : false;
```
tom2drum, line 34: We already have the regular expression for address hash. Please refer to `lib/regexp.ts`.

ui/tx/assetFlows/components/NovesTokenTooltipContent.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const NovesTokenTooltipContent: FC<Props> = ({ token, amount }) => {
+  const textColor = useColorModeValue('white', 'blackAlpha.900');
+
+  if (!token) {
+    return null;
+  }
+
+  const showTokenName = token.symbol !== token.name;
+  const showTokenAddress = Boolean(token.address.match(/^0x[a-fA-F\d]{40}$/));
```
tom2drum, line 22: The same comment about regular expressions here.

ui/txs/noves/useDescribeTxs.tsx
tom2drum: This issue has not been addressed yet. Even if the feature is disabled, some unnecessary computation will still occur and it will create 5 unnecessary hanging queries. 

To improve performance and memory consumption, I suggest rewriting this hook as follows:

```js
import useApiFetch from 'lib/api/useApiFetch';
import { useQuery } from '@tanstack/react-query';
import { getResourceKey } from 'lib/api/useApiQuery';

export default function useDescribeTxs(items: Array<Transaction> | undefined, viewAsAccountAddress: string | undefined) {

  const describeQuery = useQuery({
    // in queryKey we should probably also pass the page number so requests are properly cached
    // or use for example { firstHash, lastHash } props
    queryKey: [ 'noves_describe_txs', { viewAsAccountAddress } ],
    queryFn: async() => {
      // prepare data for Noves API call
      // ...
      const txChunks: Array<Array<string>> = [];

      const queries = txChunks.map((hashes) => {
        if (hashes.length === 0) {
          return Promise.resolve([]);
        }

        return apiFetch('noves_describe_txs', {
          queryParams: {
            viewAsAccountAddress,
            hashes,
          },
        });
      });

      return Promise.all(queries);
    },
    select: (data) => {
      // do all data manipulation here, like flattening and mapping
    },
    enabled: translateEnabled,
  });

  if (!translateEnabled) {
    return items;
  }

  return React.useMemo(() => {
    // modify items with translation data and isLoading flag
  }, []);

}

```

juanlenoves:
Hi Tom,

Our latest commit should address all pending comments, and we believe this should be ready to go.

I made a final adjustment and assumed that for describe_txs, the API endpoint that you expose via the custom proxy will be called "transaction-descriptions" and not just "transactions". This is much more accurate and descriptive of what that endpoint returns (just the minimalistic output with the descriptions, not the full transactions).

ui/address/AddressAccountHistory.tsx
```diff
@@ -0,0 +1,130 @@
...
+          </Tbody>
+        </Table>
+      </Show>
+    </Box>
+  );
+
+  return (
+    <>
+      { /* should stay before tabs to scroll up with pagination */ }
+      <Box ref={ scrollRef }></Box>
```
tom2drum, line 113: Shouldn't it be passed as a prop like it's done for other tabs components on the address page?

tom2drum (commented):
> I made a final adjustment and assumed that for describe_txs, the API endpoint that you expose via the custom proxy will be called "transaction-descriptions" and not just "transactions".

Has the underlying endpoint in Noves API been changed? We have already deployed our proxy service. Can we keep using the previous name? If not, we need to wait for the proxy to be redeployed before building a demo.

tom2drum:
In order to run code checks and tests, could you please merge the main branch into yours and resolve any conflicts?

juanlenoves:
> > I made a final adjustment and assumed that for describe_txs, the API endpoint that you expose via the custom proxy will be called "transaction-descriptions" and not just "transactions".
> 
> Has the underlying endpoint in Noves API been changed? We have already deployed our proxy service. Can we keep using the previous name? If not, we need to wait for the proxy to be redeployed before building a demo.

No the underlying endpoint hasn't changed at all. So yes what we're requesting is that the URL path for the proxy is renamed, to better reflect what the underlying point does.

The rationale is that the proxy currently calls it "transactions", but the underlying endpoint only returns "transaction descriptions", no the full transactions. If we needed to later add a proxy endpoint for our "full transactions" endpoint, the correct proxy path for that would've already be taken. Hope that makes sense. We just want to keep the naming of everything as consistent and accurate as possible.

I understand this means having to change the proxy, but the only thing that needs to change is the URL path.

We'll resolve merge conflicts and push another commit.

tom2drum:
> I understand this means having to change the proxy, but the only thing that needs to change is the URL path.

I understand your reasons. I have forwarded the request to change the proxy endpoint to our devops team. After they make the change, I will deploy the demo app.

tom2drum:
@juanlenoves could you please apply this fix of tsc build?

```
diff --git a/lib/api/resources.ts b/lib/api/resources.ts
index 526a7315..dc259df4 100644
--- a/lib/api/resources.ts
+++ b/lib/api/resources.ts
@@ -854,9 +854,6 @@ Q extends 'zkevm_l2_txn_batches_count' ? number :
 Q extends 'zkevm_l2_txn_batch' ? ZkEvmL2TxnBatch :
 Q extends 'zkevm_l2_txn_batch_txs' ? ZkEvmL2TxnBatchTxs :
 Q extends 'config_backend_version' ? BackendVersionConfig :
-Q extends 'noves_transaction' ? NovesResponseData :
-Q extends 'noves_address_history' ? NovesAccountHistoryResponse :
-Q extends 'noves_describe_txs' ? NovesDescribeTxsResponse :
 Q extends 'addresses_lookup' ? EnsAddressLookupResponse :
 Q extends 'domain_info' ? EnsDomainDetailed :
 Q extends 'domain_events' ? EnsDomainEventsResponse :
@@ -880,6 +877,9 @@ Q extends 'shibarium_deposits' ? ShibariumDepositsResponse :
 Q extends 'shibarium_withdrawals_count' ? number :
 Q extends 'shibarium_deposits_count' ? number :
 Q extends 'contract_security_audits' ? SmartContractSecurityAudits :
+Q extends 'noves_transaction' ? NovesResponseData :
+Q extends 'noves_address_history' ? NovesAccountHistoryResponse :
+Q extends 'noves_describe_txs' ? NovesDescribeTxsResponse :
 never;
 /* eslint-enable @typescript-eslint/indent */
```

juanlenoves:
> @juanlenoves could you please apply this fix of tsc build?
> 
> ```
> diff --git a/lib/api/resources.ts b/lib/api/resources.ts
> index 526a7315..dc259df4 100644
> --- a/lib/api/resources.ts
> +++ b/lib/api/resources.ts
> @@ -854,9 +854,6 @@ Q extends 'zkevm_l2_txn_batches_count' ? number :
>  Q extends 'zkevm_l2_txn_batch' ? ZkEvmL2TxnBatch :
>  Q extends 'zkevm_l2_txn_batch_txs' ? ZkEvmL2TxnBatchTxs :
>  Q extends 'config_backend_version' ? BackendVersionConfig :
> -Q extends 'noves_transaction' ? NovesResponseData :
> -Q extends 'noves_address_history' ? NovesAccountHistoryResponse :
> -Q extends 'noves_describe_txs' ? NovesDescribeTxsResponse :
>  Q extends 'addresses_lookup' ? EnsAddressLookupResponse :
>  Q extends 'domain_info' ? EnsDomainDetailed :
>  Q extends 'domain_events' ? EnsDomainEventsResponse :
> @@ -880,6 +877,9 @@ Q extends 'shibarium_deposits' ? ShibariumDepositsResponse :
>  Q extends 'shibarium_withdrawals_count' ? number :
>  Q extends 'shibarium_deposits_count' ? number :
>  Q extends 'contract_security_audits' ? SmartContractSecurityAudits :
> +Q extends 'noves_transaction' ? NovesResponseData :
> +Q extends 'noves_address_history' ? NovesAccountHistoryResponse :
> +Q extends 'noves_describe_txs' ? NovesDescribeTxsResponse :
>  never;
>  /* eslint-enable @typescript-eslint/indent */
> ```

Done

nextjs/nextjs-routes.d.ts
```diff
@@ -39,6 +39,7 @@ declare module "nextjs-routes" {
     | StaticRoute<"/login">
     | StaticRoute<"/name-domains">
     | DynamicRoute<"/name-domains/[name]", { "name": string }>
+    | StaticRoute<"/name-domains">
```
tom2drum, line 42: There are two duplicated routes `StaticRoute<"/name-domains">` now

ui/address/AddressAccountHistory.tsx
```diff
@@ -58,7 +57,6 @@ const AddressAccountHistory = ({ scrollRef }: Props) => {
         isLoading={ pagination.isLoading }
       />
 
-      <Pagination ml={{ base: 'auto', lg: 8 }} { ...pagination }/>
```
tom2drum: If pagination is no longer used, consider using the regular `useApiQuery` hook instead of `useQueryWithPages`.

ui/address/accountHistory/AddressAccountHistoryTableItem.tsx
```diff
@@ -16,16 +16,22 @@ type Props = {
...
+  const parsedDescription = useMemo(() => {
+    const description = props.tx.classificationData.description;
+
+    return description.endsWith('.') ? description.substring(0, description.length - 1) : description;
+  }, [ props.tx.classificationData.description ]);
+
   return (
     <Tr>
-      <Td px={ 3 } py="18px" fontSize="sm" borderColor="gray.200">
+      <Td px={ 3 } py="18px" fontSize="sm" borderColor="gray.200" _dark={{ borderColor: 'whiteAlpha.200' }}>
```
tom2drum, line 27: I believe the standard styles for the table should be used, you don't have to redefine them.

ui/shared/pagination/useQueryWithPages.ts
```diff
@@ -41,6 +41,17 @@ function getNextPageParams<R extends PaginatedResources>(data: ResourcePayload<R
...
+  const nextPageParams = data.next_page_params;
+  if (nextPageParams) {
+    // check if a page param is "null" and remove it
+
+    return Object.entries(nextPageParams)
+    //eslint-disable-next-line @typescript-eslint/no-unused-vars
+      .filter(([ _, v ]) => v != null)
+      .reduce((acc, [ k, v ]) => ({ ...acc, [k]: v }), {});
+
+  }
```
tom2drum, line 53: Can you explain logic behind this? Why should it be removed? The `null` is a valid value of `next_page_params` property for all of our API response type

ui/tx/assetFlows/TxAssetFlowsTableItem.tsx
```diff
@@ -15,10 +15,10 @@ const TxAssetFlowsTableItem = (props: Props) => {
 
   return (
     <Tr >
-      <Td px={ 3 } py={ 5 } fontSize="sm" borderColor="gray.200">
+      <Td px={ 3 } py={ 5 } fontSize="sm" borderColor="gray.200" _dark={{ borderColor: 'whiteAlpha.200' }}>
```
tom2drum, line 18: Same comment about table borders.

tom2drum (changes requested):
+ The issue with space between tx subheading text and icon is not fixed for loading state

<details><summary>Screenshot</summary>
Screenshot 2024-03-06 at 11 33 47 AM
</details>

ui/shared/pagination/useQueryWithPages.ts
francisco-noves: This code is neccesary because the endpoint thats being use in Account History can return some null fields that are important for getting the next page. If we remove it, the second page can fail.

juanlenoves: Yeah so essentially to have pagination working for `noves_address_history` this appears to be necessary at the moment.

tom2drum: > This code is neccesary because the endpoint thats being use in Account History can return some null fields that are important for getting the next page

As I understand the code, it removes from `next_page_params` object all fields with `null` value. As you said, why are they being removed if they are important? Nevertheless, this change will break integration with our API since it relies on all fields being present in the request for the next page, even on those whose value is `null`.

tom2drum: My guess is if the parameter is not needed for the next page request, it should be omitted from `next_page_params`.

juanlenoves: We changed how our backend works to accommodate this. The code mentioned in this convo has been removed.

ui/tx/TxSubHeading.tsx
```diff
@@ -105,7 +105,7 @@ const TxSubHeading = ({ hash, hasTag, txQuery }: Props) => {
     <Box display={{ base: 'block', lg: 'flex' }} alignItems="center" w="100%">
       { content }
       <Flex alignItems="center" justifyContent={{ base: 'start', lg: 'space-between' }} flexGrow={ 1 }>
-        { !hasTag && <AccountActionsMenu mr={ 3 } mt={{ base: 3, lg: 0 }}/> }
+        { !hasTag && <AccountActionsMenu m={ 3 } mt={{ base: 3, lg: 0 }}/> }
```
tom2drum, line 108: I think this change was a mistake.

ui/tx/assetFlows/components/NovesSubHeadingInterpretation.tsx
```diff
@@ -61,7 +62,17 @@ const NovesSubHeadingInterpretation: FC<Props> = ({ data, isLoading }) => {
                   fontSize="lg"
                   w="fit-content"
                 />
-              ) }
+              )
+            }
+            {
+              item.address && (
+                <AddressEntity
+                  address={{ hash: item.address, is_contract: true, is_verified: true }}
```
tom2drum, line 70: Not all contracts are verified. I suggest using a regular contract icon here instead of a verified one to avoid misleading the user.

tom2drum (commented):
👍 I will re-deploy the demo

tom2drum (approved):

Issue #1418, opened 2023-12-14
CredShields: request only if contract is verified (or partially verified)


NikitaSavik:
Related to https://github.com/orgs/blockscout/projects/6/views/1?filterQuery=cred&pane=issue&itemId=45799554
To avoid spam on QuickScan service, lets check for contract verification before making request

PR #1422, opened 2023-12-14
soldity check only for validated contracts + keys fix

isstuev:
resolves https://github.com/blockscout/frontend/issues/1418

### DIFF
--- a/ui/address/SolidityscanReport.tsx
+++ b/ui/address/SolidityscanReport.tsx
@@ -45,6 +45,30 @@ interface Props {
   hash: string;
 }
 
+type ItemProps = {
+  item: DistributionItem;
+  vulnerabilities: SolidityscanReport['scan_report']['scan_summary']['issue_severity_distribution'];
+  vulnerabilitiesCount: number;
+}
+
+const SolidityScanReportItem = ({ item, vulnerabilities, vulnerabilitiesCount }: ItemProps) => {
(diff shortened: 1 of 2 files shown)


ui/address/SolidityscanReport.tsx
```diff
@@ -45,6 +45,30 @@ interface Props {
...
 
+type ItemProps = {
+  item: DistributionItem;
+  vulnerabilities: SolidityscanReport['scan_report']['scan_summary']['issue_severity_distribution'];
+  vulnerabilitiesCount: number;
+}
+
+const SolidityScanReportItem = ({ item, vulnerabilities, vulnerabilitiesCount }: ItemProps) => {
+  const bgBar = useColorModeValue('blackAlpha.50', 'whiteAlpha.50');
+  const yetAnotherGrayColor = useColorModeValue('gray.400', 'gray.500');
```
tom2drum, line 56: can we use `text_secondary` variable?

tom2drum (approved):

ui/address/SolidityscanReport.tsx
isstuev: no, it's the opposite

tom2drum: Should we consider changing the design? The excessive use of multiple shades of gray for the text feels overwhelming.

Issue #1431, opened 2023-12-15
Add the banner for connect wallet button


maikReal:
## What to do?

If a user opens an app on the marketplace, we need to show him a banner with information to connect a wallet 

**Copies:** "Connect your wallet once to use it with all dApps on the marketplace!" @andogro Can you review pls? 

**Criteria to show:**
1. If a user opens an app and the global wallet is not connected, show the tooltip
2. The tooltip can be closed by clicking on any space 
3. The tooltip will be closed automatically in 3 seconds if a user doesn't interact with the UI
4. Show once for a user on the dapp page. If a user already saw the tooltip, but not inside the dapp, he'll see this tooltip, because it's different

> **Let's also automatically close the banner for users who see the "Connect button" for the first time. The banner's behavior is described in this issue: https://github.com/blockscout/frontend/issues/1270**

![Image](https://github.com/blockscout/frontend/assets/26190360/c8805694-172e-4137-ad8d-7767166631c0)

andogro:
message suggestion "Connect once to use your wallet with all apps in the DAppscout marketplace!"

PR #1450, opened 2023-12-20
Add new text for `connect wallet` button tooltip

maxaleks:
Resolves https://github.com/blockscout/frontend/issues/1431


ui/snippets/walletMenu/WalletTooltip.tsx
```diff
@@ -8,20 +9,31 @@ type Props = {
...
+  const router = useRouter();
   const [ isTooltipShown, setIsTooltipShown ] = useBoolean(false);
 
+  const { defaultLabel, label, localStorageKey } = React.useMemo(() => {
+    const isAppPage = router.pathname === '/apps/[id]';
+    const defaultLabel = <span>Your wallet is used to interact with<br/>apps and contracts in the explorer</span>;
+    const label = isAppPage ?
+      <span>Connect once to use your wallet with<br/>all apps in the DAppscout marketplace!</span> :
+      defaultLabel;
+    const localStorageKey = `${ isAppPage ? 'dapp-' : '' }wallet-connect-tooltip-shown-${ isMobile ? 'mobile' : 'desktop' }`;
```
tom2drum, line 21: I wonder why there are separate keys for mobile and desktop devices. It seems unnecessary since user sessions and data cannot be shared between the two.

tom2drum (approved):

ui/snippets/walletMenu/WalletTooltip.tsx
maxaleks: The problem is that both components are rendered at the same time, and if you use 1 key, only the tooltip for one of the device types will be displayed. I tried using Show/Hide components, but the result was even worse, because the button itself appeared with a delay. If you have an idea, I'm ready to redo it, because I don't really like this solution myself

tom2drum: Oh, I see now. Then, maybe the parent component should pass `isDisabled=true` prop if viewport size doesn't match. Will it work?

tom2drum (approved):

Issue #1435, opened 2023-12-18
Transaction interpretation


NikitaSavik:
Add a subtitle, based on our tx-interpreter service
<img width="394" alt="image" src="https://github.com/blockscout/frontend/assets/1701762/70be5504-1731-44bb-ac73-331bca4d1971">

- API call structure: /api/v2/transactions/{hash}/summary
- Summary result [JSON variants](https://docs.google.com/spreadsheets/d/1BDty9daydEAAScp-aoHdtH1GQTEMLE8NEo4uI6d_wq4/edit#gid=0)
- [Figma prototype](https://www.figma.com/file/CEgxqWOzVulwfTUHhs0gUC/%F0%9F%94%A5-RESEARCH%2FDRAFTs?type=design&node-id=29-1633&mode=design&t=cE50iJDxJ1RuxhBW-0)

PR #1457, opened 2023-12-21
Transaction interpretations

isstuev:

resolves https://github.com/blockscout/frontend/issues/1435

### Additional Information

added NEXT_PUBLIC_TRANSACTION_INTERPRETATION_ENABLED variable to enable transaction interpretations

## Checklist for PR author


ui/tx/details/txDetailsActions/TxDetailsActionsFallback.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import TxDetailsAction from './TxDetailsAction';
+import TxDetailsActionsWrapper from './TxDetailsActionsWrapper';
+
+interface Props {
+  actions: Array<TxAction>;
+  isLoading: boolean;
+}
+
+const TxDetailsActionsFallback = ({ actions, isLoading }: Props) => {
```
tom2drum, line 13: The "fallback" suffix in the component name is a bit confusing. Could we use "TxDetailsActionsInterpreted" and "TxDetailsActionsRaw" instead? In my understanding, "fallback" is typically used to display something when an error occurs.

ui/tx/interpretation/TxInterpretation.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const TxInterpretationElementByType = ({ variable }: { variable?: TxInterpretationVariable }) => {
+  if (!variable) {
+    return null;
+  }
+
+  const { type, value } = variable;
+  switch (type) {
+    case 'address':
+      return <AddressEntity address={ value } truncation="constant" sx={{ ':not(:first-child)': { marginLeft: 1 } }} whiteSpace="initial"/>;
```
tom2drum, line 29:
```suggestion
      return <AddressEntity address={ value } truncation="constant" _notFirst={{ ml: 1 }} whiteSpace="initial"/>;
```

ui/tx/interpretation/TxInterpretation.tsx
```diff
@@ -0,0 +1,94 @@
...
+        />
+      );
+    case 'currency': {
+      let numberString = '';
+      if (BigNumber(value).isLessThan(0.1)) {
+        numberString = BigNumber(value).toPrecision(2);
+      } else if (BigNumber(value).isLessThan(10000)) {
+        numberString = BigNumber(value).dp(2).toFormat();
+      } else {
+        numberString = BigNumber(value).dividedBy(1000).toFormat(2) + 'K';
```
tom2drum, line 49: If the value is greater than a million, should we compact it with the "M" suffix then?

ui/tx/interpretation/TxInterpretation.tsx
```diff
@@ -0,0 +1,94 @@
...
+      }
+      return <Text>{ numberString + ' ' }</Text>;
+    }
+    case 'timestamp':
+      // timestamp is in unix format
+      return <Text color="text_secondary">{ dayjs(Number(value) * 1000).format('llll') + ' ' }</Text>;
```
tom2drum, line 55: Perhaps it would be better to include a space for each non-last item in the parent component instead.

tom2drum (changes requested):

ui/tx/interpretation/TxInterpretation.tsx
isstuev: we don't need space for addresses and tokens

tom2drum (approved):

PR #1485, opened 2024-01-08
Fix wallet tooltip

maxaleks:
- automatically show the tooltip only on the marketplace page
- increase the auto-hide time from 3 to 5 seconds
- add a 1 second delay before showing the tooltip to allow the page to load a bit

### DIFF
--- a/ui/snippets/walletMenu/WalletTooltip.tsx
+++ b/ui/snippets/walletMenu/WalletTooltip.tsx
@@ -2,6 +2,8 @@ import { Tooltip, useBoolean, useOutsideClick } from '@chakra-ui/react';
 import { useRouter } from 'next/router';
 import React from 'react';
 
+import { SECOND } from 'lib/consts';
+
 type Props = {
   children: React.ReactNode;
   isDisabled?: boolean;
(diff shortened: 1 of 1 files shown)


ui/snippets/walletMenu/WalletTooltip.tsx
```diff
@@ -26,12 +26,15 @@ const WalletTooltip = ({ children, isDisabled, isMobile }: Props) => {
...
-    if (!isDisabled && !wasShown) {
-      setIsTooltipShown.on();
-      window.localStorage.setItem(localStorageKey, 'true');
-      setTimeout(() => setIsTooltipShown.off(), 3000);
+    const isMarketplacePage = [ '/apps', '/apps/[id]' ].includes(router.pathname);
+    if (!isDisabled && !wasShown && isMarketplacePage) {
+      setTimeout(() => {
+        setIsTooltipShown.on();
+        window.localStorage.setItem(localStorageKey, 'true');
+        setTimeout(() => setIsTooltipShown.off(), 5000);
```
tom2drum, line 34: for better readability you could write `5 * SECOND` and `SECOND` comes from `lib/consts.ts`

but this way is ok too

tom2drum (approved):

tom2drum (approved):

Issue #1496, opened 2024-01-12
Integration mark for the marketplace apps


maikReal:
🦄 [PROPDUCT TASK](https://www.notion.so/blockscout/100-integration-mark-for-an-app-in-the-marketplace-4cc50321181248b8a9a9aaa718f62258?pvs=4)

## What to do?

We need to show an integration label on cards for users, so they have a clear understanding of where and how apps will be opened 

**IMPORTANT ASPECTS:**
1. This task should be launched as an AB experiment with the following tasks:
    - https://github.com/blockscout/frontend/issues/1498
    - https://github.com/blockscout/frontend/issues/1499
    **[All information about the experiment](https://www.notion.so/blockscout/Experiment-1-858bd18b6207478481add09d189ce9e6?pvs=4)**
3. This task blocks the following tasks: https://github.com/blockscout/frontend/issues/1498

Cheatsheet about AB test: https://www.notion.so/blockscout/AB-testing-fafb7ad113c14eaeb742fadd8032e4b0?pvs=4

PR #1528, opened 2024-01-23
Add the integration mark

maxaleks:
- added the integration mark (icon)
- updated styles for an app card
- wrapped the changes into A/B testing

Resolves https://github.com/blockscout/frontend/issues/1496


icons/integration/full.svg
tom2drum, line 1: If the icon is monochrome, it's better to use `fill=currentColor` here and later set the necessary color in the specific component in the code.

tom2drum (approved):

Issue #1559, opened 2024-01-31
fix svg with no dimensions in fullscreen


isstuev:
https://blockscout.testnet.ocean.jellyfishsdk.com/token/0x86c9440182DF725fec2d3ecAAAFCFe1b2A70BdB5/instance/46?tab=metadata

PR #1562, opened 2024-01-31
fix svg fullscreen

isstuev:
resolves https://github.com/blockscout/frontend/issues/1559

also found unused components and removed them


ui/shared/nft/NftImageFullscreen.tsx
```diff
@@ -11,12 +11,21 @@ interface Props {
...
 }
 
-const NftImageWithFullscreen = ({ src, isOpen, onClose }: Props) => {
+const NftImageFullscreen = ({ src, isOpen, onClose }: Props) => {
+  const imgRef = React.useRef<HTMLImageElement>(null);
+
+  const checkWidth = React.useCallback(() => {
+    if (imgRef.current?.getBoundingClientRect().width === 0) {
+      imgRef.current.style.width = '90vw';
+      imgRef.current.style.height = '90vh';
```
tom2drum, line 20: Shouldn't we set these properties through React component props instead of directly manipulating the DOM tree?

tom2drum (approved):

Issue #1539, opened 2024-01-24
Swap button


maikReal:
[PRODUCT TASK](https://www.notion.so/blockscout/Swap-button-e69a8122362b45a6a206a36c40bb9bd9?pvs=4)

## What do? 

Support the Swap button with additional analytical events on the header
Design: https://www.figma.com/file/4In0X8UADoZaTfZ34HaZ3K/Blockscout_2.0?type=design&node-id=15028:86613&mode=design&t=yUYgK1TBkc5Gh3SG-1

All detailed information inside the product task

ulyanas:
@maxaleks @maikReal 
-Can the button be hidden via config?
- Button should be hidden by default in the version we distribute.

maikReal:
@ulyanas By default the value of the env will be empty. It means the button will be hidden for all instances

PR #1569, opened 2024-02-02
Add swap button

maxaleks:
Resolves: https://github.com/blockscout/frontend/issues/1539

Added a new env var `NEXT_PUBLIC_SWAP_BUTTON_URL`


lib/router/removeQueryParam.ts
```diff
@@ -0,0 +1,7 @@
+import type { NextRouter } from 'next/router';
+
+export default function removeQueryParam(router: NextRouter, param: string) {
+  const { pathname, query } = router;
+  delete router.query[param];
```
tom2drum, line 5: I am not sure if it is a good idea to mutate the Nextjs router. Can we just call `router.replace()` with new query and let Nextjs do its job?

ui/marketplace/useWalletConnection.tsx
```diff
@@ -0,0 +1,39 @@
+import { useRouter } from 'next/router';
+import { useEffect, useRef } from 'react';
+
+import removeQueryParam from 'lib/router/removeQueryParam';
+import updateQueryParam from 'lib/router/updateQueryParam';
+import useWallet from 'ui/snippets/walletMenu/useWallet';
+
+export default function useMarketplaceWallet() {
```
tom2drum, line 8:
```suggestion
export default function useWalletConnection() {
```

tom2drum: maybe I would rename it into `useConnectWalletAction`, `useAutoConnectWallet` or something similar

lib/router/removeQueryParam.ts
tom2drum: the same comment to `lib/router/updateQueryParam.ts`

tom2drum (changes requested):

tom2drum (approved):

PR #1572, opened 2024-02-05
fix(tx): update tx details degraded result set to success

luchenqun:

update `TxDetailsDegraded` result set to success, due to here (first line)

https://github.com/blockscout/frontend/blob/fe81cfe66ce8f80419e1399bc18832abf03713a5/ui/tx/details/TxInfo.tsx#L91-L97

### DIFF
--- a/ui/tx/TxDetailsDegraded.tsx
+++ b/ui/tx/TxDetailsDegraded.tsx
@@ -95,7 +95,7 @@ const TxDetailsDegraded = ({ hash, txQuery }: Props) => {
         created_contract: txReceipt?.contractAddress ?
           { ...unknownAddress, hash: txReceipt.contractAddress, is_contract: true } :
           null,
-        result: '',
+        result: 'success',
         priority_fee: null,
         tx_burnt_fee: null,
         revert_reason: null,


ui/tx/TxDetailsDegraded.tsx
```diff
@@ -95,7 +95,7 @@ const TxDetailsDegraded = ({ hash, txQuery }: Props) => {
         created_contract: txReceipt?.contractAddress ?
           { ...unknownAddress, hash: txReceipt.contractAddress, is_contract: true } :
           null,
-        result: '',
+        result: 'success',
```
tom2drum, line 98: I guess we need to add an extra condition then:
- either here and check what status is present
- or in [TxInfo](https://github.com/blockscout/frontend/blob/c2d398ecd55115f1ec70cd11557e6f95dd3ebb9b/ui/tx/details/TxInfo.tsx#L91) component

luchenqun: I think there is no problem. Once the request is successful, the result should be `'success'`, otherwise the returned data will be null.
https://github.com/blockscout/frontend/blob/fe81cfe66ce8f80419e1399bc18832abf03713a5/ui/tx/details/TxInfo.tsx#L74-L76

tom2drum: Yes, if there is no transaction receipt, we cannot determine the final status of the transaction. In this case, the `status` field will be null, but the `result` will still be `success`. I believe this is a mismatch.

tom2drum:
@luchenqun, there has been no activity in this pull request for a month. Are you able to continue working on it, or should we close it?

PR #1573, opened 2024-02-05
use eslint check and format code

luchenqun:

use eslint check and format code, need install VS Code ESLint extension

https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,8 @@
 {
-    "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "editor.formatOnSave": false,
+  "eslint.enable": true,
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": "always"
+  }
 }


.vscode/settings.json
```diff
@@ -1,3 +1,8 @@
 {
-    "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "editor.formatOnSave": false,
```
tom2drum, line 3: Can we let the user decide? For instance, I prefer formatting the code when I save the file, rather than when I copy/paste it from another source.

luchenqun: In addition, in fact, this configuration is that once you save the code, it will be automatically formatted. The effective configuration is the last line `"source.fixAll.eslint": "always"`

tom2drum: Yes, we don't use Prettier in our project, all formatting is done by eslint. You should not run Prettier here.

The code formatting will happen automatically before every commit. But you can run it manually for all files with `yarn lint:eslint:fix` if needed.

My point was that not all users want to opt-out `editor.formatOnSave` option. We should let them decide when to run the formatter.

tom2drum:
@luchenqun, there has been no activity in this pull request for a month. Are you able to continue working on it, or should we close it?

Issue #1552, opened 2024-01-30
Support admin marketplace


maikReal:
## What to do? 

We need to integrate and use Blockcout admin endpoints to manage marketplace content: apps + all additional information

Endpoints info:
- Get the list of all apps: /api/v1/chains/1/marketplace/dapps
- Get info about a specific app by its id: /api/v1/chains/1/marketplace/dapps/test-dapp

NOTE: Instead of  `chains/1/` can be used necessary `chain_id`

RS task: https://github.com/blockscout/blockscout-rs/issues/703
Product task: https://www.notion.so/blockscout/Marketplace-s-dApps-managing-aee65e1b32ff47b7b477aee36618ca87?pvs=4

PR #1581, opened 2024-02-07
Fetch dapps from admin service

maxaleks:
Resolves https://github.com/blockscout/frontend/issues/1552


configs/app/features/marketplace.ts
```diff
@@ -7,21 +7,31 @@ import { getEnvValue, getExternalAssetFilePath } from '../utils';
...
 
 const title = 'Marketplace';
 
-const config: Feature<{ configUrl: string; submitFormUrl: string; categoriesUrl: string | undefined }> = (() => {
+const config: Feature<{
+  configUrl: string | undefined;
+  submitFormUrl: string;
+  categoriesUrl: string | undefined;
+  api: { endpoint: string; basePath: string } | undefined;
+}> = (() => {
```
tom2drum, line 19: Since the `configUrl` and `api` properties are mutually exclusive, I would suggest the following structure of the config:

```js
type Config = (
  { configUrl: string } | 
  { api: { endpoint: string; basePath: string } }
) & { submitFormUrl: string; categoriesUrl: string | undefined }
const config: Feature<Config> = ...
```

ui/marketplace/useMarketplaceApps.tsx
```diff
@@ -47,19 +47,36 @@ function sortApps(apps: Array<MarketplaceAppOverview>, isExperiment: boolean) {
...
+    pathParams: { chainId: config.chain.id },
+    queryOptions: {
+      select: (data) => sortApps(data as Array<MarketplaceAppOverview>, isExperiment),
+      placeholderData,
+      enabled: feature.isEnabled,
+    },
+  });
+
+  return feature.isEnabled && feature.configUrl ? data1 : data2;
+}
```
tom2drum, line 75: Can we have just one query and inside `queryFn` decide which resource to call?

tom2drum: And please place one hook to one file.

deploy/values/review/values.yaml.gotmpl
```diff
@@ -63,7 +63,6 @@ frontend:
     NEXT_PUBLIC_HOMEPAGE_CHARTS: "['daily_txs','coin_price','market_cap']"
     NEXT_PUBLIC_NETWORK_RPC_URL: https://rpc.ankr.com/eth_goerli
     NEXT_PUBLIC_NETWORK_EXPLORERS: "[{'title':'Bitquery','baseUrl':'https://explorer.bitquery.io/','paths':{'tx':'/goerli/tx','address':'/goerli/address','token':'/goerli/token','block':'/goerli/block'}},{'title':'Etherscan','logo':'https://github.com/blockscout/frontend-configs/blob/main/configs/explorer-logos/etherscan.png?raw=true','baseUrl':'https://goerli.etherscan.io/','paths':{'tx':'/tx','address':'/address','token':'/token','block':'/block'}}]"
-    NEXT_PUBLIC_MARKETPLACE_CONFIG_URL: https://raw.githubusercontent.com/blockscout/frontend-configs/dev/configs/marketplace/eth-goerli.json
```
tom2drum, line 66: Should we add API url instead of this static config file then?

ui/marketplace/useMarketplaceApps.tsx
```diff
@@ -52,8 +51,17 @@ export default function useMarketplaceApps(filter: string, selectedCategoryId: s
...
 
   const { isPlaceholderData, isError, error, data } = useQuery<unknown, ResourceError<unknown>, Array<MarketplaceAppOverview>>({
-    queryKey: [ 'marketplace-apps' ],
-    queryFn: async() => apiFetch(configUrl, undefined, { resource: 'marketplace-apps' }),
+    queryKey: [ 'marketplace-dapps' ],
+    queryFn: async() => {
+      if (!feature.isEnabled) {
+        return [];
+      }
+      const url = 'configUrl' in feature ?
```
tom2drum, line 59: I believe it is better to register new resources in the `lib/api/resources.ts` for admin API, and then either call a regular fetch or an API fetch:


```js
const fetch = useFetch();
...
return feature.configUrl ? fetch(feature.configUrl): apiFetch('marketplace_dapps');
...
```

tom2drum (changes requested):
By referring to this guide at https://github.com/blockscout/frontend/blob/main/docs/CONTRIBUTING.md#adding-new-env-variable, please make the following updates:
-  Include a new variable in the documentation.
-  Add the suitable values for this variable to the dev presets file and CI config files.
-  Mention this change in the PR description.

tom2drum (approved):

Issue #1603, opened 2024-02-12
Show social info for protocols


maikReal:
[PRODUCT TASK](https://www.notion.so/blockscout/Social-info-about-the-project-b1f50c9377f748cba4490c830ae7110d?pvs=4)

## What to do? 

We need to add additional information inside the app. All details inside the product task

**Laucnh for all users with AB testing** 

Design: https://www.figma.com/file/4In0X8UADoZaTfZ34HaZ3K/Blockscout_2.0?type=design&node-id=18013%3A99991&mode=design&t=r4EZ6FjlF0tnoLR3-1

maikReal:
@maxaleks I think you will get info from the marketplace DB (admin panel). That's why I created a task for the admin panel to support new **"Discord"** field in the database and changed the type format of **"github"** field from stroing to array

Here is the task: https://github.com/blockscout/blockscout-rs/issues/776

maikReal:
@maxaleks Strange layouts when you open the mobile platform 
@tgladilina Can review this one as well. Seems we need to move buttons to the left side, near an arrow 





![Image](https://github.com/blockscout/frontend/assets/26190360/94281417-552f-4e18-817f-a12edcf63a79)

tgladilina:
@maxaleks please move buttons to the left side

![Image](https://github.com/blockscout/frontend/assets/103045386/c01aa177-fd6d-4e0e-b39d-ebed39b64e2e)

maikReal:
@maxaleks All looks good for now

PR #1635, opened 2024-02-21
Add a top bar to the dapp page

maxaleks:
Resolves https://github.com/blockscout/frontend/issues/1603

### DIFF
--- a/ui/marketplace/MarketplaceAppAlert.tsx
+++ b/ui/marketplace/MarketplaceAppAlert.tsx
@@ -10,7 +10,7 @@ type Props = {
   internalWallet: boolean | undefined;
 }
 
-const MarketplaceAppTopBar = ({ internalWallet }: Props) => {
+const MarketplaceAppAlert = ({ internalWallet }: Props) => {
   const { address } = useMarketplaceWallet();
   const isWalletConnected = Boolean(address);
 
@@ -34,7 +34,6 @@ const MarketplaceAppTopBar = ({ internalWallet }: Props) => {
   return (
     <Alert
       status={ message.status }
-      width={{ base: '100%', md: 'auto' }}
       borderRadius="base"
       px={ 3 }
       py={{ base: 3, md: 1.5 }}
@@ -53,4 +52,4 @@ const MarketplaceAppTopBar = ({ internalWallet }: Props) => {
   );
 };
 
-export default MarketplaceAppTopBar;
+export default MarketplaceAppAlert;
--- a/ui/marketplace/MarketplaceAppAlert.tsx
+++ b/ui/marketplace/MarketplaceAppAlert.tsx
@@ -1,0 +1,56 @@
+import { Alert } from '@chakra-ui/react';
+import React from 'react';
+
+import type { IconName } from 'ui/shared/IconSvg';
+import IconSvg from 'ui/shared/IconSvg';
+
+import useMarketplaceWallet from './useMarketplaceWallet';
+
+type Props = {
+  internalWallet: boolean | undefined;
+}
+
+const MarketplaceAppTopBar = ({ internalWallet }: Props) => {
+  const { address } = useMarketplaceWallet();
+  const isWalletConnected = Boolean(address);
+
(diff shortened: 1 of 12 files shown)


ui/marketplace/MarketplaceAppTopBar.tsx
```diff
@@ -0,0 +1,106 @@
...
+      iconColor = 'green.600';
+      bgColor = 'green.100';
+      text = 'Your wallet is connected with Blockscout';
+    } else if (isWalletConnected) {
+      icon = 'integration/partial';
+      text = 'Connect your wallet in the app below';
+    }
+
+    return { icon, iconColor, bgColor, text };
+  }, [ isWalletConnected, data?.internalWallet ]);
```
tom2drum, line 49: This way is okay, but it would be more readable to have a function called `renderAlert` that renders the target component instead of just retrieving its props and then use them "x" line of codes below. Or it could be moved to a separate component, for example.

Also, can we use the standard `Alert` instead of just `Flex`?

ui/marketplace/MarketplaceAppTopBar.tsx
```diff
@@ -0,0 +1,106 @@
...
+  if (isPending) {
+    return (
+      <Flex alignItems="center" mb={ 2 } gap={ 2 }>
+        <Skeleton w="25px" h="32px" borderRadius="base"/>
+        <Skeleton w="300px" h="32px" borderRadius="base"/>
+        <Skeleton w="75px" h="32px" borderRadius="base"/>
+        <Skeleton w="150px" h="32px" borderRadius="base"/>
+      </Flex>
+    );
+  }
```
tom2drum, line 60: The preferred approach is to enclose each element in the `Skeleton` component. This ensures that the loading layout remains consistent with the actual layout.

Note, that some of our components already implement the loading state. So you just have to pass the prop to them.

ui/marketplace/MarketplaceAppTopBar.tsx
```diff
@@ -0,0 +1,106 @@
...
+import IconSvg from 'ui/shared/IconSvg';
+import LinkExternal from 'ui/shared/LinkExternal';
+import LinkInternal from 'ui/shared/LinkInternal';
+
+import MarketplaceAppInfo from './MarketplaceAppInfo';
+
+type Props = {
+  isWalletConnected: boolean;
+  data: MarketplaceAppOverview | undefined;
+  isPending: boolean;
```
tom2drum, line 19: We typically use `isLoading` props throughout the app to display the loading state of components. It's best to maintain this naming convention.

ui/marketplace/MarketplaceAppAlert.tsx
```diff
@@ -0,0 +1,56 @@
...
+import type { IconName } from 'ui/shared/IconSvg';
+import IconSvg from 'ui/shared/IconSvg';
+
+import useMarketplaceWallet from './useMarketplaceWallet';
+
+type Props = {
+  internalWallet: boolean | undefined;
+}
+
+const MarketplaceAppTopBar = ({ internalWallet }: Props) => {
```
tom2drum, line 13:
```suggestion
const MarketplaceAppAlert = ({ internalWallet }: Props) => {
```

ui/marketplace/MarketplaceAppInfo/WebsiteLink.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <Link
+      href={ url }
+      target="_blank"
+      display="inline-flex"
+      alignItems="center"
+      columnGap={ 1 }
+      mt={ 3 }
+    >
+      <IconSvg name="link" boxSize={ 5 } color="text_secondary"/>
+      <span>{ (new URL(url)).hostname }</span>
```
tom2drum, line 24: Are we certain that the `url` prop contains a valid URL? Was it check somewhere previously for validity? If not, this expression could potentially break the entire page. Wrapping it in a `try {} catch` block can prevent this issue.

tom2drum (approved):

Issue #1634, opened 2024-02-21
Add new ad provider: Getit


ulyanas:
Please add one more option for ad providers:
Getit https://getittech.io/ 
Docs: https://github.com/Getit-Tech/getit-sdk 
Desktop: 729x90
Mobile: 270x90

PR #1647, opened 2024-02-26
Getit ad banner

isstuev:
resolves https://github.com/blockscout/frontend/issues/1634

### DIFF
--- a/ui/shared/ad/GetitBanner.tsx
+++ b/ui/shared/ad/GetitBanner.tsx
@@ -1,5 +1,5 @@
 import { Flex, chakra } from '@chakra-ui/react';
-import { GetitAdPlugin } from 'getit-sdk';
+import dynamic from 'next/dynamic';
 import React from 'react';
 import { useAccount } from 'wagmi';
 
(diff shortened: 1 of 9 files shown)


ui/shared/ad/GetitBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+import React from 'react';
+import { useAccount } from 'wagmi';
+
+import useIsMobile from 'lib/hooks/useIsMobile';
+
+import Web3ModalProvider from '../Web3ModalProvider';
+
+const GetitAdPlugin = dynamic(() => import('getit-sdk').then(module => module.GetitAdPlugin), { ssr: false });
+
+const GETIT_API_KEY = 'ZmGXVvwYUAW4yXL8RzWQHNKmpSyQmt3TDXsXUxqFqXPdoaiSSFyca3BOyunDcWdyOwTkX3UVVQel28qbjoOoWPxYVpPdNzbUNkAHyFyJX7Lk9TVcPDZKTQmwHlSMzO3a';
```
tom2drum, line 12: shouldn't it be configured via ENV variable?

tom2drum (approved):

ui/shared/ad/GetitBanner.tsx
isstuev: we don't configure api keys for ad providers bc there's no such case when user need a custom api key (yet)

Issue #1692, opened 2024-03-09
Numbers have spaces when read contract


ericlee42:
### Description

<img width="600" alt="image" src="https://github.com/blockscout/frontend/assets/92196970/fc221441-bd0e-4880-9fbd-ed53f2f459d7">

<img width="534" alt="image" src="https://github.com/blockscout/frontend/assets/92196970/6cfc2847-4415-438a-9c30-38a4809337ae">

I have tested it on local, the result is not correct, I think the bug it's related to the request data.

### Link to the page

https://explorer.holesky.metisdevops.link/address/0xdE8d56212118906a0CeCD331e842429714b4c47B?tab=read_proxy

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### App version

v1.26.0-alpha

### Operating system

None

ericlee42:
I tried main branch, but the api returns an errror

```
[...]

PR #1693, opened 2024-03-10
fix number input value

Phinx765:
solve #1692 
form submit number value will contains ' ' if thousandSeparator set to it

## Checklist for PR author


tom2drum:
Sorry, we can't accept this change because we want to display separators in the form fields for a better user experience. It's best to remove extra spaces from form values in the form submit handler instead.

Issue #1779, opened 2024-04-03
Re-work header and tooltips in the marketplace


maikReal:
**[PRODUCT TASK](https://www.notion.so/blockscout/Re-work-an-apps-header-and-tooltips-8fc206e4683340d8b33e2c92f2f528d3?pvs=4)**

## What to do?

We need to re-work the user experience of working with marketplace dapps and make it more convenient. Also, we need to re-work our current tooltips and hints in the marketplace and inside apps

All detailed information can be found in **[the product task](https://www.notion.so/blockscout/Re-work-an-apps-header-and-tooltips-8fc206e4683340d8b33e2c92f2f528d3?pvs=4)**

PR #1822, opened 2024-04-15
Rework marketplace app header and tooltips

maxaleks:

Resolves https://github.com/blockscout/frontend/issues/1779

## Checklist for PR author


lib/contexts/marketplace.tsx
```diff
@@ -0,0 +1,35 @@
...
+export function MarketplaceContextProvider({ children }: Props) {
+  const router = useRouter();
+  const [ isAutoConnectDisabled, setIsAutoConnectDisabled ] = React.useState(false);
+
+  React.useEffect(() => {
+    setIsAutoConnectDisabled(false);
+  }, [ router.pathname ]);
+
+  return (
+    <MarketplaceContext.Provider value={{ isAutoConnectDisabled, setIsAutoConnectDisabled }}>
```
tom2drum, line 27: I think, to avoid unnecessary re-rendering, it is better to memoize the value.

ui/marketplace/MarketplaceAppTopBar.tsx
```diff
@@ -25,7 +28,7 @@ type Props = {
   securityReport?: MarketplaceAppSecurityReport;
 }
 
-const MarketplaceAppTopBar = ({ data, isLoading, isWalletConnected, securityReport }: Props) => {
```
tom2drum, line 28: If the `isWalletConnected` prop is not used anymore, please delete it.

ui/marketplace/MarketplaceAppTopBar.tsx
```diff
@@ -85,6 +70,25 @@ const MarketplaceAppTopBar = ({ data, isLoading, isWalletConnected, securityRepo
             { getHostname(data?.url) }
           </chakra.span>
         </LinkExternal>
+        <Skeleton isLoaded={ !isLoading }>
+          <MarketplaceAppInfo data={ data }/>
+        </Skeleton>
+        { (isExperiment && (securityReport || isLoading)) && (
```
tom2drum, line 76:
```suggestion
        { (isExperiment && securityReport) && (
```

I assume it should be enough to check only the presence of the `securtiyReport` data.

ui/shared/UserAvatar.tsx
```diff
@@ -34,7 +35,7 @@ const UserAvatar = ({ size }: Props) => {
       boxSize={ `${ size }px` }
       borderRadius="full"
       overflow="hidden"
-      fallback={ isImageLoadError || !data?.avatar ? <IconSvg name="profile" boxSize={ 5 }/> : undefined }
+      fallback={ isImageLoadError || !data?.avatar ? <IconSvg name="profile" boxSize={ `${ fallbackIconSize }px` }/> : undefined }
```
tom2drum, line 38: Should the image and its fallback be the same size?

ui/snippets/walletMenu/WalletIdenticon.tsx
```diff
@@ -0,0 +1,47 @@
...
+};
+
+const WalletIdenticon = ({ address, isAutoConnectDisabled, className }: Props) => {
+  const { themedBackgroundOrange } = useMenuButtonColors();
+  const isMobile = useIsMobile();
+
+  return (
+    <Box className={ className } position="relative">
+      <AddressIdenticon size={ 20 } hash={ address }/>
+      { isAutoConnectDisabled && (
```
tom2drum, line 23: Please consider adding a test to cover this functionality.

The same comment about the new code in `ui/snippets/walletMenu/WalletMenuContent.tsx`

ui/marketplace/MarketplaceAppTopBar.tsx
maxaleks: `securityReport` is `undefined` while the loading is in progress, so in this case everything is correct here 🙂

ui/shared/UserAvatar.tsx
maxaleks: No, they're different sizes in this task 🤷‍♂️

ui/snippets/walletMenu/WalletIdenticon.tsx
maxaleks: Agreed to do it in a separate task

tom2drum (approved):

PR #1875, opened 2024-05-01
Contract interaction improvements

tom2drum:

- Add contract method id tag 
- Simulate write method call
- Fix bug with WEI checkbox

## Checklist for PR author


ui/address/contract/ABI/utils.ts
```diff
@@ -0,0 +1,7 @@
+export const getNativeCoinValue = (value: unknown) => {
```
isstuev, line 1: is it specific for the contract values only or mb move it to /lib folder?

isstuev (approved):

ui/address/contract/ABI/utils.ts
tom2drum: it is used only in ABI form submit handler for parsing value of native coin value input in order to pass it to wagmi in correct format

Issue #1820, opened 2024-04-15
Action button on Blockscout pages


maikReal:
[**PRODUCT TASK**](https://www.notion.so/blockscout/Action-button-for-dapps-on-the-transaction-page-6bf9efb23c964836ac6d6634c2d63e99?pvs=4)

## What to do?
Support action buttons on different Blockscout pages that will help to quickly redirect a user to our marketplace. Launch as an AB experiment. All info in the product task

Blocked by:
- https://github.com/orgs/blockscout/projects/6/views/1?pane=issue&itemId=59026437 
Need to sync with @tom2drum and @maxaleks to understand the next steps

maikReal:
**UPD 17.04:** 

The design is finalized @maxaleks

PR #1883, opened 2024-05-03
Action button

maxaleks:

Resolves https://github.com/blockscout/frontend/issues/1820

## Checklist for PR author


ui/shared/ActionButton/ActionButton.tsx
isstuev, line 1: maybe use less generic name like AppActionButton?

ui/shared/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+import LinkExternal from '../LinkExternal';
+
+type Props = {
+  data: NonNullable<AddressMetadataTagFormatted['meta']>;
+  className?: string;
+  txHash?: string;
+  source: 'Txn' | 'NFT collection' | 'NFT item';
+}
+
+const ActionButton = ({ data, className, txHash, source }: Props) => {
```
tom2drum, line 20: Let's make the name more specific. For example, prepend it with "Marketplace".

ui/shared/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+      bg={ bgColor || defaultBg }
+      _hover={{ bg: bgColor, opacity: 0.9 }}
+      _active={{ bg: bgColor, opacity: 0.9 }}
+    >
+      { content }
+    </Button>
+  ) : (
+    <LinkExternal
+      className={ className }
+      href={ actionURL }
```
isstuev, line 69: is it possible that there is neither appID nor actionURL? we handle that case in the click handler. doesn't look correct to have a link without a href.

ui/token/TokenDetails.tsx
```diff
@@ -27,13 +31,17 @@ interface Props {
 const TokenDetails = ({ tokenQuery }: Props) => {
   const router = useRouter();
   const isMounted = useIsMounted();
+  const { value: isExperiment } = useFeatureValue('action_button_exp', false);
```
isstuev, line 34: let's rename isExperiment to isActionButtonExperiment

ui/shared/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+        borderRadius="sm"
+        mr={ 2 }
+      />
+      <Text fontSize="sm" fontWeight="500" color="currentColor">
+        { text }
+      </Text>
+    </>
+  );
+
+  return appID ? (
```
tom2drum, line 50: I think here we should be more strict in displaying this component. What if I have a random public tag for a token address which has some sort of `actionURL` in metadata but it is not actually a marketplace or dapp URL? Should the button still be visible? Because we will already show a tag with this link in the page header.

ui/shared/ActionButton/useActionData.tsx
```diff
@@ -0,0 +1,11 @@
+import useAddressMetadataInfoQuery from 'lib/address/useAddressMetadataInfoQuery';
+
+export default function useActionData(address: string | undefined = '') {
+  const { data } = useAddressMetadataInfoQuery([ address ]);
```
tom2drum, line 4: Let's memoize the passed array of addresses here.

ui/token/TokenNftMarketplaces.tsx
```diff
@@ -1,16 +1,23 @@
...
+import ActionButton from 'ui/shared/ActionButton/ActionButton';
 import DetailsInfoItem from 'ui/shared/DetailsInfoItem';
+import TextSeparator from 'ui/shared/TextSeparator';
 
 interface Props {
   hash: string | undefined;
   id?: string;
   isLoading?: boolean;
+  actionData?: AddressMetadataTagFormatted['meta'];
+  source: 'NFT collection' | 'NFT item';
```
tom2drum, line 16: Strictly speaking, this property is only required for experiments. Therefore, shouldn't it be optional?

ui/token/TokenNftMarketplaces.tsx
```diff
@@ -1,16 +1,23 @@
...
+  isExperiment?: boolean;
 }
 
-const TokenNftMarketplaces = ({ hash, id, isLoading }: Props) => {
+const TokenNftMarketplaces = ({ hash, id, isLoading, actionData, source, isExperiment }: Props) => {
```
tom2drum, line 20: Adding tests to cover various component states is recommended.

ui/shared/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const actionURL = data.actionURL?.replace('{chainId}', config.chain.id || '').replace('{txHash}', txHash || '');
+
+  const handleClick = React.useCallback(() => {
+    const info = appID || actionURL;
+    if (info) {
+      mixpanel.logEvent(mixpanel.EventTypes.PAGE_WIDGET, { Type: 'Action button', Info: info, Source: source });
+    }
+  }, [ source, appID, actionURL ]);
+
+  const content = (
```
isstuev, line 35: logo url and text are optional in types. either change types or provide a fallback

ui/shared/ActionButton/useActionData.tsx
tom2drum: Also all empty strings should be filtered out from array.

tom2drum (changes requested):

ui/shared/ActionButton/ActionButton.tsx
maxaleks: The check is done in the hook, so one of the parameters should not be null: https://github.com/blockscout/frontend/pull/1883/files#diff-559281ad921d1747ee6b36479238f8501539199a414110b7df149851f9edb13fR7

ui/token/TokenNftMarketplaces.tsx
maxaleks: The experiment is temporary and with high probability will be used permanently later, and the component is used only where it is necessary to pass this parameter, so I don't think it should be made optional

ui/shared/ActionButton/ActionButton.tsx
maxaleks: What kind of fallback should I provide? If `logoURL` or `text` are not defined, they will simply not be displayed

ui/shared/ActionButton/ActionButton.tsx
isstuev: component doesn't know about the hook

maxaleks: Ok, I'll make it return `null` in this case 🙂

ui/shared/AppActionButton/useAppActionData.tsx
```diff
@@ -0,0 +1,14 @@
+import { useMemo } from 'react';
+
+import useAddressMetadataInfoQuery from 'lib/address/useAddressMetadataInfoQuery';
+
+export default function useAppActionData(address: string | undefined = '') {
```
tom2drum, line 5: we should check if it is not a hash from placeholder data

tom2drum (approved):

Issue #1776, opened 2024-04-03
Add a lightning label to the side menu


maikReal:
[**PRODUCT TASK**](https://www.notion.so/blockscout/Add-a-Lightning-label-to-the-side-menu-5fb46b6994ac408783e7c46e6adab242?pvs=4)

## What to do?

We need to add a new label to a side menu. More info in a product task

PR #1910, opened 2024-05-13
Menu lightning labels

maxaleks:

Resolves https://github.com/blockscout/frontend/issues/1776

Adds a new env var `NEXT_PUBLIC_MENU_LIGHTNING_LABELS`

## Checklist for PR author


public/icons/name.d.ts
```diff
@@ -71,6 +71,7 @@
 	| "integration/full"
 	| "integration/partial"
 	| "key"
+	| "lightning_sidebar"
```
tom2drum, line 74:
```suggestion
	| "lightning_navbar"
```

This will be better, since soon we will have a horizontal nav bar as an option of the page layout.

ui/snippets/navigation/LightningLabel.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const color = React.useMemo(() => {
+    if (isCollapsed || (!isExpanded && isLgScreen)) {
+      return (bgColor && bgColor !== 'transparent') ? bgColor : themeBgColor;
+    }
+    return 'transparent';
+  }, [ bgColor, themeBgColor, isCollapsed, isExpanded, isLgScreen ]);
+
+  return (
+    <IconSvg
+      className="lightning-label"
```
tom2drum, line 23: Let's make it a `const` and export from the file to reference in other places.

ui/snippets/navigation/useLightningLabel.ts
```diff
@@ -0,0 +1,16 @@
+import type { NavItem } from 'types/client/navigation-items';
+
+import config from 'configs/app';
+import { isInternalItem } from 'lib/hooks/useNavItems';
+
+export default function useLightningLabel(item: NavItem | Array<NavItem> | Array<Array<NavItem>>) {
```
tom2drum, line 6: Why it is made as a hook and not as a regular utility function?

tom2drum: And I guess the external function is redundant.

tom2drum (approved):

JOY:
It could be great if you guys can add an example/image for the new/changed env for every update.

tom2drum:
@JOY example values you always can find in the `./docs/ENVS.md` file
Not sure about the images though, we will think about it. Maybe we will the examples to the same file where it is applicable.

Issue #1381, opened 2023-11-30
Token icon (stats section homepage)


tgladilina:
**Current view**

![Image](https://github.com/blockscout/frontend/assets/103045386/9a54fb72-657f-4c73-adfe-c0776803bff0)

**To do**

- If network has token, we have to display it in the graph "price tab"
- Price changes for a 24 hours

Source ?

![Image](https://github.com/blockscout/frontend/assets/103045386/82412b62-ff82-4a56-a13c-bae4ce29473a)

ulyanas:
Need to create a task for backend @vbaranov

NikitaSavik:
Backend complete in v6.1.0

NikitaSavik:
Todo: list all places to update the icon

NikitaSavik:
This icon to be shown on:

- Homepage (graph button)
- Gas tracker page

isstuev:
waiting for https://github.com/blockscout/blockscout/issues/9658

PR #1932, opened 2024-05-17
Fe 1381 main

isstuev:

resolves https://github.com/blockscout/frontend/issues/1381

## Checklist for PR author


ui/shared/entities/token/TokenEntity.tsx
```diff
@@ -167,3 +168,16 @@ export {
...
+type NativeIconProps = Omit<IconProps, 'token'> & { src?: string | null };
+
+const NATIVE_TOKEN_DATA = {
+  name: config.chain.currency.name || '',
+  symbol: '',
+  address: '',
+  type: 'ERC-20' as const,
+};
+
+export const NativeIcon = (props: NativeIconProps) => {
```
tom2drum, line 181: The component is better placed in its own file and named `NativeTokenIcon` if it is primarily used independently from `TokenEntity`. Currently, it is unclear from the contexts where `NativeIcon` is used what it actually represents.

tom2drum (commented):
We should add a mock for this icon to some tests to ensure it will be displayed in the UI.

ui/shared/entities/token/TokenEntity.tsx
isstuev: I think it won't be clear why we use TokenEntity for the icon and why we need that "native token data", if I move it to a separate file.

tom2drum: Maybe we don't have to use `TokenIcon` then. Could be just a regular `Icon` with `TokenPlaceholder` fallback.

tom2drum (approved):

Issue #1860, opened 2024-04-25
'Certified' contracts


NikitaSavik:
in the same fashion as we do for tokens:

![Image](https://github.com/blockscout/frontend/assets/1701762/fb6e65a3-03f4-40c1-914d-ebd8c2a77e44)

tooltip: _This contract has been certified by the chain developers_

add checkmark icon to:
- search results
- all transaction lists
- contract page
- verified contracts page

[Related backend task](https://github.com/orgs/blockscout/projects/8/views/1?filterQuery=contract&pane=issue&itemId=58515371)

isstuev:
0xdC9D5762a7F3274753B91F087351ec25A3b9A5Ad
0x793EEdff1F64dFaE65E47bb310c1bd89d889eF94

PR #1940, opened 2024-05-20
add certified conract label

isstuev:

resolves https://github.com/blockscout/frontend/issues/1860

## Checklist for PR author


ui/token/TokenPageTitle.tsx
```diff
@@ -98,7 +98,7 @@ const TokenPageTitle = ({ tokenQuery, addressQuery, hash }: Props) => {
       { verifiedInfoQuery.data?.tokenAddress && (
         <Tooltip label={ `Information on this token has been verified by ${ config.chain.name }` }>
           <Box boxSize={ 6 }>
-            <IconSvg name="verified_token" color="green.500" boxSize={ 6 } cursor="pointer"/>
+            <IconSvg name="certified" color="green.500" boxSize={ 6 } cursor="pointer"/>
```
tom2drum, line 101: Can we use the `ui/shared/ContractCertifiedLabel.tsx` component here?

tom2drum (approved):

ui/token/TokenPageTitle.tsx
isstuev: it has a different tooltip text. I would not join them, logically they have nothing in common.

Issue #1955, opened 2024-05-27
Add possibility to change 'ERC' to something else


NikitaSavik:
To support custom naming, like BEP, TRC lets introduce new ENV, which will change naming on the front-end only

All token standards should be renamed with this change, e.g. ERC20 -> TRC20, ERC721 -> TRC721 etc

PR #1961, opened 2024-05-28
Add possibility to change ERC to something else

isstuev:

resolves https://github.com/blockscout/frontend/issues/1955

### Proposed Changes
added NEXT_PUBLIC_NETWORK_TOKEN_STANDARD_NAME variable to configure the display name of a token creation standard

## Checklist for PR author


tom2drum (approved):
Consider adding some tests to cover new functionality.

isstuev:
> Consider adding some tests to cover new functionality.

I don't think we need tests here, the uncovered logic is only reading env from variable. all the rest is covered because even the default ERC name now comes from the config (and we have tests with token types), not from the API directly.

Issue #2006, opened 2024-06-10
Arbitrum views


NikitaSavik:
[Specs](https://www.notion.so/blockscout/Arbitrum-rollup-specific-requirements-85dcd6da939847eba866f7ce339103d1?pm=c)
[API changes](https://www.notion.so/blockscout/API-endpoints-changes-26f1e93fdd424d89944c454200912fa8)

TLDR:

- Add Batches page
- Add Batch page
- Add Deposits (L1 -> L2) page
- Add Withdrawals (L2 -> L1) page
- Modify transaction page
- Modify block page
- Modify home page

PR #2024, opened 2024-06-13
Arbitrum views

isstuev:

resolves https://github.com/blockscout/frontend/issues/2006

### Proposed Changes
added 'arbitrum' option to the NEXT_PUBLIC_ROLLUP_TYPE variable

## Checklist for PR author


ui/blocks/BlocksContent.tsx
```diff
@@ -23,9 +23,11 @@ const TABS_HEIGHT = 88;
 interface Props {
   type?: BlockType;
   query: QueryWithPagesResult<'blocks'>;
+  showSocketInfo?: boolean;
```
tom2drum, line 26:
```suggestion
  enableSocket?: boolean;
```

since it is not only responsible for showing the info but also allowing to get data from the socket.

ui/messages/ArbitrumL2Messages.tsx
```diff
@@ -0,0 +1,102 @@
+import { Hide, Show, Skeleton } from '@chakra-ui/react';
```
tom2drum, line 1: To keep the folder structure consistent with `/deposits`, `/withdrawals`, and others, I suggest placing the component in the `/arbitrum` sub-folder.

ui/messages/ArbitrumL2MessagesListItem.tsx
```diff
@@ -0,0 +1,120 @@
...
+            isLoading={ isLoading }
+            hash={ l1TxHash }
+            fontSize="sm"
+            lineHeight={ 5 }
+            truncation="constant_long"
+          />
+        ) : (
+          <chakra.span color="text_secondary">
+            N/A
+          </chakra.span>
```
tom2drum, line 112: I believe by default `ListItemMobileGrid.Value` has `text_secondary` color. So this style is unnecessary. 

And the upper too.

tom2drum (commented):
- I think the status should be formatted to Proper-case
image

ui/messages/ArbitrumL2Messages.tsx
isstuev: we don't have other types with messages, so I think it's unnecessary here

tom2drum (approved):

Issue #2016, opened 2024-06-11
Combine security score and discovery views


maikReal:
[**PRODUCT TASK**](https://www.notion.so/blockscout/Re-work-security-score-view-5284e5b994a742d9ab02333979508263?pvs=4)

## What to do? 

We need to combine 2 different views into one, so we can show security info directly on cards
All detailed info in the product task

This is the blocker for the following task: https://github.com/blockscout/frontend/issues/2015

PR #2031, opened 2024-06-18
Combine default and security score view on the Marketplace page

maxaleks:

Resolves https://github.com/blockscout/frontend/issues/2016

## Checklist for PR author


ui/marketplace/AppSecurityReport.tsx
```diff
@@ -44,23 +54,43 @@ const AppSecurityReport = ({ id, securityReport, showContractList, isLoading, on
...
           onlyIcon={ onlyIcon }
-          label="The security score is based on analysis of a DApp's smart contracts."
+          label={ <>The security score is based on analysis<br/>of a DApp{ apos }s smart contracts.</> }
+          className={ className }
         />
       </PopoverTrigger>
       <PopoverContent w={{ base: '100vw', lg: '328px' }}>
         <PopoverBody px="26px" py="20px" fontSize="sm">
+          <Text fontWeight="500" fontSize="xs" mb={ 2 } variant="secondary">Smart contracts info</Text>
+          <Flex alignItems="center" justifyContent="space-between" h="32px">
```
tom2drum, line 77: Is the hard-coded height of the element necessary here?

ui/marketplace/MarketplaceAppCard.tsx
```diff
@@ -92,93 +95,96 @@ const MarketplaceAppCard = ({
...
             alignItems="center"
-            justifyContent="center"
+            justifyContent="space-between"
+            marginTop="auto"
+          >
+            <Link
+              fontSize="sm"
+              fontWeight="500"
+              paddingRight={{ md: 2 }}
+              href="#"
```
tom2drum, line 154: I think that if you don't provide the `href` prop at all, it will not lead anywhere in the app. So, it is redundant here, no?

ui/marketplace/MarketplaceAppTopBar.tsx
```diff
@@ -27,7 +27,8 @@ type Props = {
 }
 
 const MarketplaceAppTopBar = ({ data, isLoading, securityReport }: Props) => {
-  const [ showContractList, setShowContractList ] = useBoolean(false);
+  const [ isContractListShown, setIsContractListShown ] = useBoolean(false);
+  const [ contractListType, setContractListType ] = React.useState(ContractListTypes.ALL);
```
tom2drum, line 31: Maybe it could be combined into just one state `contractListType` with an enum or undefined value.

ui/shared/sort/ButtonDesktop.tsx
```diff
@@ -0,0 +1,57 @@
...
+        variant="outline"
+        onClick={ onClick }
+        color={ primaryColor }
+        fontWeight="600"
+        borderColor="transparent"
+        px={ 2 }
+        data-selected={ isMenuOpen }
+      >
+        <Box
+          as={ isMenuOpen ? 'div' : 'span' }
```
tom2drum, line 39: Can you explain the logic behind different types of html-elements when the menu is opened and not?

tom2drum: The same question about the `transition` property.

ui/shared/sort/Sort.tsx
```diff
@@ -1,58 +1,70 @@
...
-          )) }
-        </MenuOptionGroup>
-      </MenuList>
-    </Menu>
+    <Popover isOpen={ isOpen } onClose={ onClose } placement="bottom-start" isLazy>
+      <PopoverTrigger>
+        { isMobile ? (
+          <SortButtonMobile isActive={ isOpen || Boolean(value) } onClick={ onToggle } isLoading={ isLoading }/>
+        ) : (
+          <SortButtonDesktop isMenuOpen={ isOpen } isLoading={ isLoading } onClick={ onToggle }>
```
tom2drum, line 50: Maybe, similar to `SortButtonMobile` component, this prop should be named `isActive`

```suggestion
          <SortButtonDesktop isActive={ isOpen } isLoading={ isLoading } onClick={ onToggle }>
```

ui/marketplace/AppSecurityReport.tsx
maxaleks: This is the height specified in the design. I can change it to `py={ 1.5 }`, wdyt?

ui/marketplace/MarketplaceAppCard.tsx
maxaleks: It turned out that without `href` the styles were not applied correctly in this case. We decided to leave it as it is for now

ui/shared/sort/ButtonDesktop.tsx
maxaleks: It is made in such a way to apply the correct styles, because there are 2 different text components in the button

tom2drum (approved):

Issue #1655, opened 2024-02-29
Support arrays of structs/tuples in contract method result


yvaskov:
## Description

Currently we parse arrays of tuples incorrectly, example:

https://unreal.blockscout.com/address/0x747c6530dC1e2dfD19978955119da93703b1f746?tab=read_proxy
Method 5 `getAllPair`, args:
```
0x1111110000000000000000000000000000000000
6
0
```

![Image](https://github.com/blockscout/frontend/assets/72267126/a40ef768-ad3e-4d9d-95c6-7e23eb80d930)

Need to add support for this type of results. It would look like this:
```
*pair1 (tuple):
    version (uint8) : 1
    pair_address (address): 0x7bae255618b0e5783a81e2abc82219dd62bebb23,
    box_address (address): 0x9675c1a6ec968c2ab16ac7c04f9cad5a9c8d023a,
    ...
*pair2 (tuple):
    version (uint8) : 1
    pair_address (address): 0x6145dbad9b8afc226d66d29df35b9228c4763b67,
    box_address (address): 0xdb477e5dc1c36a168e324a086ae65c35bfd35052,
    ...
```

## Version
[v1.25.0](https://github.com/blockscout/frontend/tree/v1.25.0)

tom2drum:
wait for https://github.com/blockscout/blockscout/issues/9558

PR #2032, opened 2024-06-18
Call contract read methods from the client

### LINKED ISSUE #2001: Call contract read methods from the client
**Dev:**
- Remove calls to API to the following resource: `/api/v2/smart-contracts/:hash/methods-read`, `/api/v2/smart-contracts/:hash/methods-read-proxy`, `/api/v2/smart-contracts/:hash/query-read-method`, `/api/v2/smart-contracts/:hash/methods-write`, `/api/v2/smart-contracts/:hash/methods-write-proxy`
- Get contract ABI from `/api/v2/smart-contracts/:hash` and display contract methods in the corresponding tabs; for proxy-contract also requests ABI if its implementation(s)
- [x] If there is no public RPC node for a blockchain, the client should make all requests via the Blockscout proxy node (need an example)

**Design:**
- [x] For methods that return a constant value show a "Retry" button when a query fails. Also, add a button to re-request the displayed value.
- [x] In the proxy tabs, we need a selector for contract implementations. This will allow users to choose methods from the desired contract.

**Testing:**
- [ ] Deploy a contract where we can test a major part of all corner cases (e.g with complex method arguments structure)

tom2drum:

Resolves #2001, #1655

## Checklist for PR author


ui/address/contract/methods/form/resultPublicClient/ItemPrimitive.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { matchInt } from '../utils';
+import ItemLabel from './ItemLabel';
+import { printRowOffset } from './utils';
+
+function castValueToString(value: unknown): string {
+  switch (typeof value) {
+    case 'string':
+      return value === '' ? `""` : value;
+    case 'boolean':
+      return String(value);
```
isstuev, line 20: it's default case

isstuev (approved):

PR #2091, opened 2024-07-11
Block Countdown page

### LINKED ISSUE #2068: Block Countdown page
Separate page to show countdown timer

- Separate route for the page `block/countdown/{blocknum}`
- [API method example](https://eth.blockscout.com/api?module=block&action=getblockcountdown&blockno=21000000)
- If search is numbers-only and Blocks not found - show url to the countdown page 

[Design](https://www.notion.so/blockscout/3ce714e723c24c2e935932f3d186637e?v=c46a9c6d9ae24217b8f0f020ef408c1f&p=b3c80097a68942bea1ed777406c8573a&pm=c)

tom2drum:

Resolves #2068

## Checklist for PR author


lib/api/resources.ts
```diff
@@ -852,6 +852,9 @@ export const RESOURCES = {
   graphql: {
     path: '/api/v1/graphql',
   },
+  block_countdown: {
```
isstuev, line 855: maybe it shouldn't be "block countdown" or we can specify query parameters here?

isstuev (approved):

isstuev (approved):

lib/api/resources.ts
tom2drum: It is only for typescript purposes. All resources in the old API share the same path, but this way we can define response types more granularly. Since API is deprecated I don't want to add more logic here now.

PR #2122, opened 2024-07-27
Fix: using `===` to check `NEXT_PUBLIC_PROMOTE_BLOCKSCOUT_IN_TITLE`

cheng404:
The `NEXT_PUBLIC_PROMOTE_BLOCKSCOUT_IN_TITLE` env is of type `boolean`. It is better to use `===` for strict comparison.

### DIFF
--- a/configs/app/meta.ts
+++ b/configs/app/meta.ts
@@ -4,7 +4,7 @@ import { getEnvValue, getExternalAssetFilePath } from './utils';
 const defaultImageUrl = '/static/og_placeholder.png';
 
 const meta = Object.freeze({
-  promoteBlockscoutInTitle: getEnvValue('NEXT_PUBLIC_PROMOTE_BLOCKSCOUT_IN_TITLE') || 'true',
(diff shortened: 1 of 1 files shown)


configs/app/meta.ts
```diff
@@ -4,7 +4,7 @@ import { getEnvValue, getExternalAssetFilePath } from './utils';
 const defaultImageUrl = '/static/og_placeholder.png';
 
 const meta = Object.freeze({
-  promoteBlockscoutInTitle: getEnvValue('NEXT_PUBLIC_PROMOTE_BLOCKSCOUT_IN_TITLE') || 'true',
+  promoteBlockscoutInTitle: getEnvValue('NEXT_PUBLIC_PROMOTE_BLOCKSCOUT_IN_TITLE') === 'true',
```
tom2drum, line 7: I assume this change will break the default behavior of the variable. If the variable is not passed, the value should be `true`. However, I agree that the current code has a bug. The proper fix would be:
```suggestion
  promoteBlockscoutInTitle: getEnvValue('NEXT_PUBLIC_PROMOTE_BLOCKSCOUT_IN_TITLE') === 'false' ? false : true,
```

tom2drum (changes requested):

tom2drum (approved):

Issue #2255, opened 2024-09-19
Zora: implement custom tag


NikitaSavik:
Address/Contract page needs a custom, frontend-based tag similar to what we did to Farcaster

- Get the response from https://api.zora.co/discover/user/[0x_address]
- If `user_profile.username` is present: draw a tag on the first place, with URL to zora.co/[username]

![Image](https://github.com/user-attachments/assets/7edc4baf-d0a4-4134-ac39-766c56a0ac5e)

PR #2263, opened 2024-09-24
Zora: implement custom tag

isstuev:

resolves https://github.com/blockscout/frontend/issues/2255

### Proposed Changes
added NEXT_PUBLIC_ADDRESS_USERNAME_TAG env variable
config for Zora: `{'api_url_template': 'https://api.zora.co/discover/user/{address}', 'tag_link_template': 'httpszora.co/{username}', 'tag_icon': 'https://raw.githubusercontent.com/blockscout/frontend-configs/main/configs/network-icons/zora.svg', 'tag_bg_color': 'rgba(0,0,0)', 'tag_text_color': 'rgba(255,255,255)'}`

## Checklist for PR author

### DIFF
--- a/configs/app/features/addressProfileAPI.ts
+++ b/configs/app/features/addressProfileAPI.ts
@@ -1,0 +1,45 @@
+import type { Feature } from './types';
+import type { AddressProfileAPIConfig } from 'types/client/addressProfileAPIConfig';
+
+import { getEnvValue, parseEnvJson } from '../utils';
+
+const value = parseEnvJson<AddressProfileAPIConfig>(getEnvValue('NEXT_PUBLIC_ADDRESS_USERNAME_TAG'));
+
+function checkApiUrlTemplate(apiUrlTemplate: string): boolean {
+  try {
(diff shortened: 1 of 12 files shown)


tom2drum (commented):
I find the current naming a bit misleading.

When I see the `userProfile` prefix, I immediately think of the user account and its profile within the app. However, that is not the case here.

I suggest making the names more specific by including the word `address`, as it relates to the address page (the tag on the address page) and omit `API` from the ENV name, since the config that it holds is not related only to the API but to the tag as well. For example, we could use `NEXT_PUBLIC_ADDRESS_USERNAME_TAG` or something similar for ENV variable name and change the naming in the code accordingly.

Additionally, I wonder if this feature is less generic than it appears. The API's response schema is hardcoded, and I doubt that other protocols will adopt the same schema as Zora. Therefore, we could be more specific in both the code and the naming to indicate that it pertains to the Zora API. However, this is just my perspective, and I do not have all the details about the task or whether other protocols will follow the same API schema.

PR #2285, opened 2024-10-08
V1351

isstuev:


abubakarfarooq13:
Screenshot 2024-10-08 163012
 why it's showing white on my side

isstuev:
> Screenshot 2024-10-08 163012 why it's showing white on my side

Hi! I don't know exactly what you're trying to run, but the console error shows that there was an inline-script blocked by csp policy. Don't think it relates to the PR, if you still have the problem please create an issue with the detailed description of your case.

PR #2323, opened 2024-10-25
Update proxy names

k1rill-fedoseev:

Update displayed proxy names

## Checklist for PR author


ui/address/contract/ContractCodeProxyPattern.tsx
```diff
@@ -44,18 +44,18 @@ const PROXY_TYPES: Record<NonNullable<SmartContractProxyType>, {
     link: 'https://github.com/wighawag/clones-with-immutable-args',
   },
   master_copy: {
-    name: 'GnosisSafe',
+    name: 'Safe proxy',
```
rimrakhimov, line 47: I'm voting for the second variant with "standard" or "pattern" word at the end (e.g., 'Safe proxy standard' or 'Safe proxy pattern'). Imho, "This proxy contract is detected via Safe proxy pattern" would be more descriptive. And also, the same goes with Compound protocol, imo

k1rill-fedoseev: What do you think about EIP-based proxies? I've decided to keep "Safe proxy" for consistency, because we don't write "EIP-1167 pattern". I think if we decide to use word "pattern", then it should be added for all variants.

rimrakhimov: I agree. "EIP-1167 pattern" looks fine for me

rimrakhimov: Though, in general, I'm not sure, that the word "pattern" is necessary in case of EIPs, even if we decide to use it for non-EIP patterns. In my understanding, the word "EIP" implies by itself that it is a standard, while "Safe proxy" is not

tom2drum (approved):

PR #2452, opened 2024-12-08
Fix read/write contract tooltip and stubs

aiden069:

* fix read/write contract tooltip unit
* chart widget unit
* stubs name, symbol and decimals

### DIFF
--- a/stubs/noves/NovesTranslate.ts
+++ b/stubs/noves/NovesTranslate.ts
@@ -1,7 +1,8 @@
 import type { NovesResponseData, NovesClassificationData, NovesRawTransactionData } from 'types/api/noves';
+import config from 'configs/app';
 
 const NOVES_TRANSLATE_CLASSIFIED: NovesClassificationData = {
-  description: 'Sent 0.04 ETH',
+  description: `Sent 0.04 ${ config.chain.currency.name }`,
   received: [ {
     action: 'Sent Token',
     actionFormatted: 'Sent Token',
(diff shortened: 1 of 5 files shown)


stubs/noves/NovesTranslate.ts
```diff
@@ -10,9 +11,9 @@ const NOVES_TRANSLATE_CLASSIFIED: NovesClassificationData = {
     to: { name: '', address: '0xa0393A76b132526a70450273CafeceB45eea6dEE' },
     token: {
       address: '',
-      name: 'ETH',
-      symbol: 'ETH',
-      decimals: 18,
+      name: config.chain.currency.name ?? '',
+      symbol: config.chain.currency.symbol ?? '',
+      decimals: config.chain.currency.decimals,
```
tom2drum, line 16: We never show data from stubs to the user; they are solely used as placeholders for rendering skeletons correctly. Therefore, we would like to keep this data static without any configuration.

ui/shared/chart/ChartWidget.pw.tsx
```diff
@@ -25,7 +26,7 @@ const props: Props = {
   ],
   title: 'Native coin circulating supply',
   description: 'Amount of token circulating supply for the period',
-  units: 'ETH',
+  units: currencyUnits.ether,
```
tom2drum, line 29: This data is returned by the Stats microservice and cannot be configured through the frontend app's environment variables. Therefore, this change is irrelevant.

ui/address/contract/methods/form/resultPublicClient/ItemPrimitive.tsx
```diff
@@ -54,7 +55,7 @@ const ItemPrimitive = ({ abiParameter, data, level, hideLabel }: Props) => {
     if (intMatch && typeof data === 'bigint' && intMatch.max > INT_TOOLTIP_THRESHOLD && data > INT_TOOLTIP_THRESHOLD) {
       const dividedValue = BigNumber(data.toString()).div(WEI);
       return (
-        <Tooltip label={ dividedValue.toLocaleString() + ' ETH' }>
+        <Tooltip label={ dividedValue.toLocaleString() + ' ' + currencyUnits.ether }>
```
tom2drum, line 58: Generally, this value does not represent an amount of native currency. It could be any arbitrary number, such as the "total supply" of a token. So, I think using the native currency symbol here would be incorrect.

tom2drum (changes requested):

Issue #2515, opened 2025-01-10
Unify stats across relevant pages


ulyanas:
The task is to unify the implementation of stats across all relevant pages

#### Counters and graphs
- **Endpoint:** `/api/v1/pages/main`
Fetches counters and graphs data for the main page.

#### Counters
- **Endpoint:** `/api/v1/pages/transactions`
Provides transaction counters, including pending and daily transaction statistics.

#### Counters
- **Endpoint:** `/api/v1/pages/contracts`
Retrieves statistics for verified contracts, including new and total contracts.

[Swagger](https://github.com/blockscout/blockscout-rs/blob/c4e8527abb628589e97e34c58d1698315e5dc709/stats/stats-proto/swagger/stats.swagger.yaml)

bragov4ik:
Should have an option to switch between "operational transactions" and "regular transactions" charts on the frontend.
Stats service now always returns corresponding fields on the main page handle.
It's possible that the arbitrum charts are disabled on the microservice, which will result in these fields being `null`:
```
    "total_operational_transactions": null,
    "yesterday_operational_transactions": null,
    ...
    "daily_new_operational_transactions": null
```

Also I published 2.4.0-beta npm package with stats types

bragov4ik:
Consequently, [the swagger was also updated](https://github.com/blockscout/blockscout-rs/blob/f42a961d86d27660e046db9cddd3cdce42a5e242/stats/stats-proto/swagger/stats.swagger.yaml)

PR #2561, opened 2025-02-07
Unify stats across relevant pages

isstuev:

resolves https://github.com/blockscout/frontend/issues/2515

### Proposed Changes
added daily_operational_txs option to NEXT_PUBLIC_HOMEPAGE_CHARTS and total_operational_txs to NEXT_PUBLIC_HOMEPAGE_STATS. Values can be used only if NEXT_PUBLIC_STATS_API_HOST is provided

## Checklist for PR author


ui/home/indicators/utils/getIndicatorValues.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    if (statsApiData) {
+      return indicator?.value(statsApiData);
+    }
+
+    return 'N/A';
+  })();
+
+  // we have diffs only for coin adn second coin price charts that get data from stats api
```
tom2drum, line 20:
```suggestion
  // we have diffs only for coin and second coin price charts that get data from stats api
```

ui/verifiedContracts/VerifiedContractsCounters.tsx
```diff
@@ -3,37 +3,59 @@ import React from 'react';
...
   });
 
-  if (!countersQuery.data) {
+  if (!(isStatsFeatureEnabled ? countersStatsQuery.data : countersApiQuery.data)) {
     return null;
   }
 
+  const isLoading = isStatsFeatureEnabled ? countersStatsQuery.isPlaceholderData : countersApiQuery.isPlaceholderData;
+
+  const contractsCount = isStatsFeatureEnabled ? countersStatsQuery.data?.total_contracts?.value : countersApiQuery.data?.smart_contracts;
```
tom2drum, line 32: Shouldn't we always fall back to our regular API response, regardless of whether microservices data is enabled? For example, if there is an error in the response from the stats microservices, should we still display the information from our base API?

ui/txs/TxsStats.tsx
```diff
@@ -6,13 +6,23 @@ import useApiQuery from 'lib/api/useApiQuery';
...
+    },
+  });
+
+  const txsStatsApiQuery = useApiQuery('txs_stats', {
     queryOptions: {
-      placeholderData: TXS_STATS,
+      enabled: !isStatsFeatureEnabled,
+      placeholderData: !isStatsFeatureEnabled ? TXS_STATS : undefined,
     },
   });
```
tom2drum, line 27: The same comment about fallback applies here. 
I suggest abstracting these two hooks into a separate one that will orchestrate all requests and return the `isLoading` flag and `data` in a common format. This way, all necessary logic will be encapsulated in one place rather than be spreaded in the view component.

ui/verifiedContracts/VerifiedContractsCounters.tsx
isstuev: I don't want to fetch both resources where it's not necessary. We assume that if stats service was configured it should work as expected.

PR #2573, opened 2025-02-14
Update package.json version to match latest release

nathanhleung:

We vendor the code of the Blockscout frontend on our side without storing the `.git` folder or any commit/tag data, so it's helpful to know what version the frontend is without that

### Proposed Changes

Sets `version` in `package.json` to match the version of the latest release on the releases page

https://github.com/blockscout/frontend/releases

Assuming this field has never really been used before since it's always been 1.0.0?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "blockscout-frontend",
-  "version": "1.0.0",
+  "version": "1.37.6",
   "private": false,
   "homepage": "https://github.com/blockscout/frontend#readme",
   "engines": {


isstuev:
we don't serve blockscout as an npm package, we use GitHub releases and tags for versioning, so version in package.json is redundant

nathanhleung:
> we don't serve blockscout as an npm package, we use GitHub releases and tags for versioning, so version in package.json is redundant

Understood. Is there any way to determine the currently running version of the blockscout frontend if we didn't end up storing any Git tags in our deployment?

We have a very very old version of the frontend up on one of our older chains but don't know what version it is so not sure whether we'll break anything by upgrading

PR #2636, opened 2025-03-13
fix: correctly handle token.decimals when value is 0

### LINKED ISSUE #2635: TokenSelectItem: Incorrect fallback to 18 for token.decimals when value is 0
### Description

Title:
TokenSelectItem: Incorrect fallback to 18 for token.decimals when value is 0

Description:
In ui/address/tokenSelect/TokenSelectItem.tsx (around line 23), the code uses:

const tokenDecimals = Number(data.token.decimals) || 18;

Since 0 is falsy in JavaScript, if data.token.decimals is 0 (which is a valid value for some tokens), the expression incorrectly falls back to the default value 18. This results in erroneous calculations for tokens that actually have 0 decimals.

Proposed Fix:
To handle this case correctly, you can modify the code to only apply the default value when data.token.decimals is null or undefined, like so:

const tokenDecimals = data.token.decimals != null ? Number(data.token.decimals) : 18;

This change will ensure that if data.token.decimals is 0, it will be correctly used in the calculations.

Additional Context:
	•	File: ui/address/tokenSelect/TokenSelectItem.tsx
	•	Version: Blockscout Frontend v1.38.1
	•	Issue: Dropdown calculations are based on the wrong token decimals value when token.decimals is 0.

### Link to the page

https://explorer.oasys.games/address/0xCE1527F5650763b386921Dc6aF924Cd30E99b731?tab=tokens

### Steps to reproduce

1. Go to 'Address details'
2. see Token dropdown
3. see Asset Table Quantity

### App version

v1.38.1

### Browser

any

### Operating system

macOS

### Additional information

_No response_

girafferz:
Previously, using "Number(data.token.decimals) || 18" caused a valid token.decimals value of 0 to fall back to 18. This commit updates the logic to correctly use 0 when defined, ensuring accurate calculations for tokens with 0 decimals.

This pull request addresses a bug in the file `ui/address/tokenSelect/TokenSelectItem.tsx`. When `data.token.decimals` is 0 (a valid value for some ERC20 tokens), the current logic incorrectly falls back to 18 because 0 is falsy in JavaScript. This change updates the logic to properly respect a 0 value, ensuring correct token amount calculations.

Fixes #2635 

### Proposed Changes
- Update the token decimals assignment in `ui/address/tokenSelect/TokenSelectItem.tsx`   
- This change ensures that if `data.token.decimals` is 0, it is used as the valid value rather than falling back to 18.
- No changes to any environment variables were made.

There are no breaking or incompatible changes introduced by this pull request. This is solely a bug fix for token decimals calculation.

### Additional Information
- The fix has been tested locally.
- All unit tests and Playwright tests pass.
- This change aligns with the ERC20 token specification, where a decimals value of 0 is valid.
- Screenshots or further details can be provided upon request.

## Checklist for PR author
- [x] I added tests to cover any new functionality, following this [guide](./CONTRIBUTING.md#writing--running-tests).
    - [ ] I updated the list of environment variables in the [documentation](ENVS.md).
    - [ ] I made the necessary changes to the validator script according to the [guide](./CONTRIBUTING.md#adding-new-env-variable).
    - [ ] I added "ENVs" label to this pull request.

### DIFF
--- a/ui/address/tokenSelect/TokenSelectItem.tsx
+++ b/ui/address/tokenSelect/TokenSelectItem.tsx
@@ -20,7 +20,7 @@ const TokenSelectItem = ({ data }: Props) => {
   const secondRow = (() => {
     switch (data.token.type) {
       case 'ERC-20': {
-        const tokenDecimals = Number(data.token.decimals) || 18;
+        const tokenDecimals = Number(data.token.decimals ?? 18);
         const text = `${ BigNumber(data.value).dividedBy(10 ** tokenDecimals).dp(8).toFormat() } ${ data.token.symbol || '' }`;
 
         return (
--- a/ui/shared/address/AddressAddToWallet.tsx
+++ b/ui/shared/address/AddressAddToWallet.tsx
@@ -24,7 +24,7 @@ function getRequestParams(token: TokenInfo, tokenId?: string): WatchAssetParams 
         options: {
           address: token.address,
           symbol: token.symbol || '',
-          decimals: Number(token.decimals) || 18,
+          decimals: Number(token.decimals ?? '18'),
           image: token.icon_url || '',
         },
       };


ui/address/tokenSelect/TokenSelectItem.tsx
```diff
@@ -20,7 +20,7 @@ const TokenSelectItem = ({ data }: Props) => {
   const secondRow = (() => {
     switch (data.token.type) {
       case 'ERC-20': {
-        const tokenDecimals = Number(data.token.decimals) || 18;
+        const tokenDecimals = Number(data.token.decimals ?? 18);
```
tom2drum, line 23: Can you describe the case when the current logic will not work?

girafferz: @tom2drum 
Yes, some tokens indeed have 0 decimals set in the contract, and in those cases, the fallback || 18 would incorrectly scale the values. Instead, using ?? 18 ensures that only null or undefined values default to 18, preserving the intended precision for tokens with explicitly defined 0 decimals.

girafferz:
```
export default function getCurrencyValue({ value, accuracy, accuracyUsd, decimals, exchangeRate }: Params) {
--
14 | const valueCurr = BigNumber(value).div(BigNumber(10 ** Number(decimals \|\| '18')));
```
 
 
 ```
 const decimals = Number(data.token.decimals || '18');
 ```
 
  There is no issue with handling decimals = '0' (as a string) in the current code because Number('0') === 0,


[ui/shared/address/AddressAddToWallet.tsx](https://github.com/blockscout/frontend/blob/165f289b41b20d06235192b141cd29271d80b883/ui/shared/address/AddressAddToWallet.tsx#L27)
“Indeed, this part also needs to be changed.”

girafferz:
I have made additional modification. Please review

tom2drum (approved):

PR #2816, opened 2025-06-27
Clusters Feature Integration

ayv8er:

This PR introduces Clusters Universal Name Service integration into the Blockscout explorer. Clusters is a cross-chain identity service that allows users to track multiple wallets across various networks under a single unified name (like `vitalik/` or `foobar/`).

This optional integration adds a new Clusters directory, Clusters leaderboard, individual cluster details page, and search capability. The feature follows established patterns for optional service and includes proper env handling.

### Proposed Changes

**Environment Variables:**
- Added two new environment variables:
    - `NEXT_PUBLIC_CLUSTERS_API_HOST` - for Clusters API endpoint
    - `NEXT_PUBLIC_CLUSTERS_CDN_URL` - for Clusters CDN assets
- Updated environment preset files in `configs/envs/.env.jest` and `configs/envs/.env.localhost`
- Added validation schema in `deploy/tools/envs-validator/schema.ts`
- Updated Helm deployment values in review YAML templates - `deploy/values/`

**Documentation and Configuration:**
- Added Clusters feature documentation to `docs/ENVS.md`
- Updated Jest configuration `jest.config.ts` and added test mocks `jest/mocks/` and `jest/setupAfterEnv.ts`

**API Configuration and Types:**
- Added `clustersApi` configuration in `configs/app/apis.ts` with optional Clusters API integration
- Created `lib/api/services/clusters.ts` for Clusters API service layer
- Added Clusters feature configuration in `configs/app/features/clusters.ts`
- Updated API resources and types in `lib/api/resources.ts` and `lib/api/types.ts`

**TypeScript Types and Stubs:**
- Created TypeScript type definitions in `types/api/clusters.ts`
- Added test stubs in `stubs/clusters.ts`
- Updated search types in `types/api/search.ts`

**Core Library and Utilities:**
- Added complete `lib/clusters/` directory with utilities for action bar, page utils, and input detection
- Created specialized hooks in `lib/hooks/`:
    - `useAddressClusters.ts` - for address specific cluster data
    - `useClusterPagination.ts` - for cluster pagination logic
    - `useClusterSearch.ts` - for cluster search functionality
    - `useClustersData.ts` - for main clusters data fetching
    - `useQueryParams.ts` - for URL query parameter handling
- Updated navigation in `lib/hooks/useNavItems.tsx`
- Updated metadata and analytics mappings for cluster pages in `lib/metadata/` and `lib/mixpanel/`

**Testing and QA:**
- Added comprehensive test coverage for all new utilities and hooks:
  - `lib/clusters/` utilities - action bar, page utils, input detection
  - All custom hooks - `useAddressClusters`, `useClusterPagination`, `useClusterSearch`, `useClustersData`
  - Search integration - `useSearchWithClusters`
  - UI components - `ClustersEntity`
- Updated Jest configuration and added testing infrastructure
- Added test mocks for Next.js routes, SVG imports, and React hooks

**Pages and Routing:**
- Added `pages/clusters/index.tsx` - main clusters directory page
- Added `pages/clusters/[name].tsx` - individual cluster details page
- Created corresponding page components:
    - `ui/pages/Clusters.tsx` - clusters directory page component
    - `ui/pages/Cluster.tsx` - individual cluster page component
- Updated Next.js routing types in `nextjs/nextjs-routes.d.ts`

**UI Components:**
- Clusters Directory/Leaderboard - `ui/clusters/`:
    - `ClustersActionBar.tsx` - search and filter controls
    - `ClustersDirectoryTable.tsx` and `ClustersDirectoryTableItem.tsx` - directory views
    - `ClustersDirectoryListItem.tsx` - directory list item
    - `ClustersLeaderboardTable.tsx` and `ClustersLeaderboardTableItem.tsx` - leaderboard views
    - `ClustersLeaderboardListItem.tsx` - leaderboard list items
- Individual Cluster Details - `ui/cluster/`:
    - `ClusterDetails.tsx` - main cluster detail component
- Address Integration - `ui/address/clusters/`:
    - `AddressClusters.tsx` - clusters tab for address pages
- Shared Components:
    - `ui/shared/ClusterIcon.tsx` - cluster icon component
    - `ui/shared/entities/clusters/ClustersEntity.tsx` - cluster entity display

**Search and Navigation Integration:**
- Added cluster search suggestions in `ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggestCluster.tsx`
- Created `ui/snippets/searchBar/useSearchWithClusters.tsx`
- Updated search utilities in `ui/shared/search/utils.ts`
- Updated navigation components in `ui/snippets/navigation/NavLinkIcon.tsx`
- Updated search bar components: `SearchBar.tsx`, `SearchBarInput.tsx`, `SearchBarSuggest.tsx`, `SearchBarSuggestItem.tsx`

**Assets and Icons:**
- Added `icons/clusters.svg` for cluster-specific logo use
- Updated icon type definitions in `public/icons/name.d.ts`

None. This is an additive feature that introduces optional clusters functionality. Existing installations will continue to work unchanged, and the new feature can be optionally enabled by setting the NEXT_PUBLIC_CLUSTERS_API_HOST environment variable.

### Additional Information

5 minute Loom video displaying the Clusters feature - https://www.loom.com/share/6278c920204249b398a6f227b1cb701d?sid=5b7d6a88-b13e-4002-8fab-0c531ac92090

## Checklist for PR author
- [X] I have tested these changes locally.
- [X] I added tests to cover any new functionality, following this [guide](./CONTRIBUTING.md#writing--running-tests)
- [X] Whenever I fix a bug, I include a regression test to ensure that the bug does not reappear silently.


ayv8er:
**Requesting Review From**: @tom2drum  @isstuev  as specified in the contributing guidelines.

I could not add reviewers or the ENVs label to the PR.

configs/app/features/index.ts
```diff
@@ -45,3 +45,4 @@ export { default as validators } from './validators';
 export { default as verifiedTokens } from './verifiedTokens';
 export { default as web3Wallet } from './web3Wallet';
 export { default as xStarScore } from './xStarScore';
+export { default as clusters } from './clusters';
```
tom2drum: Please move this line after the `csvExport` to maintain the exports in alphabetical order.

docs/ENVS.md
```diff
@@ -661,6 +661,17 @@ This feature allows name tags and other public tags for addresses.
 
 &nbsp;
 
+### Clusters Universal Name Service
```
tom2drum, line 676: Please add this section to the Table of Contents above on the page.

jest/mocks/nextjs-routes.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  route: jest.fn((opts) => {
+    if (opts?.pathname?.includes('[hash]')) {
+      return `/address/${ opts.query?.hash || 'test-hash' }`;
+    }
+    if (opts?.pathname?.includes('[name]')) {
+      return `/clusters/${ opts.query?.name || 'test-cluster' }`;
+    }
```
tom2drum: Not only these routes can have path params like `hash` or `name`, but also, for example, `/tx/[hash]`. Should the conditions be stricter then?

ui/address/clusters/AddressClusters.tsx
```diff
@@ -0,0 +1,116 @@
...
+const ClustersGrid = ({ data }: ClustersGridProps) => {
+  const itemsToShow = getClustersToShow(data, 10);
+  const numberOfRows = getGridRows(itemsToShow.length, 5);
+
+  return (
+    <Grid templateRows={ `repeat(${ numberOfRows }, auto)` } autoFlow="column" gap={ 4 } mt={ 2 }>
+      { itemsToShow.map((cluster) => (
+        <Flex key={ cluster.name } alignItems="center">
+          <ClustersEntity clusterName={ cluster.name } fontWeight={ 600 } noCopy/>
+        </Flex>
```
tom2drum: Is the wrapper `<Flex/>` component necessary here?

ui/shared/entities/clusters/ClustersEntity.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+const Icon = (props: IconProps) => {
+  if (props.noIcon) {
+    return null;
+  }
+
+  const styles = getIconProps(props.variant);
+
+  if (props.isLoading) {
+    return <Skeleton loading boxSize={ styles.boxSize } borderRadius="6px" mr={ 2 }/>;
```
tom2drum: Is it possible to use the standard border radius defined in `toolkit/theme/foundations/borders.ts`? (here and below)

ui/shared/entities/clusters/ClustersEntity.tsx
```diff
@@ -0,0 +1,189 @@
...
+      width={ styles.boxSize }
+      height={ styles.boxSize }
+      backgroundColor="#de6061"
+      borderRadius="6px"
+      mr={ 2 }
+      flexShrink={ 0 }
+    >
+      <IconSvg
+        name="clusters"
+        boxSize={ 6 }
```
tom2drum: I believe the image size should match that of its container.
```suggestion
        width="100%"
        height="100%"
```

ui/shared/entities/clusters/ClustersEntity.tsx
```diff
@@ -0,0 +1,189 @@
...
+  );
+
+  const profileImageElement = (
+    <Image
+      width={ styles.boxSize }
+      height={ styles.boxSize }
+      borderRadius="6px"
+      mr={ 2 }
+      flexShrink={ 0 }
+      src={ `${ (config.features.clusters as { cdnUrl?: string })?.cdnUrl || '' }/profile-image/${ props.clusterName }` }
```
tom2drum: Please use `getFeaturePayload` util instead of type coercion.

ui/shared/entities/clusters/ClustersEntity.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  const tooltipContent = (
+    <>
+      <Flex alignItems="center" textStyle="md">
+        <Box
+          display="inline-flex"
+          alignItems="center"
+          justifyContent="center"
+          boxSize={ 5 }
+          backgroundColor="#de6061"
```
tom2drum: It might be better to add this color to the theme - `toolkit/theme/foundations/colors.ts` - and use its name reference here and in `ClusterIcon` component.

ui/shared/entities/clusters/ClustersEntity.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  return (
+    <Tooltip
+      content={ tooltipContent }
+      variant="popover"
+      positioning={{
+        placement: 'bottom-start',
+      }}
+      contentProps={{
+        maxW: { base: '100vw', lg: '440px' },
```
tom2drum: All tooltips have overflow padding, so we need to reduce their `max-width` to prevent a shift in page width when the tooltip is open.

```suggestion
        maxW: { base: 'calc(100vw - 8px)', lg: '440px' },
```

ui/shared/entities/clusters/ClustersEntity.test.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+
+import { render, screen } from 'jest/lib';
+
+import ClustersEntity from './ClustersEntity';
+
+describe('ClustersEntity', () => {
```
tom2drum, line 7: We typically write visual tests for our components using the Playwright library. Please consider writing these tests for this component and the new pages for the clusters dashboard and cluster details. This will ensure that your feature remains intact if we decide to refactor or make any changes to our design guide.

ui/shared/ClusterIcon.tsx
```diff
@@ -0,0 +1,62 @@
+import { Box } from '@chakra-ui/react';
+import React from 'react';
+
+import config from 'configs/app';
+import { Image } from 'toolkit/chakra/image';
+import IconSvg from 'ui/shared/IconSvg';
+
+interface ClusterIconProps {
```
tom2drum: I believe it is better to extend the `ImageProps` interface and remove CSS-specific properties such as margins, flex, size, and so on. In the component itself, we can easily gather all these properties and spread them to the rendered component, overriding the default values when the user of this component needs to do so.

tom2drum: And instead of `size` it is easier to use the `boxSize` prop.

lib/clusters/detectInputType.ts
```diff
@@ -0,0 +1,22 @@
+export type InputType = 'address' | 'cluster_name';
+
+const EVM_ADDRESS_PATTERN = /^0x[a-fA-F0-9]{40}$/;
```
tom2drum: You can re-use `ADDRESS_REGEXP` here from `toolkit/components/forms/validators/address.ts`

lib/hooks/useClustersData.ts
```diff
@@ -0,0 +1,122 @@
...
+
+import type { ClustersByAddressObject } from 'types/api/clusters';
+import { ClustersOrderBy } from 'types/api/clusters';
+
+import useApiQuery from 'lib/api/useApiQuery';
+import { detectInputType } from 'lib/clusters/detectInputType';
+import { CLUSTER_ITEM } from 'stubs/clusters';
+
+export function useClustersData(debouncedSearchTerm: string, viewMode: string, orderBy: ClustersOrderBy, page: number) {
+  const itemsPerPage = 50;
```
tom2drum: It seems that the `orderBy` argument is unused in the function. Additionally, please use UPPER_SNAKE_CASE for variable names that hold constant values.

lib/hooks/useClustersData.ts
```diff
@@ -0,0 +1,122 @@
...
+        };
+      },
+    },
+  });
+
+  const clusterDetailsQuery = useApiQuery('clusters:get_cluster_by_id', {
+    queryParams: {
+      input: JSON.stringify({
+        id: (addressQuery.data?.result?.data?.[0] as ClustersByAddressObject & { clusterId?: string })?.clusterId || '',
+      }),
```
tom2drum: I believe here and below type coercions are not necessary.

lib/hooks/useClustersData.ts
```diff
@@ -0,0 +1,122 @@
...
+        Boolean((addressQuery.data?.result?.data?.[0] as ClustersByAddressObject & { clusterId?: string })?.clusterId)
+      ),
+    },
+  });
+
+  const currentQuery = React.useMemo(() => {
+    if (!showDirectoryView) return leaderboardQuery;
+    if (inputType === 'address') return addressQuery;
+    return directoryQuery;
+  }, [ showDirectoryView, inputType, leaderboardQuery, addressQuery, directoryQuery ]);
```
tom2drum: Minor comment: Since we are not passing this variable to another component or hook, memoization is not really necessary.

ui/pages/Clusters.tsx
```diff
@@ -0,0 +1,158 @@
...
+          ))
+        ) }
+      </Box>
+      <Box hideBelow="lg">
+        { showDirectoryView ? (
+          <ClustersDirectoryTable
+            data={ directoryData }
+            isLoading={ isLoading }
+            isClusterDetailsLoading={ isClusterDetailsLoading && inputType === 'address' }
+            top={ 80 }
```
tom2drum: I believe the height of the action bar should be 68px (no it is 72px). If the current height is not correct, it is better to use `ACTION_BAR_HEIGHT_DESKTOP` const instead here and below)

ui/clusters/ClustersDirectoryTable.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const ClustersDirectoryTable = ({ data, isLoading, top, isClusterDetailsLoading }: Props) => {
+  return (
+    <TableRoot>
+      <TableHeaderSticky top={ top }>
+        <TableRow>
+          <TableColumnHeader width="20%">Cluster Name</TableColumnHeader>
+          <TableColumnHeader width="30%">Address</TableColumnHeader>
+          <TableColumnHeader width="10%">Joined</TableColumnHeader>
```
tom2drum: Please consider to add `TimeFormatToggle` for this column.

ui/clusters/ClustersDirectoryTableItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+      </TableCell>
+      <TableCell verticalAlign="middle">
+        { item.owner && isEvmAddress(item.owner) && (
+          <AddressEntity address={{ hash: item.owner }} isLoading={ isLoading } fontWeight={ 500 }/>
+        ) }
+        { item.owner && !isEvmAddress(item.owner) && (
+          <Box display="inline-flex" alignItems="center" minWidth={ 0 }>
+            <AddressEntity address={{ hash: item.owner }} isLoading={ isLoading } fontWeight={ 500 } noLink={ true }/>
+          </Box>
+        ) }
```
tom2drum:
```suggestion
        { item.owner && (
          <AddressEntity address={{ hash: item.owner }} isLoading={ isLoading } fontWeight={ 500 } noLink={ !isEvmAddress(item.owner) }/>
        ) }
```

tom2drum: Same for mobile item.

ui/clusters/ClustersActionBar.tsx
```diff
@@ -0,0 +1,165 @@
...
+      </SegmentedButton>
+      <SegmentedButton
+        isSelected={ viewMode === 'directory' }
+        onClick={ handleDirectoryClick }
+        isDisabled={ viewToggleDisabled }
+      >
+        Directory
+      </SegmentedButton>
+    </HStack>
+  );
```
tom2drum: It might be better to use the `ButtonGroupRadio` component here. However, let me first check with our designer to determine which control would be appropriate.

ui/cluster/ClusterDetails.tsx
```diff
@@ -0,0 +1,99 @@
...
+interface Props {
+  clusterData?: ClusterByNameResponse['result']['data'];
+  clusterName: string;
+  isLoading: boolean;
+}
+
+const ClusterDetails = ({ clusterData, clusterName, isLoading }: Props) => {
+  if (!clusterData && !isLoading) {
+    return <div>Cluster not found</div>;
+  }
```
tom2drum: I believe it would be better to display our standard 404 error screen in this case.

ui/cluster/ClusterDetails.tsx
```diff
@@ -0,0 +1,99 @@
...
+        hint="The amount of ETH backing this cluster name"
+        isLoading={ isLoading }
+      >
+        Backing
+      </DetailedInfo.ItemLabel>
+      <DetailedInfo.ItemValue>
+        <Skeleton loading={ isLoading }>
+          { backingEth } ETH
+        </Skeleton>
+      </DetailedInfo.ItemValue>
```
tom2drum: I believe it is better to maintain consistent displays of native coin values across all views regarding precision and USD equivalent. Please look at how this is done for the transferred value in the transaction details view, for example.

ui/cluster/ClusterDetails.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+      <DetailedInfo.ItemLabel
+        hint="When this cluster name was created"
+        isLoading={ isLoading }
+      >
+        Created
+      </DetailedInfo.ItemLabel>
+      <DetailedInfo.ItemValue>
+        <Skeleton loading={ isLoading }>
+          { clusterData?.createdAt ? dayjs(clusterData.createdAt).format('llll') : 'N/A' }
```
tom2drum: Consider using the same component for displaying timestamps as for other detailed views - `DetailedInfoTimestamp`.

lib/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,22 @@
+import { useRouter } from 'next/router';
+import { useCallback } from 'react';
+
+export function useQueryParams() {
```
tom2drum: Please move it to the `lib/router` directory. Is the function of this hook similar to the `lib/router/updateQueryParam.ts` utility?

lib/hooks/useClustersData.ts
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
+
+import type { ClustersByAddressObject } from 'types/api/clusters';
+import { ClustersOrderBy } from 'types/api/clusters';
+
+import useApiQuery from 'lib/api/useApiQuery';
+import { detectInputType } from 'lib/clusters/detectInputType';
+import { CLUSTER_ITEM } from 'stubs/clusters';
+
+export function useClustersData(debouncedSearchTerm: string, viewMode: string, orderBy: ClustersOrderBy, page: number) {
```
tom2drum: It might be better to put all these new hooks related specifically to clusters feature into the `lib/clusters` folder, so all the utility functions for clusters will be in one place.

ui/snippets/searchBar/useSearchWithClusters.tsx
```diff
@@ -0,0 +1,97 @@
...
+      expires_at: cluster.expiresAt,
+      total_wei_amount: cluster.backingWei,
+      is_testnet: cluster.isTestnet,
+    },
+  };
+}
+
+export default function useSearchWithClusters() {
+  const quickSearch = useQuickSearchQuery();
+  const debouncedSearchTerm = useDebounce(quickSearch.searchTerm, 300);
```
tom2drum: The `useQuickSearchQuery` hook already returns the `debouncedSearchTerm`

ui/snippets/searchBar/useSearchWithClusters.tsx
```diff
@@ -0,0 +1,97 @@
...
+import useDebounce from 'lib/hooks/useDebounce';
+
+import useQuickSearchQuery from './useQuickSearchQuery';
+
+function isClusterSearch(term: string): boolean {
+  const trimmed = term.trim();
+  const hasTrailingSlash = trimmed.endsWith('/');
+  const looksLikeCluster = trimmed.includes('/') || hasTrailingSlash;
+
+  return looksLikeCluster && config.features.clusters.isEnabled;
```
tom2drum: My suggestion is to move condition `config.features.clusters.isEnabled` to the top of the function, so we will not do unnecessary work on instances where feature is disabled.

ui/snippets/searchBar/useSearchWithClusters.tsx
```diff
@@ -0,0 +1,97 @@
...
+      } as typeof quickSearch.query;
+    }
+
+    const clusterData = clusterQuery.data.result.data;
+    const transformedResults = [ transformClusterToSearchResult(clusterData, clusterData.owner) ];
+
+    return {
+      ...clusterQuery,
+      data: transformedResults,
+    } as typeof quickSearch.query;
```
tom2drum: An alternative solution is to use the `select` function of the `useQuery` hook to transform the data. In this case, all caching will be handled by the `react-query` library. However, you may want to construct a different `queryKey` to avoid conflicts with the same query on the cluster details page.

ui/snippets/searchBar/SearchBarInput.tsx
```diff
@@ -156,7 +156,7 @@ const SearchBarInput = (
       >
         <Input
           size="md"
-          placeholder={ isMobile ? 'Search by address / ... ' : 'Search by address / txn hash / block / token... ' }
+          placeholder={ isMobile ? 'Search by address / ... ' : 'Search by address / txn hash / block / token / cluster /... ' }
```
tom2drum: I assume we only need to change the placeholder if the feature is enabled.

lib/clusters/detectInputType.ts
```diff
@@ -0,0 +1,22 @@
...
+  const trimmedInput = input.trim();
+
+  if (EVM_ADDRESS_PATTERN.test(trimmedInput)) {
+    return 'address';
+  }
+
+  return 'cluster_name';
+}
+
+export function isEvmAddress(address: string): boolean {
```
tom2drum: Since it is a general-purpose utility function for any address, let's move it to the `lib/address` directory.

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggestItem.tsx
```diff
@@ -112,12 +119,19 @@ const SearchBarSuggestItem = ({ data, isMobile, searchTerm, onClick, addressForm
...
+      }
       case 'tac_operation': {
         return <SearchBarSuggestTacOperation data={ data } searchTerm={ searchTerm } isMobile={ isMobile } addressFormat={ addressFormat }/>;
       }
     }
   })();
 
+  if (data.type === 'cluster' && !isClusterClickable) {
+    return content;
+  }
```
tom2drum: Let's move the `isClusterClickable` condition closer to this statement and rename it to `hasLink` to make it more generic, as we might alter it for other purposes later.

```suggestion
  if (!hasLink) {
    return content;
  }
```

lib/clusters/detectInputType.ts
tom2drum: Or perhaps we can add this regular expression to `toolkit/utils/regexp.ts`.

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggestCluster.tsx
```diff
@@ -0,0 +1,79 @@
...
+    <Text
+      overflow="hidden"
+      whiteSpace="nowrap"
+      color="text.secondary"
+    >
+      <HashStringShortenDynamic hash={ hash } noTooltip/>
+    </Text>
+  );
+
+  const isContractVerified = data.is_smart_contract_verified && (
```
tom2drum: I guess the `is_smart_contract_verified` flag is always `false` for the cluster data.

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggestCluster.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <Flex
+          alignItems="center"
+          overflow="hidden"
+          gap={ 1 }
+          mt={{ base: 0, md: 0 }}
+        >
+          { address }
+          { isContractVerified }
+        </Flex>
+        <Box display={{ base: 'none', md: 'block' }}/>
```
tom2drum: If there is no data to display in the third column, could we consider removing it entirely?

ui/shared/ClusterIcon.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <Image
+      width={ size }
+      height={ size }
+      borderRadius={ borderRadius }
+      mr={ mr }
+      flexShrink={ flexShrink }
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      src={ `${ (config.features.clusters as any).cdnUrl }/profile-image/${ clusterName }` }
```
tom2drum: Is type coercion necessary here? If you are turning off the linter rules, please add a comment explaining why you need to do so.

tom2drum (changes requested):
Looks good! 👍 I left some comments on logic misconceptions, code style, and ways to improve readability.

deploy/tools/envs-validator/schema.ts
```diff
@@ -962,6 +962,9 @@ const schema = yup
     NEXT_PUBLIC_VISUALIZE_API_BASE_PATH: yup.string(),
     NEXT_PUBLIC_CONTRACT_INFO_API_HOST: yup.string().test(urlTest),
     NEXT_PUBLIC_NAME_SERVICE_API_HOST: yup.string().test(urlTest),
+    NEXT_PUBLIC_METADATA_SERVICE_API_HOST: yup.string().test(urlTest),
+    NEXT_PUBLIC_CLUSTERS_API_HOST: yup.string().test(urlTest),
+    NEXT_PUBLIC_CLUSTERS_CDN_URL: yup.string().test(urlTest),
```
tom2drum: Please explain why `NEXT_PUBLIC_METADATA_SERVICE_API_HOST` was added directly to this configuration list. I also suggest creating a separate sub-schema for the clusters feature and making the `NEXT_PUBLIC_CLUSTERS_CDN_URL` environment variable dependent on `NEXT_PUBLIC_CLUSTERS_API_HOST`, so users cannot pass the first without the latter.

configs/app/features/index.ts
ayv8er: will move clusters export in alphabetical order

docs/ENVS.md
ayv8er: thank you for catching this, will add

ui/address/clusters/AddressClusters.tsx
ayv8er: It is not necessary, it is a redundant wrapper

ui/shared/entities/clusters/ClustersEntity.tsx
ayv8er: yes, will update borderRadius to use standard values

ui/shared/entities/clusters/ClustersEntity.tsx
ayv8er: Agreed, appreciate you pointing that out!

ui/shared/ClusterIcon.tsx
ayv8er: Done!

deploy/tools/envs-validator/schema.ts
ayv8er: `NEXT_PUBLIC_METADATA_SERVICE_API_HOST` was added while resolving merge conflicts during a rebase with the latest `main` branch. Since it wasn’t part of the original scope, I preserved it in this list to avoid breaking any existing expectations, but happy to move it to a more appropriate schema if needed.

ayv8er: And that separate sub-schema for the clusters feature to make the `NEXT_PUBLIC_CLUSTERS_CDN_URL` environment variable dependent on `NEXT_PUBLIC_CLUSTERS_API_HOST` is a good idea. Will implement!

ui/shared/ClusterIcon.tsx
ayv8er: It was unnecessary, the disable line was removed.

lib/clusters/detectInputType.ts
ayv8er: Added that regex to `toolkit/utils/regexp.ts` and re-used in both detectInputType.ts and `validators/address.ts`

lib/hooks/useClustersData.ts
ayv8er: Indeed `orderBy` was unused, it's been removed from the hook and the component using it was also updated. 

Corrected variable name for constant value.

ui/pages/Clusters.tsx
ayv8er: Really appreciate the detailed review, this is done!

ui/clusters/ClustersDirectoryTable.tsx
ayv8er: Great addition, thank you.

lib/hooks/useQueryParams.ts
ayv8er: Makes sense to move to move `useQueryParams.ts` to `lib/router`.

The 2 hooks overlap, but the main reason I created `useQueryParams` hook was to support updating multiple params at once (which is useful for this integration).

Differentiators
- React hook vs plain util functions (Clusters integration only uses in React components)
- `router.push()` vs `router.replace()` (UX consideration)
- Param deletion logic for cleaner URLs
- No hash preservation

If `useQueryParams` hook included hash preservation, it could replace `updateQueryParam`

ui/clusters/ClustersActionBar.tsx
tom2drum: As I discussed with our designer, could you please adjust this control to align with our design guidelines? It should resemble the one on the contract page. As mentioned earlier, in our code, it is the `ButtonGroupRadio`. Additionally, please arrange it in a single row with the filter input on mobile. Thank you.

image

image

lib/hooks/useQueryParams.ts
tom2drum: Yeah, that totally makes sense. Let's keep the hook, and later we will refactor and remove the `updateQueryParam` utility.

deploy/tools/envs-validator/schema.ts
tom2drum: Oh, I see now.
That is because we recently moved the `NEXT_PUBLIC_METADATA_SERVICE_API_HOST ` variable into `addressMetadataSchema` sub-scheme. So there is no need now to include it in the main one.

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggestCluster.tsx
ayv8er: Yes, it should be removed.

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggestCluster.tsx
ayv8er: Ah yes, it's set to false in the `useSearchWithClusters` hook

ui/snippets/searchBar/SearchBarInput.tsx
ayv8er: Yes, so the conditional should be based on the clusters feature enablement. Will refactor.

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggestItem.tsx
ayv8er: Sounds good!

lib/hooks/useClustersData.ts
ayv8er: Better organization, it is done

ui/clusters/ClustersActionBar.tsx
ayv8er: Thank you for the pointers, I've refactored to your design guidelines and it looks much better ^^

**Web View**
Screenshot 2025-07-03 at 22 05 42

**Mobile View**
Screenshot 2025-07-03 at 22 06 09

jest/mocks/nextjs-routes.js
ayv8er: It should, and account for other's tests.

ui/shared/entities/clusters/ClustersEntity.test.tsx
ayv8er: Just pushed Playwright tests for this component and the newly added pages. Included docker generated screenshots as well.

ui/pages/Clusters.pw.tsx
```diff
@@ -0,0 +1,75 @@
...
+    );
+
+    await expect(component.getByText('campnetwork/lol').first()).toBeVisible();
+    await expect(component.getByText('duck/quack').first()).toBeVisible();
+    await expect(component.getByText('test/cluster').first()).toBeVisible();
+
+    await expect(component).toHaveScreenshot();
+  });
+
+  test('clusters directory empty state +@mobile', async({ render, page }) => {
```
tom2drum: I believe this test and the one below are unnecessary since the screenshots only include our standard components, such as PageTitle. The test for the component's default view with data should suffice for visual testing.

ui/pages/Clusters.pw.tsx
```diff
@@ -0,0 +1,75 @@
...
+import Clusters from './Clusters';
+
+test.beforeEach(async({ mockEnvs, mockTextAd }) => {
+  await mockEnvs(ENVS_MAP.clusters);
+  await mockTextAd();
+});
+
+test.describe('Clusters Directory Page', () => {
+  test('clusters directory with data @mobile', async({ render, page }) => {
+    await page.route('**/v1/trpc/names.search*', (route) => {
```
tom2drum: Is it possible to reuse the `mockApiResponse` fixtures here and avoid route matching using regex?
(Same comment to all tests)

ui/cluster/ClusterDetails.tsx
```diff
@@ -0,0 +1,109 @@
...
+  clusterData?: ClusterByNameResponse['result']['data'];
+  clusterName: string;
+  isLoading: boolean;
+}
+
+const ClusterDetails = ({ clusterData, clusterName, isLoading }: Props) => {
+  if (!clusterData && !isLoading) {
+    const error = new Error('Cluster not found');
+    error.cause = { status: 404 };
+    return <AppError error={ error }/>;
```
tom2drum: I think throwing an Error here should be ok.

```suggestion
    throw new Error('Cluster not found', { cause: { status: 404 } });
```

ui/shared/entities/clusters/ClustersEntity.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  const tooltipContent = (
+    <>
+      <Flex alignItems="center" textStyle="md">
+        <Box
+          display="inline-flex"
+          alignItems="center"
+          justifyContent="center"
+          boxSize={ 5 }
+          backgroundColor={ colors.clusters.value }
```
tom2drum: The standard method in Chakra UI for referencing a color is to use its semantic token.

```suggestion
          backgroundColor="clusters"
```

ui/shared/entities/clusters/ClustersEntity.pw.tsx
```diff
@@ -0,0 +1,166 @@
...
+      <ClustersEntity
+        clusterName={ testnetCluster.name }
+        href={ `/cluster/${ testnetCluster.name }` }
+      />,
+    );
+
+    await expect(component.getByText('test/cluster')).toBeVisible();
+
+    await expect(component).toHaveScreenshot();
+  });
```
tom2drum: What is the main purpose of these test cases? What scenarios do they cover, and how are they distinct from the ones mentioned above? 

Please keep the number of tests reasonably small. I am unsure if we need to test the loading state on both desktop and mobile devices; the same applies to truncation and other functionalities. If the behavior or appearance is not device-specific, let's keep only one test for desktop as the default project in Playwright. You can also set the viewport to a smaller value to reduce white space in the screenshots (refer to `ui/shared/entities/address/AddressEntity.pw.tsx` as an example).

ui/shared/ClusterIcon.tsx
```diff
@@ -0,0 +1,64 @@
...
+}: ClusterIconProps) => {
+  const clustersFeature = getFeaturePayload(config.features.clusters);
+
+  const fallbackElement = (
+    <Box
+      display="inline-flex"
+      alignItems="center"
+      justifyContent="center"
+      boxSize={ boxSize }
+      backgroundColor={ colors.clusters.value }
```
tom2drum: Same comment about color token.

ui/snippets/searchBar/useSearchWithClusters.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const combinedQuery = React.useMemo(() => {
+    if (!isClusterQuery) {
+      return quickSearch.query;
+    }
+
+    return {
+      ...clusterQuery,
+      data: clusterQuery.data || [],
+      isError: false,
```
tom2drum: Could you explain the reasoning behind why the `isError` flag should always be `false`?

tom2drum: Cloning the entire query object here doesn't seem like a good idea to me in terms of memory usage. I just want to ensure that it is truly necessary.

ui/snippets/searchBar/useSearchWithClusters.tsx
tom2drum: Reopen this conversation. The condition was deleted, but I believe it should remain either within the function or at the point where it is called.

ui/pages/Cluster.pw.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    await expect(component.getByText('test/cluster').first()).toBeVisible();
+    await expect(component.getByText('Cluster Name')).toBeVisible();
+    await expect(component.getByText('Owner address')).toBeVisible();
+    await expect(component.getByText('Backing')).toBeVisible();
+
+    await expect(component).toHaveScreenshot();
+  });
+
+  test('loading state +@mobile', async({ render }) => {
```
tom2drum: As mentioned above, there is no need to test the loading, empty, or error states for every page unless it has a custom UI elements in these scenarios.

ui/pages/Cluster.pw.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+import Cluster from './Cluster';
+
+test.beforeEach(async({ mockEnvs, mockTextAd }) => {
+  await mockEnvs(ENVS_MAP.clusters);
+  await mockTextAd();
+});
+
+test.describe('Cluster Details Page', () => {
+  test('cluster details campnetwork/lol +@mobile', async({ render, page }) => {
```
tom2drum: If the difference between the test cases is that the first tests an EVM cluster and the second tests a non-EVM cluster, this distinction should be reflected in the test case names. Currently, the test suite names don't indicate the differences between them.

ui/pages/Cluster.pw.tsx
```diff
@@ -0,0 +1,121 @@
...
+    await page.route('**/v1/trpc/names.get*', (route) => route.fulfill({
+      status: 200,
+      json: campNetworkClusterByName,
+    }));
+
+    const component = await render(
```
tom2drum, line 27: Please consider mocking the cluster icon with the `mockAssetResponse` fixture. This way, all external resource calls will be mocked, and the test will not fail if the external resource is down.

tom2drum (changes requested):

ui/snippets/searchBar/useSearchWithClusters.tsx
ayv8er: Just pushed optimization

ui/snippets/searchBar/useSearchWithClusters.tsx
ayv8er: **isError: false UI reason**
In [SearchBarSuggest](https://github.com/blockscout/frontend/blob/main/ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggest.tsx#L137), when `query.isError` is true, it displays "Something went wrong. Try refreshing the page or come back later."

For better UX where cluster service outages don't break the search experience, cluster search failures show "No results found" while regular search failures show the error message. This behavior is tested and intentional (see test line 415 in `useSearchWithClusters.test.tsx`).

**query object cloning**
The `UseQueryResult` interface has many properties (data, isError, isFetching, refetch, etc.) and needs to return an object matching `useQuickSearchQuery`'s interface since [SearchBar.tsx uses them interchangeably](https://github.com/blockscout/frontend/blob/main/ui/snippets/searchBar/SearchBar.tsx#L41).

The cloning preserves all other properties the UI depends on while overriding specific ones for cluster behavior. Memory impact should be minimal since it's a shallow spread of lightweight React Query metadata objects, only for cluster searches.

ui/shared/entities/clusters/ClustersEntity.pw.tsx
ayv8er: Thank you for the detailed feedback. You're right about the redundant test cases and excessive mobile annotations.

I'll remove the redundant data tests as they're only testing different string values and remove the mobile annotations since the component behavior is not device specific.

I appreciate the viewport settings tip to reduce whitespace in screenshots.

ui/pages/Cluster.pw.tsx
ayv8er: The difference between the 3 was 2 mainnet tests with different values and a testnet test.
I've updated the test names to be more clear.

ui/pages/Clusters.pw.tsx
```diff
@@ -0,0 +1,58 @@
...
+test.beforeEach(async({ mockEnvs, mockTextAd }) => {
+  await mockEnvs(ENVS_MAP.clusters);
+  await mockTextAd();
+});
+
+test.describe('Clusters Directory Page', () => {
+  test.describe('mobile', () => {
+    test.use({ viewport: devices['iPhone 13 Pro'].viewport });
+
+    test('clusters directory with data @mobile', async({ render, mockApiResponse, mockAssetResponse }) => {
```
tom2drum: I believe you missed the "+" sign here, so the screenshot for the "mobile" project was not generated. I apologize if the usage of tags like "+@mobile" or "+@dark-mode" is unclear in our contribution guide. I will add some amendments to clarify this later.

Since the component has different views for desktop and mobile devices, let's remove the viewport setting and correct the tag name so that the screenshot will be generated for each device type.

I also noticed that there are some previously generated screenshots for this test suite stored in the repository. Could you please clean them up?

`Clusters.pw.tsx_mobile_Clusters-Directory-Page-clusters-directory-empty-state-mobile-1.png`
`Clusters.pw.tsx_mobile_Clusters-Directory-Page-clusters-directory-with-data-mobile-1.png`
`Clusters.pw.tsx_mobile_Clusters-Directory-Page-loading-state-mobile-1.png`


```suggestion
    test('clusters directory with data +@mobile', async({ render, mockApiResponse, mockAssetResponse }) => {
```

ui/shared/entities/clusters/ClustersEntity.tsx
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  const fallbackElement = (
+    <Box
+      display="inline-flex"
+      alignItems="center"
+      justifyContent="center"
+      width={ styles.boxSize }
+      height={ styles.boxSize }
+      backgroundColor={ colors.clusters.value }
```
tom2drum: Sorry if I didn't point it out earlier, but the color here should also be set through the color token name, as it is done below.

ui/shared/entities/clusters/ClustersEntity.pw.tsx
```diff
@@ -0,0 +1,125 @@
+import React from 'react';
```
tom2drum, line 1: Same comment about removing the unused screenshots that were generated earlier.

ui/pages/Cluster.pw.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { test, expect } from 'playwright/lib';
+
+import Cluster from './Cluster';
+
+test.beforeEach(async({ mockEnvs, mockTextAd }) => {
+  await mockEnvs(ENVS_MAP.clusters);
+  await mockTextAd();
+});
+
+test.describe('Cluster Details Page', () => {
```
tom2drum, line 15: There are some previously generated screenshots for this test suite as well.

ui/snippets/searchBar/useSearchWithClusters.tsx
tom2drum: Thanks for the explanation. Now I understand your objectives. 

I believe we can achieve the same result by replacing `useApiQuery` with a combination of `useQuery` and `useApiFetch` to create a custom `queryFn`. We will need to repeat some of the logic from `useApiQuery`, but I assume that won't be a significant issue. Or perhaps we could extend the functionality of `useApiQuery` to allow it to return `null` instead of an error. Since we have not encountered this problem so far, the ad-hoc solution seems preferable for now.

Please review the code snippet below and let me know if it works for you:
```typescript
const RESOURCE_NAME = 'clusters:get_cluster_by_name';
type ClusterQueryResult = ResourcePayload<typeof RESOURCE_NAME>;

const clusterQuery = useQuery<ClusterQueryResult | null, ResourceError<unknown>, Array<SearchResultCluster>>({
  queryKey: getResourceKey(RESOURCE_NAME, { queryParams: { input: clusterName } }),
  queryFn: async({ signal }) => {
    try {
      const result = await apiFetch(RESOURCE_NAME, {
        queryParams: { input: JSON.stringify({ name: clusterName }) },
        fetchParams: { signal },
      }) as ClusterQueryResult;
      return result;
    } catch (error) {
      return null;
    }
  },
  enabled: config.features.clusters.isEnabled && isClusterQuery && clusterName.length > 0,
  select: (data) => {
    if (!data?.result?.data) return [];
    return [ transformClusterToSearchResult(data.result.data, data.result.data.owner) ];
  },
});
```

ayv8er: Yes, this works with me!

tom2drum (commented):
Looks good to me now.
Awesome job 🙌

tom2drum (approved):

ui/snippets/searchBar/useSearchWithClusters.test.tsx
```diff
@@ -304,11 +333,30 @@ describe('useSearchWithClusters', () => {
...
+        failureReason: null,
+        errorUpdateCount: 0,
+        isFetched: true,
+        isFetchedAfterMount: true,
+        isPlaceholderData: false,
+        isRefetching: false,
+        isInitialLoading: false,
+        isFetching: false,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } as any);
```
tom2drum: Can we store all default fields in a global object and then clone it in each test case using the spread operator? We can redefine only the properties necessary for a specific test scenario. This way we avoid code duplication.

Additionally, please avoid using `any`. In this case, I assume you can use coercion to `UseQueryResult` instead.

tom2drum:
I am closing this Pull Request in favor of the new one, #2994. The main change I made is moving the Cluster page inside the tab on the Name Service page.

@ayv8er, thank you for your contribution! 🙌

Issue #2837, opened 2025-07-07
Mobile UI: search (drawer)


tgladilina:
- Chakra component - [drawer](https://chakra-ui.com/docs/components/drawer)
- Min mobile width - 360px

- [ ]  Move search field to magnifying glass icon (on all pages except the main one).  [FIGMA](https://www.figma.com/design/OjqHcLt95E6kq5ElO79EKT/%F0%9F%93%B1-Blockscout-mobile-version?node-id=1-37475&t=hrZLLs5YaquSyZYo-4)

![Image](https://github.com/user-attachments/assets/aa886ee1-46d2-4543-93a7-fffe215279a5)
![Image](https://github.com/user-attachments/assets/1111bc7d-3ef7-44ff-9f34-24cbc9c214fe)

- [ ]  Clicking on the search icon (header) or search input (on the main page) opens a drawer (fix hight) = 75% of the screen. [FIGMA ](https://www.figma.com/design/ErLH8qySSkpprB87H0onWw/Blockscout_Pages?node-id=27521-157710&t=KnkiutNt3vnhWsQu-4)

- [ ]  change tabs text (ONLY MOBILE version): 
Tokens (ERC-20) > Tokens
NFTs (ERC-721 & 1155) > NFTs 

![Image](https://github.com/user-attachments/assets/dcac0fd0-b639-4517-bcda-a21cf6ae249b)

isstuev:
also I'm not changing the design of the search items here. Only the results list's layout. If needed, we can change items later but I will need mockups for all categories and different cases.

PR #2855, opened 2025-07-14
Mobile UI: search (drawer)

isstuev:

resolves https://github.com/blockscout/frontend/issues/2837

## Checklist for PR author


toolkit/theme/foundations/semanticTokens.ts
```diff
@@ -256,7 +256,7 @@ const semanticTokens: ThemingConfig['semanticTokens'] = {
         error: { value: '{colors.text.error}' },
       },
       bg: {
-        DEFAULT: { value: { _light: '{colors.white}', _dark: '{colors.black}' } },
+        DEFAULT: { value: 'transparent' },
```
tom2drum: It should not be transparent; otherwise, the input background will be incorrect if there are underlying elements with a custom background. If the input is rendered in a dialog, it is better to pass the background color as a prop to it (see the example [here](https://github.com/blockscout/frontend/blob/bb50fc6daebbd43484bcf28018415f8023faec49/ui/apiKey/ApiKeyModal/ApiKeyForm.tsx#L111)).

https://www.dropbox.com/scl/fi/rty05j7ii46i2lpjhhyn2/loating_label.png?rlkey=namvqrzif4or8rqb2mbx3szpd&dl=0

toolkit/chakra/input-group.tsx
```diff
@@ -31,33 +31,68 @@ export const InputGroup = React.forwardRef<HTMLDivElement, InputGroupProps>(
...
 
+    React.useEffect(() => {
+      if (!groupRef.current) return;
+
+      const intersectionObserver = new IntersectionObserver(
+        (entries) => {
+          const entry = entries[0];
+          if (entry && entry.isIntersecting) {
+            // Small delay to ensure rendering is complete
+            setTimeout(calculateInlinePaddings, 50);
```
tom2drum: I think this timeout should be cleared when component unmounts from the DOM.

toolkit/theme/recipes/drawer.recipe.ts
```diff
@@ -73,8 +73,9 @@ export const recipe = defineSlotRecipe({
       pb: '4',
     },
     title: {
-      textStyle: 'lg',
-      fontWeight: 'semibold',
+      textStyle: 'heading.md',
+      lineHeight: '32px',
```
tom2drum: I believe the line height should remain at the default value, but the minimum height for the header should be set. Additionally, there is no need to define the font weight here, as it is already specified in the typography settings.

ui/shared/layout/LayoutApp.tsx
```diff
@@ -20,7 +20,7 @@ const LayoutDefault = ({ children }: Props) => {
         flexDirection="column"
       >
         <Layout.TopRow/>
-        <HeaderMobile hideSearchBar/>
+        <HeaderMobile/>
```
tom2drum, line 23: The position of action bar on app page is not correct.

https://www.dropbox.com/scl/fi/dtinusjhndzevldh7n8g3/Screenshot-2025-08-01-at-1.50.10-PM.png?rlkey=rk1vyathjgf7b8p77o6i3pf7u&dl=0

tom2drum: It appears that the issue affects all pages.

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggest.tsx
```diff
@@ -151,71 +167,71 @@ const SearchBarSuggest = ({ query, searchTerm, onItemClick, containerId }: Props
...
+            size="sm"
+            pb={ 5 }
+            overflowX="hidden"
+            minH="52px"
+            h="52px"
+            listProps={{
+              overflowX: 'auto',
+              mb: 0,
+              mt: 0,
+              background: { _light: 'white', _dark: 'gray.900' },
```
tom2drum:
```suggestion
              background: 'dialog.bg',
```

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggest.tsx
```diff
@@ -151,71 +167,71 @@ const SearchBarSuggest = ({ query, searchTerm, onItemClick, containerId }: Props
...
-                    key={ cat.id }
-                    value={ cat.id }
-                  >
-                    { cat.title }
-                  </TabsTrigger>
-                )) }
-              </TabsList>
-            </TabsRoot>
-          </Box>
+          <AdaptiveTabs
```
tom2drum, line 181: I am unsure whether implementing this functionality with the Tabs component is a good idea. Could it simply be `AdaptiveTabsList` if there is no TabContent? Should it adapt at all? According to the mock-ups, the list should be scrollable on mobile, and on desktop, there aren't enough search categories to justify hiding anything under the cut.

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggest.tsx
```diff
@@ -121,13 +121,29 @@ const SearchBarSuggest = ({ query, searchTerm, onItemClick, containerId }: Props
...
+    return searchCategories.filter(cat => itemsGroups[cat.id]).map(cat => ({
+      id: cat.id,
+      value: cat.id,
+      title: isMobile ? cat.tabTitle : cat.title,
+      component: null,
+    }));
+  }, [ itemsGroups, isMobile ]);
 
   const content = (() => {
     if (query.isPending || marketplaceApps.isPlaceholderData) {
```
tom2drum: Due to changes in the layout, loader now have the full width. But I guess it should be restricted to the maximum of 250px or so.

ui/snippets/searchBar/SearchBarMobile.pw.tsx
```diff
@@ -0,0 +1,313 @@
...
+  await mockTextAd();
+  await mockAssetResponse(searchMock.token1.icon_url as string, './playwright/mocks/image_s.jpg');
+  await mockEnvs([
+    [ 'NEXT_PUBLIC_MARKETPLACE_ENABLED', 'false' ],
+  ]);
+});
+
+test.use({ viewport: devices['iPhone 13 Pro'].viewport });
+
+test('search by token name +@dark-mode', async({ render, page, mockApiResponse }) => {
```
tom2drum, line 43: I believe we don't need to test for dark mode in all mobile component test suites, as the colors are essentially the same as those for the desktop component.

ui/snippets/searchBar/SearchBarInput.tsx
```diff
@@ -17,52 +15,34 @@ interface Props extends Omit<HTMLChakraProps<'form'>, 'onChange'> {
...
+    if (!readOnly) {
+      onChange(event.target.value);
+    }
+  }, [ onChange, readOnly ]);
 
-  React.useEffect(() => {
-    if (!isMobile) {
+  const handleFocus = React.useCallback((event: FocusEvent<HTMLInputElement>) => {
+    if (readOnly) {
+      event.target.blur();
```
tom2drum: It's a bit confusing to read. Why should the focus event handler trigger the blur event on the target?

ui/snippets/searchBar/SearchBarInput.tsx
```diff
@@ -133,36 +111,33 @@ const SearchBarInput = (
...
+          onFocus={ handleFocus }
+          border={ isHeroBanner ? 'none' : '2px solid' }
+          borderColor="input.border"
           color={{ _light: 'black', _dark: 'white' }}
+          backgroundColor="input.bg"
+          _placeholder={{
+            textOverflow: 'ellipsis',
+            overflow: 'hidden',
+            whiteSpace: 'nowrap',
+          }}
```
tom2drum: Is the truncation of the placeholder already implemented in `input.recipe.ts`?

ui/snippets/searchBar/SearchBarInput.tsx
```diff
@@ -133,36 +111,33 @@ const SearchBarInput = (
...
+          size={{ base: isHeroBanner ? 'md' : 'sm', lg: 'md' }}
+          placeholder="Search by address / txn hash / block / token..."
           value={ value }
           onChange={ handleChange }
-          onFocus={ onFocus }
-          border={ isHomepage ? 'none' : '2px solid' }
-          borderColor={{ _light: 'blackAlpha.100', _dark: 'whiteAlpha.200' }}
+          onFocus={ handleFocus }
+          border={ isHeroBanner ? 'none' : '2px solid' }
+          borderColor="input.border"
```
tom2drum: I think this property is not necessary.

ui/snippets/searchBar/SearchBarInput.tsx
```diff
@@ -133,36 +111,33 @@ const SearchBarInput = (
...
           color={{ _light: 'black', _dark: 'white' }}
+          backgroundColor="input.bg"
+          _placeholder={{
+            textOverflow: 'ellipsis',
+            overflow: 'hidden',
+            whiteSpace: 'nowrap',
+          }}
           _hover={{ borderColor: 'input.border.hover' }}
           _focusWithin={{ _placeholder: { color: 'gray.300' }, borderColor: 'input.border.focus', _hover: { borderColor: 'input.border.focus' } }}
```
tom2drum, line 141: This can be safely deleted now, since in the default state input has its default border color.

ui/snippets/searchBar/SearchBarRecentKeywords.tsx
```diff
@@ -42,53 +43,55 @@ const SearchBarSuggest = ({ onClick, onClear }: Props) => {
...
+            flexGrow={ 1 }
+          >
+            { kw.startsWith('0x') ? (
+              <Box overflow="hidden" whiteSpace="nowrap">
+                <HashStringShortenDynamic hash={ kw } noTooltip/>
+              </Box>
+            ) :
+              <Text overflow="hidden" whiteSpace="nowrap" textOverflow="ellipsis">{ kw }</Text>
+            }
+            <ClearButton onClick={ removeKeyword(kw) } color={{ _light: 'gray.300', _dark: 'gray.600' }}/>
```
tom2drum, line 90: As discussed with designer, the color should be updated for all ClearButtons.

ui/snippets/searchBar/SearchBarInput.tsx
```diff
@@ -111,7 +89,7 @@ const SearchBarInput = (
 
   const endElement = (
     <>
-      <ClearButton onClick={ onClear } visible={ value.length > 0 } mx={ 2 }/>
+      <ClearButton onClick={ onClear } visible={ Boolean(value?.length) } mx={ 2 } color="input.placeholder"/>
```
tom2drum, line 94: As discussed with designer, the color should be updated for all ClearButtons.

ui/snippets/searchBar/SearchBarMobile.pw.tsx
```diff
@@ -0,0 +1,313 @@
...
+test('search by contract name +@dark-mode', async({ render, page, mockApiResponse }) => {
+  const apiUrl = await mockApiResponse('general:quick_search', [
+    searchMock.contract1,
+    searchMock.contract2,
+    searchMock.address2,
+  ], { queryParams: { q: 'o' } });
+
+  await render(<SearchBarMobile/>);
+  await openSearchDrawer(page);
+  await page.getByPlaceholder(/search/i).fill('o');
```
tom2drum: Should other occurrences of `page.getByPlaceholder(/search/i)` be replaced with the `getSearchInput()` function, similar to how it is done for the homepage test?

```suggestion
  await getSearchInput(page).fill('o');
```

ui/snippets/searchBar/SearchBarMobile.tsx
```diff
@@ -0,0 +1,151 @@
...
+import SearchBarSuggest from './SearchBarSuggest/SearchBarSuggest';
+import useQuickSearchQuery from './useQuickSearchQuery';
+
+type Props = {
+  isHeroBanner?: boolean;
+};
+
+const SearchBarMobile = ({ isHeroBanner }: Props) => {
+  const inputRef = React.useRef<HTMLFormElement>(null);
+  const voidFn = React.useCallback(() => {}, []);
```
tom2drum: It can be replaced with the `noop` function from `es-toolkit`.

ui/snippets/searchBar/SearchBarMobile.tsx
```diff
@@ -0,0 +1,151 @@
...
+  let trigger: React.ReactNode | null = null;
+  if (isHeroBanner) {
+    trigger = (
+      <SearchBarInput
+        onChange={ voidFn }
+        onSubmit={ voidFn }
+        onFocus={ voidFn }
+        onHide={ voidFn }
+        onBlur={ voidFn }
+        onClear={ voidFn }
```
tom2drum: As for me, it will be simpler to have a transparent overlay over the input, so clicking on it will trigger the opening of the drawer. So none of the adjustments to the SearchBarInput component will not be necessary.

ui/snippets/searchBar/SearchBarMobile.tsx
```diff
@@ -0,0 +1,151 @@
...
+          <DrawerCloseTrigger/>
+        </DrawerHeader>
+        <DrawerBody overflow="hidden" display="flex" flexDirection="column">
+          <SearchBarInput
+            ref={ inputRef }
+            onChange={ handleSearchTermChange }
+            onSubmit={ voidFn }
+            onFocus={ voidFn }
+            onHide={ voidFn }
+            onBlur={ voidFn }
```
tom2drum: If we don't need these handlers in some scenarios, it might be better to make those props optional.

ui/snippets/searchBar/SearchBarMobile.tsx
```diff
@@ -0,0 +1,151 @@
...
+            onClear={ handleClear }
+            onFormClick={ onTriggerClick }
+            value={ searchTerm }
+            mt={ -5 }
```
tom2drum: Can we redefine the top padding of the drawer body instead of applying a negative margin here?

ui/snippets/searchBar/SearchBarMobile.tsx
```diff
@@ -0,0 +1,151 @@
...
+              searchTerm={ debouncedSearchTerm }
+              onItemClick={ handleItemClick }
+            />
+          ) }
+        </DrawerBody>
+        { (query.data && query.data?.length > 0) && (
+          <DrawerFooter
+            borderTop="1px solid"
+            borderColor="border.divider"
+            bg="background.primary"
```
tom2drum: What is the reason for changing the background here? It seems we don't have that token at the moment.

ui/pages/SearchResults.pw.tsx
```diff
@@ -9,8 +9,42 @@ import { test, expect } from 'playwright/lib';
...
-  await page.evaluate(() => window.scrollTo(0, document.body.scrollHeight));
+  // Scroll to top and ensure we stay there
   await page.evaluate(() => window.scrollTo(0, 0));
+
+  // Wait for sticky headers to be in their default (hidden) state
+  // Check for ActionBar - it should not be visible when at top
+  await page.waitForFunction(() => {
+    const actionBar = document.querySelector('[data-testid="action-bar"]') ||
+                     document.querySelector('[class*="ActionBar"]') ||
+                     document.querySelector('[class*="action-bar"]');
```
tom2drum: Is this code correct? I cannot find any occurrences of the ActionBar class name or the action-bar test ID.

ui/pages/SearchResults.pw.tsx
```diff
@@ -9,8 +9,42 @@ import { test, expect } from 'playwright/lib';
...
+      const rect = actionBar.getBoundingClientRect();
+      // ActionBar should be in its default position (not sticky) when at top
+      return rect.top > 0;
+    }
+    return true; // If no ActionBar found, consider it "hidden"
+  }, { timeout: 500 });
+
+  // Check for TableHeaderSticky - it should not be stuck when at top
+  await page.waitForFunction(() => {
+    const stickyHeaders = document.querySelectorAll('[class*="TableHeaderSticky"]');
```
tom2drum: Same comment here.
And not all table headers are stuck at the top of the page.

ui/snippets/searchBar/SearchBarMobile.pw.tsx
isstuev: not 100% same. input bg for example :)

toolkit/theme/foundations/semanticTokens.ts
isstuev: I think in general input should have the same background as the underlying element, and if not, it can be changed by a prop. because now we have (afaik) only one case when we need background to not be transparent and it's hero banner. So in general I want to be sure that input bg is same as underlying bg without any additional styling.

agree that field label styles should be changed accordingly.

tom2drum: > it's hero banner

It is not true.
The designer thinks differently. In main scenarios it has the color of the page body. The only exceptions are dialogs.

image
image

ui/snippets/searchBar/SearchBarMobile.pw.tsx
tom2drum: Okay, but I think we only need to test it in one test suite. That should be enough.

toolkit/theme/foundations/semanticTokens.ts
isstuev: The designer told me the opposite :))

isstuev: will be resolved in a separate pr

toolkit/theme/recipes/drawer.recipe.ts
isstuev: if I remove font weight property, the title gets 600 (semibold), and I can't find the cause of that problem

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggest.tsx
isstuev: list of categories may change and width of the suggest is not fixed, so theoretically cut is possible here.

ui/snippets/searchBar/SearchBarInput.tsx
isstuev: I see that it works incorrect without this style

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggest.tsx
isstuev: AdaptiveTabs has cut logics I want to keep, so I think it's ok to use AdaptiveTabs

ui/snippets/searchBar/SearchBarInput.tsx
isstuev: no it shouldn't 
and if somehow it should, we will implement it in a separate task 
image

ui/snippets/searchBar/SearchBarSuggest/SearchBarSuggest.tsx
isstuev: doesn't work this way

tom2drum: yep, it sould be `bgColor` prop

`bgColor: 'dialog.bg'`

ui/snippets/searchBar/SearchBarInput.tsx
tom2drum: This button primarily appears in input fields and should, by default, match the color of the input placeholder. Nevertheless, this property is unnecessary since all starting and ending elements of the input have the default color of the placeholder. For the ClearButton related to the recent keyword item, we can use the same color token instead of defining a custom color.

ui/snippets/searchBar/SearchBarMobile.pw.tsx
```diff
@@ -0,0 +1,313 @@
...
+  await expect(page).toHaveScreenshot({ clip: { x: 0, y: 0, width: 1200, height: 700 } });
+});
+
+test('search by tag', async({ render, page, mockApiResponse }) => {
+  const apiUrl = await mockApiResponse('general:quick_search', [
+    searchMock.label1,
+  ], { queryParams: { q: 'o' } });
+  await render(<SearchBarMobile/>);
+  await openSearchDrawer(page);
+  await page.getByPlaceholder(/search/i).fill('o');
```
tom2drum:
```suggestion
  await getSearchInput(page).fill('o');
```

ui/snippets/searchBar/SearchBarMobile.tsx
```diff
@@ -0,0 +1,162 @@
...
+          style={{
+            position: 'absolute',
+            top: 0,
+            left: 0,
+            right: 0,
+            bottom: 0,
+            zIndex: 1,
+            backgroundColor: 'transparent',
+            cursor: 'pointer',
+          }}
```
tom2drum: Is there a reason for using the inline styles here and above?

ui/snippets/searchBar/SearchBarMobile.tsx
```diff
@@ -0,0 +1,162 @@
...
+    event.stopPropagation();
+    onTriggerClick(event);
+  }, [ onTriggerClick ]);
+
+  let trigger: React.ReactNode | null = null;
+  if (isHeroBanner) {
+    trigger = (
+      <div style={{ position: 'relative', width: '100%' }}>
+        <SearchBarInput
+          onFormClick={ onTriggerClick }
```
tom2drum: Is this prop still necessary if the drawer opens by clicking on the overlay?

ui/snippets/searchBar/SearchBarMobile.tsx
```diff
@@ -0,0 +1,162 @@
...
+        <DrawerHeader>
+          <DrawerTitle>Search</DrawerTitle>
+          <DrawerCloseTrigger/>
+        </DrawerHeader>
+        <DrawerBody overflow="hidden" display="flex" flexDirection="column">
+          <SearchBarInput
+            ref={ inputRef }
+            onChange={ handleSearchTermChange }
+            onClear={ handleClear }
+            onFormClick={ onTriggerClick }
```
tom2drum: Is this prop still necessary if the drawer opens by clicking on the overlay?

ui/snippets/searchBar/SearchBarInput.tsx
tom2drum: I've created a separate task - https://github.com/blockscout/frontend/issues/3012 - since the color of clear button and the end input element now should be different.

PR #2916, opened 2025-07-30
Support new API for dapp ratings

maxaleks:

Resolves https://github.com/blockscout/frontend/issues/2619

Removed envs:
```
NEXT_PUBLIC_MARKETPLACE_RATING_AIRTABLE_API_KEY
NEXT_PUBLIC_MARKETPLACE_RATING_AIRTABLE_BASE_ID
```

## Checklist for PR author

### DIFF
--- a/deploy/tools/envs-validator/schema.ts
+++ b/deploy/tools/envs-validator/schema.ts
@@ -18,7 +18,7 @@ import type { DeFiDropdownItem } from '../../../types/client/deFiDropdown';
 import type { GasRefuelProviderConfig } from '../../../types/client/gasRefuelProviderConfig';
 import { GAS_UNITS } from '../../../types/client/gasTracker';
 import type { GasUnit } from '../../../types/client/gasTracker';
-import type { MarketplaceApp } from '../../../types/client/marketplace';
+import type { MarketplaceAppBase, MarketplaceAppSocialInfo } from '../../../types/client/marketplace';
 import type { MultichainProviderConfig } from '../../../types/client/multichainProviderConfig';
 import type { ApiDocsTabId } from '../../../types/views/apiDocs';
 import { API_DOCS_TABS } from '../../../types/views/apiDocs';
@@ -80,7 +80,7 @@ const getYupValidationErrorMessage = (error: unknown) =>
     error.errors.join(', ') :
     '';
 
-const marketplaceAppSchema: yup.ObjectSchema<MarketplaceApp> = yup
+const marketplaceAppSchema: yup.ObjectSchema<MarketplaceAppBase & MarketplaceAppSocialInfo> = yup
   .object({
     id: yup.string().required(),
     external: yup.boolean(),
@@ -103,9 +103,6 @@ const marketplaceAppSchema: yup.ObjectSchema<MarketplaceApp> = yup
     discord: yup.string().test(urlTest),
     internalWallet: yup.boolean(),
     priority: yup.number(),
-    rating: yup.number(),
-    ratingsTotalCount: yup.number(),
-    userRating: yup.number(),
   });
 
 const marketplaceSchema = yup
--- a/types/client/marketplace.ts
+++ b/types/client/marketplace.ts
@@ -1,4 +1,4 @@
-type MarketplaceAppBase = {
+export type MarketplaceAppBase = {
   id: string;
   author: string;
   site?: string;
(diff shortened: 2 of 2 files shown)


deploy/tools/envs-validator/schema.ts
```diff
@@ -103,6 +103,9 @@ const marketplaceAppSchema: yup.ObjectSchema<MarketplaceAppOverview> = yup
     discord: yup.string().test(urlTest),
     internalWallet: yup.boolean(),
     priority: yup.number(),
+    rating: yup.number(),
```
tom2drum: If the schema changes, should we also add these properties to the documentation?

tom2drum: Not sure if `userRating` make sense though.

ui/showcases/Rating.tsx
```diff
@@ -11,15 +11,15 @@ const RatingShowcase = () => {
         <SectionHeader>Size</SectionHeader>
         <SamplesStack>
           <Sample label="size: md">
-            <Rating defaultValue={ 3 } label={ [ 'Very bad', 'Bad', 'Average', 'Good', 'Excellent' ] }/>
+            <Rating value={ 3 } label={ [ 'Very bad', 'Bad', 'Average', 'Good', 'Excellent' ] }/>
```
tom2drum: What was the reason of these changes?

toolkit/chakra/rating.tsx
```diff
@@ -7,15 +7,15 @@ import StarOutlineIcon from 'icons/star_outline.svg';
...
   onValueChange?: ({ value }: { value: number }) => void;
   readOnly?: boolean;
 }
 
 export const Rating = React.forwardRef<HTMLDivElement, RatingProps>(
   function Rating(props, ref) {
-    const { count = 5, label: labelProp, defaultValue, onValueChange, readOnly, ...rest } = props;
-    const store = useRatingGroup({ count, defaultValue, onValueChange, readOnly });
+    const { count = 5, label: labelProp, value, onValueChange, readOnly, ...rest } = props;
+    const store = useRatingGroup({ count, value, onValueChange, readOnly });
```
tom2drum:
```suggestion
    const { count = 5, label: labelProp, defaultValue, value, onValueChange, readOnly, ...rest } = props;
    const store = useRatingGroup({ count, defaultValue, value, onValueChange, readOnly });
```

ui/marketplace/Rating/Rating.tsx
```diff
@@ -49,28 +45,25 @@ const MarketplaceRating = ({
     >
       { fullView && (
         <>
-          <Rating defaultValue={ Math.floor(rating?.value || 0) } readOnly key={ rating?.value }/>
-          <Text fontSize="md" ml={ 2 }>{ rating?.value }</Text>
-          { rating?.count && <Text color="text.secondary" textStyle="md" ml={ 1 }>({ rating?.count })</Text> }
+          <Rating value={ Math.floor(rating || 0) } readOnly key={ rating }/>
```
tom2drum: If now the value is passed (not default value), I guess there is no need for key here.

ui/pages/MarketplaceApp.tsx
```diff
@@ -131,7 +133,12 @@ const MarketplaceApp = () => {
     },
     enabled: feature.isEnabled,
   });
-  const { data, isPending } = query;
+  const { data, isPending, refetch } = query;
+
+  React.useEffect(() => {
+    refetch();
+  }, [ profileQuery.data, refetch ]);
```
tom2drum: It may not be a good idea to refetch every time the profile data changes. If we want to trigger a refetch only when the user logs in, it would be better to use a boolean flag as a dependency for the hook.

ui/marketplace/useMarketplaceApps.tsx
```diff
@@ -93,15 +94,18 @@ export default function useMarketplaceApps(
     enabled: feature.isEnabled && Boolean(snapshotFavoriteApps),
   });
 
+  React.useEffect(() => {
```
tom2drum, line 94: I notice in the developer console that the request is sent twice when I open the page. Probably, the usage of the `useUpdateEffect` hook here can fix this behavior.
The same comment about deps here.

tom2drum (commented):
Looks good, only minor comments about code. 👍

deploy/tools/envs-validator/schema.ts
maxaleks: I don't think so, because the ratings only work with API

tom2drum (approved):

PR #3025, opened 2025-09-24
display 'copy' button by default for tx hash in the lists 

### LINKED ISSUE #2592: Add 'copy' button for tx hash in the lists

iamenochlee:

This pull request enables the copy button for transaction hashes in transaction lists by default, improving user experience consistency with address entities. Previously, users could not easily copy transaction hashes from list views, requiring them to navigate to individual transaction pages.

*Resolves #2592 *

### Proposed Changes

- Modified `TxEntity.tsx` component to show copy button by default for transaction hashes
- Updated the `Copy` component to set `noCopy={ props.noCopy ?? false }` instead of hiding the copy functionality
- This change aligns transaction hash copy behavior with address entity copy behavior across the application

No changes were made to ENV variables.

No breaking changes. This is a purely additive enhancement that improves UX. The copy button can still be hidden by explicitly setting `noCopy={true}` if needed in specific implementations.

### Additional Information

This enhancement improves user workflow by allowing direct copying of transaction hashes from list views without requiring navigation to individual transaction pages. The change maintains backward compatibility while providing better default behavior.

## Checklist for PR author


iamenochlee:
@tom2drum requesting review

ui/shared/entities/tx/TxEntity.tsx
```diff
@@ -84,8 +84,7 @@ const Copy = (props: CopyProps) => {
     <EntityBase.Copy
       { ...props }
       text={ props.hash }
-      // by default we don't show copy icon, maybe this should be revised
-      noCopy={ props.noCopy ?? true }
+      noCopy={ props.noCopy ?? false }
```
isstuev, line 87: default is not needed here.

isstuev (changes requested):
All the places where TxEntity is used should be reviewed.
Since task is about list views (and tables as well, I suppose), there still can be places where copy icon should not appear.

Also I'd recommend to remove noCopy={ false } prop from TxEntity component across the codebase since it's not needed anymore.

ui/shared/entities/tx/TxEntity.tsx
enochakinbode: you are right about this

PR #3030, opened 2025-09-26
add retries for assets downloading, fail build

isstuev:

resolves https://github.com/blockscout/frontend/issues/3029

### DIFF
--- a/deploy/scripts/download_assets.sh
+++ b/deploy/scripts/download_assets.sh
@@ -131,7 +131,7 @@ done
 
 # Check if any downloads failed and exit with error code
 if [ $FAILED_DOWNLOADS -gt 0 ]; then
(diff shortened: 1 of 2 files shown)


deploy/scripts/download_assets.sh
```diff
@@ -125,5 +129,11 @@ for env_var in "${ASSETS_ENVS[@]}"; do
     download_and_save_asset "$env_var" "$url" "$filename"
 done
 
+# Check if any downloads failed and exit with error code
+if [ $FAILED_DOWNLOADS -gt 0 ]; then
+    echo "🛑 Error: $FAILED_DOWNLOADS asset download(s) failed. Build cannot continue."
```
tom2drum: "Application start cannot continue." would be more correct message.

PR #3031, opened 2025-09-29
display copy button for tx hashes in relevant components

### LINKED ISSUE #2592: Add 'copy' button for tx hash in the lists

iamenochlee:

This pull request enables the copy button for transaction hash in transaction lists by default while ensuring it is only displayed in relevant pages/tabs mostly user specific ones such as:
[https://eth.blockscout.com/address/0x5AbFEc25f74Cd88437631a7731906932776356f9?tab=txs](https://eth.blockscout.com/address/0x5AbFEc25f74Cd88437631a7731906932776356f9?tab=txs)
[https://eth.blockscout.com/address/0x5AbFEc25f74Cd88437631a7731906932776356f9?tab=internal_txns](https://eth.blockscout.com/address/0x5AbFEc25f74Cd88437631a7731906932776356f9?tab=internal_txns)

Previously, users could not easily copy transaction hashes from list views, requiring them to navigate to individual transaction pages.

*Resolves #2592 *

### Proposed Changes

- **Enhanced transaction hash copy functionality**: Modified `TxEntity.tsx` and other similar `TxEntityXXX.tsx`component to display copy buttons by default for transaction hashes, improving user experience when interacting with transaction data
- **Updated default behavior**: Changed the `Copy` component logic to set `noCopy={ props.noCopy}`, making copy buttons for tx hash visible unless explicitly disabled
- **User-centric vs system-wide distinction**: 
  - **Enabled copy buttons** on user-specific pages where transaction hashes are most relevant:
    - Address transaction lists (`ui/address/[0x]/txs`)
    - Address internal transactions (`ui/address/[0x]/internalTxs`) 
    - Address token transfers (`ui/address/[0x]/tokenTransfers`)
    - Other address-specific transaction views
  - **Explicitly disabled copy buttons** (`noCopy` prop) on system-wide pages where copy functionality may be less relevant or could clutter the interface
- **Comprehensive component updates**: Modified multiple transaction entity components to adopt the new default behavior while maintaining appropriate UX for different contexts
- **Cleaned up redundant props**: Removed unnecessary `noCopy={false}` props from user-focused components where copy buttons should now display by default
- **Updated Playwright snapshots**: Refreshed visual regression test snapshots to reflect the new copy button visibility across affected components
- **Maintained UX consistency**: Ensured transaction hash copy behavior is contextually appropriate - visible where users need it most (their own transaction history) and hidden where it might create interface clutter (system-wide transaction lists)

This enhancement provides a more intuitive and user-friendly approach to transaction hash copying, making the functionality available where users expect it while keeping system-wide interfaces clean and focused.

No changes were made to ENV variables.

No breaking changes. This is a purely additive enhancement that improves UX. The copy button can still be hidden by explicitly setting `noCopy={true}` if needed in specific implementations.

### Additional Information

This enhancement improves user workflow by allowing direct copying of transaction hashes from list views without requiring navigation to individual transaction pages. The change maintains backward compatibility while providing better default behavior.

## Checklist for PR author


ui/block/BlockInternalTxs.tsx
```diff
@@ -17,10 +17,10 @@ const BlockInternalTxs = ({ query, top }: Props) => {
   const content = data?.items ? (
     <>
       <Box hideFrom="lg">
-        <InternalTxsList data={ data.items } isLoading={ isPlaceholderData } showBlockInfo={ false }/>
+        <InternalTxsList data={ data.items } isLoading={ isPlaceholderData } showBlockInfo={ false } noCopy/>
```
isstuev: I assume 'noCopy' isn't the correct prop name for table/list component. If we really need such prop, better name would be 'noTxHashCopy' or something like that.
But I don't see why we would hide copy icon for internal txs list on the block page

ui/deposits/beaconChain/BeaconChainDepositsListItem.tsx
```diff
@@ -32,7 +32,7 @@ const BeaconChainDepositsListItem = ({ item, isLoading, view }: Props) => {
 
       <ListItemMobileGrid.Label isLoading={ isLoading }>Transaction hash</ListItemMobileGrid.Label>
       <ListItemMobileGrid.Value>
-        <TxEntity hash={ item.transaction_hash } isLoading={ isLoading } truncation="constant_long"/>
+        <TxEntity hash={ item.transaction_hash } isLoading={ isLoading } truncation="constant_long" noCopy/>
```
isstuev: I think we don't need 'noCopy' here

isstuev: actually, here and further where we have tables or lists (and components like -TableItem, -ListItem) we want to add copy icons unless there's a goof reason not to do so (can't think of any examples though)

enochakinbode: alright

enochakinbode: @isstuev 

> actually, here and further where we have tables or lists (and components like -TableItem, -ListItem)

there are ui pages like the /txs , /internal-txs using this components, do we want to show. copy button for tx hashes there?

ui/block/BlockInternalTxs.tsx
enochakinbode: the prop `noTxHashCopy` is needed here as this internalTxsList is also rendered in the page /internalTxs, where i have hidden the tx copy button as i argue there is no need for it there

ui/deposits/beaconChain/BeaconChainDepositsListItem.tsx
isstuev: I think we want, /txs, /internal-txs are even specified in the task description, and the task says nothing about exceptions, so I read it as all tables/lists.

enochakinbode: thanks for clarifying

ui/advancedFilter/ItemByColumn.tsx
```diff
@@ -25,7 +25,7 @@ type Props = {
 const ItemByColumn = ({ item, column, isLoading }: Props) => {
   switch (column) {
     case 'tx_hash':
-      return <TxEntity truncation="constant_long" hash={ item.hash } isLoading={ isLoading } noIcon fontWeight={ 700 }/>;
```
tom2drum, line 28: Please keep `truncation` property here

ui/home/latestZetaChainCCTX/LatestZetaChainCCTXItem.tsx
```diff
@@ -29,9 +29,10 @@ const LatestZetaChainCCTXItem = ({ tx, isLoading, animation }: Props) => {
       p={ 4 }
       fontSize="sm"
       animation={ animation }
+
     >
       <ZetaChainCCTXReducedStatus status={ tx.status_reduced } isLoading={ isLoading }/>
-      <TxEntityZetaChainCC hash={ tx.index } isLoading={ isLoading } truncation="constant" fontWeight={ 600 }/>
```
tom2drum, line 34: Please maintain the same value for the `truncation` prop.

ui/advancedFilter/ItemByColumn.tsx
enochakinbode: Screen Shot 2025-10-17 at 6 43 28 PM

@tom2drum the hash doesn't fit which is why i set the truncation to dynamic by removing it
The full hash `0x5cb8769f7d055f4d12bdf7807811ae5af5903a8223b544ce4e36218b2d3f04fc`

tom2drum: We try to avoid dynamic truncation in the list because it can impact page performance. If the hash no longer fits in the column, we must either widen the column or use `truncation="constant"` to display fewer symbols of the hash.

enochakinbode: good to know, i have defaulted several TxEntity especially the ones showing an icon to dynamic, i will go through them.

ui/txs/TxsTableItem.tsx
```diff
@@ -53,7 +53,7 @@ const TxsTableItem = ({ tx, showBlockInfo, currentAddress, enableTimeIncrement,
             fontWeight="bold"
             noIcon
             maxW="100%"
-            truncation="constant_long"
```
tom2drum, line 56: What is the reason behind this change?

enochakinbode: same as the advancedFilter, with constant_long the hash overlaps with the copy button

PR #3138, opened 2025-11-18
Private mode

### LINKED ISSUE #3112: Disable tracking for .onion domain
**Goal**
As a part of _Tor in Ethereum Ecosystem_ initiative: https://igor53627.github.io/tor-ethereum-ecosystem/, ensure Blockscout .onion instances are fully privacy preserving by disabling analytics, ads, and any user tracking mechanisms. 

**ToDo:**
- Blockscout should either auto-detect when accessed via a .onion host or have a single configuration flag per deployment to turn on “Onion Privacy Mode”. 
- Disable all web analytics tools for .onion (GA, Mixpanel, Sentry, etc.)
- Custom tracking pixels should be disabled too
- Hide all ads 
- Turn off Dapps module - to avoid any external iframes 
- Leave only minimal session cookies required for core functionality
- Disable Connect wallet related capabilities

isstuev:

resolves #3112 

### Proposed Changes
adds private mode that can be enabled by passing a certain header to frontend requests. can be easily used later for implementing a user-toggled private mode setting.

## Checklist for PR author

### DIFF
--- a/configs/app/features/account.ts
+++ b/configs/app/features/account.ts
@@ -9,7 +9,7 @@ const title = 'My account';
 const config: Feature<{ isEnabled: true; recaptchaSiteKey: string }> = (() => {
 
   if (
-    app.appProfile !== 'private' &&
+    !app.isPrivateMode &&
     getEnvValue('NEXT_PUBLIC_IS_ACCOUNT_SUPPORTED') === 'true' &&
     services.reCaptchaV2.siteKey
   ) {
--- a/nextjs/csp/index.ts
+++ b/nextjs/csp/index.ts
@@ -36,5 +36,5 @@ export async function get(req?: NextRequest): Promise<string> {
 
   const isPrivateMode = appProfile === 'private';
 
-  return isPrivateMode ? cspPolicies!.private : cspPolicies!.default;
+  return isPrivateMode ? cspPolicies?.private || '' : cspPolicies?.default || '';
 }
--- a/ui/shared/reCaptcha/ReCaptcha.tsx
+++ b/ui/shared/reCaptcha/ReCaptcha.tsx
@@ -40,7 +40,7 @@ const ReCaptchaInvisible = ({ onInitError, hideWarning = false }: Props, ref: Re
     }
   }, [ ]);
 
-  if (!chainConfig.services.reCaptchaV2.siteKey || config.app.appProfile === 'private') {
+  if (!chainConfig.services.reCaptchaV2.siteKey || config.app.isPrivateMode) {
     return null;
   }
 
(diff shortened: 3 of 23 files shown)


configs/app/features/account.ts
```diff
@@ -1,12 +1,18 @@
...
 import services from '../services';
 import { getEnvValue } from '../utils';
 
 const title = 'My account';
 
 const config: Feature<{ isEnabled: true; recaptchaSiteKey: string }> = (() => {
-  if (getEnvValue('NEXT_PUBLIC_IS_ACCOUNT_SUPPORTED') === 'true' && services.reCaptchaV2.siteKey) {
+
+  if (
+    app.appProfile !== 'private' &&
```
tom2drum: It might be better to store the boolean flag in the config instead of comparing the `appProfile` value to the same predefined string every time.

nextjs/csp/index.ts
```diff
@@ -1,19 +1,40 @@
...
+  const appProfile = req ? (
+    req.headers.get('x-app-profile') ||
+    req.nextUrl.searchParams.get('app-profile') ||
+    cookiesLib.getFromCookieString(req.headers.get('cookie') || '', cookiesLib.NAMES.APP_PROFILE)
+  ) : undefined;
+
+  const isPrivateMode = appProfile === 'private';
 
-  return cspPolicy;
+  return isPrivateMode ? cspPolicies!.private : cspPolicies!.default;
```
tom2drum: To enhance safety.

```suggestion
  return isPrivateMode ? cspPolicies?.private : cspPolicies?.default;
```

ui/shared/reCaptcha/ReCaptcha.tsx
```diff
@@ -40,7 +40,7 @@ const ReCaptchaInvisible = ({ onInitError, hideWarning = false }: Props, ref: Re
     }
   }, [ ]);
 
-  if (!chainConfig.services.reCaptchaV2.siteKey) {
+  if (!chainConfig.services.reCaptchaV2.siteKey || config.app.appProfile === 'private') {
```
tom2drum: Why we just don't render ReCaptcha and not disable it completely for entire app?
There will be no way to bypass rate limits in private mode?

ui/shared/reCaptcha/useReCaptcha.tsx
```diff
@@ -1,25 +1,31 @@
...
 export default function useReCaptcha() {
+  const isDisabled = config.app.appProfile === 'private';
   const ref = React.useRef<ReCAPTCHA>(null);
   const rejectCb = React.useRef<((error: Error) => void) | null>(null);
 
   const [ isOpen, setIsOpen ] = React.useState(false);
   const [ isInitError, setIsInitError ] = React.useState(false);
 
   const executeAsync: () => Promise<string | null> = React.useCallback(async() => {
+    if (isDisabled) {
```
tom2drum, line 17: I guess it is better to throw an error then silently resolve the promise.

ui/shared/reCaptcha/ReCaptcha.tsx
isstuev: There will be no way to bypass rate limits in private mode, but it should not affect real users.

PR #3153, opened 2025-11-25
testnet label and network logo fixes

### LINKED ISSUE #3147: testnet label fixes
- desktop horizontal navigation: remove fixed logo width so there's no additional margin between logo and testnet label when logo's width is less than 120px
- mobile navigation drawer: add bottom margin to the testnet label
- mobile header: for testnets use icon instead of the logo and add testnet label

isstuev:

fixes #3147 

### Additional Information
Also I refactored NetworkLogo component since it was a bit overengineered 

## Checklist for PR author


toolkit/chakra/image.tsx
```diff
@@ -6,11 +6,14 @@ import { Skeleton } from './skeleton';
 
 export interface ImageProps extends ChakraImageProps {
   fallback?: React.ReactNode;
+  // for the case where the image dimensions are not known before the image is loaded
+  skeletonWidth?: BoxProps['width'];
+  skeletonHeight?: BoxProps['height'];
```
tom2drum, line 11: Since `skeletonWidth` without `skeletonHeight` doesn't make much sense (if I understand the comment correctly), it may be better to combine these props into a single object.


```suggestion
  // for the case where the image dimensions are not known before the image is loaded
  skeletonProps?: BoxProps;
```

isstuev: skeleton props are inherited from the component props, but can be overwritten by those -- I think it's clearer what are they for in case of width/height, not just generic 'props'

Issue #2293, opened 2024-10-16
Tx page: show historical $ value


NikitaSavik:
[Backend task](https://github.com/blockscout/blockscout/issues/10867)
[Design task](https://www.notion.so/blockscout/Design-For-Estimated-Value-on-Day-1133d73641f880cf8f11ff93efce9558?d=1213d73641f880269f04001c7a5262dd)

TLDR:
Show historically accurate USD value among the 'current' one
![Image](https://github.com/user-attachments/assets/e9087cd2-b3fa-4b3d-895c-e955ff5f2ffd)

PR #3272, opened 2026-02-09
Tx page: show historical $ value

isstuev:

resolves https://github.com/blockscout/frontend/issues/2293

### Additional Information
historical exchange rate would be set to null for the latest txs in the 10.0 backend release

## Checklist for PR author

---

> [!NOTE]
> **Medium Risk**
> Touches shared value/fee rendering used across transaction pages and lists, adding new toggle state and a new `historic_exchange_rate` field; mistakes could cause incorrect USD display or inconsistent tooltips across many views.
> 
> **Overview**
> Adds support for a `historic_exchange_rate` on `Transaction` data (types, RPC formatting, and stubs/mocks) so the UI can distinguish *no historical rate provided* vs *explicitly null*.
> 
> Updates `AssetValue`/`SimpleValue` (and wiring through `NativeCoinValue`/`TxFee`) to display USD using the historical rate by default when available, with an optional clickable tag to toggle between historical and current USD values and clearer tooltip labeling. Propagates this through multiple tx views (tx details, fees/burnt fees, rollup fee rows, and tx list/table items), and adds Playwright screenshot coverage for the new `AssetValue` behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a09eeb02698a77a8abeccfc05b1015b0e094791f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ui/shared/value/AssetValue.tsx
+++ b/ui/shared/value/AssetValue.tsx
@@ -1,5 +1,5 @@
 import type { BoxProps } from '@chakra-ui/react';
-import { chakra } from '@chakra-ui/react';
+import { Box, chakra } from '@chakra-ui/react';
 import React from 'react';
 
 import { Tag } from 'toolkit/chakra/tag';
@@ -20,7 +20,7 @@ export interface Props extends Omit<BoxProps, 'prefix' | 'suffix'>, Omit<Calcula
   loading?: boolean;
   layout?: 'horizontal' | 'vertical';
   tooltipContent?: React.ReactNode;
-  historicExchangeRate?: string | null;
+  historicalExchangeRate?: string | null;
   hasExchangeRateToggle?: boolean;
 }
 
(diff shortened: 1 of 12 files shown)


ui/shared/value/AssetValue.tsx
```diff
@@ -31,18 +35,37 @@ const AssetValue = ({
...
+  hasExchangeRateToggle,
   layout = 'horizontal',
   tooltipContent,
   ...rest
 }: Props) => {
+  const hasHistoric = Boolean(historicExchangeRate);
+  const hasCurrent = Boolean(exchangeRate);
+  const hasToggle = hasHistoric && hasCurrent && hasExchangeRateToggle && amount !== '0';
+
+  const [ showHistoric, setShowHistoric ] = React.useState(hasHistoric);
```
tom2drum: I believe the correct adjective will be "historical" not "historic"

> In general, historic means ‘notable in history, significant in history,’ as in a Supreme Court decision, a battlefield, or a great discovery. Historical means ‘relating to history or past events’: (‘historical society’; ‘historical documents’). To write historic instead of historical may imply a greater significance than is warranted: a historical lecture may simply tell about something that happened, whereas a historic lecture would in some way change the course of human events.

ui/shared/value/SimpleValue.tsx
```diff
@@ -40,12 +42,19 @@ const SimpleValue = ({
...
-      <Box display="inline" whiteSpace="wrap" wordBreak="break-all">
-        { prefix ?? '' }{ value.toFormat() }{ postfix ?? '' }
-        <CopyToClipboard text={ value.toFixed() } verticalAlign="bottom" noTooltip/>
-      </Box>
+      <>
+        { Boolean(tooltipAdditionalContent) && (
+          <Box>
+            { tooltipAdditionalContent }
+          </Box>
+        ) }
```
tom2drum: To simplify the code, we can define `tooltipContentBefore` as a React.ReactNode element. This change will make the prop more extensible and flexible. 

Additionally, I believe `tooltipContentBefore` is a better name than simply `tooltipAdditionalContent`.

```suggestion
        { tooltipContentBefore }
```

ui/shared/value/AssetValue.tsx
```diff
@@ -18,6 +20,8 @@ export interface Props extends Omit<BoxProps, 'prefix' | 'suffix'>, Omit<Calcula
   loading?: boolean;
   layout?: 'horizontal' | 'vertical';
   tooltipContent?: React.ReactNode;
+  historicExchangeRate?: string | null;
+  hasExchangeRateToggle?: boolean;
```
tom2drum, line 24: For me, having the `hasExchangeRateToggle` is unnecessary here.

In components where the toggle should be shown, we can simply pass both exchange rates. 
For components without the toggle, we can pass `exchangeRate={ historicalExchangeRate || exchangeRate }`, ensuring that whatever data exists will be used for calculation.

ui/shared/value/AssetValue.tsx
isstuev: it's the name from the API

ui/shared/value/AssetValue.tsx
isstuev: no, that won't work. historic and current have different tooltips. and there are places like tx lists, where we don't show toggle but show current for recent txs and historic for older ones, so they should be different props

ui/shared/value/AssetValue.tsx
tom2drum: Nevertheless, I don't think we should copy their error into our code.

tom2drum (approved):

PR #3303, opened 2026-02-27
new text ad provider

### LINKED ISSUE #3302: new text ad provider

isstuev:

resolves #3302 

### Proposed Changes
no env changed, ad configured via sevio (but partner is still coinzilla so we decided to keep the name in code)

## Checklist for PR author

### DIFF
--- a/ui/shared/ad/SevioTextAd.tsx
+++ b/ui/shared/ad/SevioTextAd.tsx
@@ -49,6 +49,11 @@ const SevioTextAd = ({ className }: { className?: string }) => {
 
     return () => {
       observer.disconnect();
+    };
+  }, []);
+
+  useEffect(() => {
+    return () => {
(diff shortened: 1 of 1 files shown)


ui/shared/ad/SevioTextAd.tsx
```diff
@@ -0,0 +1,121 @@
...
+    window.sevioads.push([ { zone, adType, inventoryId, accountId } ]);
+  }, []);
+
+  // Watch for Sevio injecting content into the ad div.
+  // When it does, cancel the timeout and mark the ad as loaded.
+  useEffect(() => {
+    const node = adRef.current;
+    if (!node) {
+      return;
+    }
```
tom2drum, line 38: Shouldn't we clear timeout in this case? I guess it is better to have a separate `useEffect` hook that will responsible only for timeout cleaning.

PR #3353, opened 2026-04-17
Use sockets to update latest block/batch on home page

### LINKED ISSUE #3245: Main page. Replace `Total blocks` with dynamic `Latest block`
## Description

Rename `Total blocks` to `Latest block`. Dynamically update the value with the latest block number from WS, same as `Latest blocks` list.

Image

 
Add dynamic updates to `Latest batch` stat accordingly.

maxaleks:

Resolves https://github.com/blockscout/frontend/issues/3245

## Checklist for PR author

---

> [!NOTE]
> **Medium Risk**
> Introduces live socket-driven cache updates for homepage stats, which can cause stale/incorrect UI if event payloads or query keys don’t match expectations, but changes are localized to the homepage widgets.
> 
> **Overview**
> Homepage stats now derive the **Latest block** value from the `general:homepage_blocks` endpoint (with a stub placeholder) instead of treating it as a "total blocks" metric, and the widget label is updated accordingly (including degraded/fallback UI and mocks).
> 
> Adds websocket subscriptions on the homepage stats panel to push real-time updates into the React Query cache for `blocks:new_block` and (for Arbitrum/zkEVM) latest batch events, with guardrails to disable sockets during placeholder/error states and to de-duplicate/sort incoming items.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2338b24ce3f329207842864cd1d45e89de5858b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/home/Stats.tsx
```diff
@@ -40,7 +52,16 @@ const Stats = () => {
     },
   });
 
-  const isPlaceholderData = statsQuery.isPlaceholderData || apiQuery.isPlaceholderData;
+  const latestBlockQueryKey = React.useMemo(() => [ ...getResourceKey('general:homepage_blocks'), 'stats_latest_block' ], []);
+  const latestBlockQuery = useApiQuery('general:homepage_blocks', {
```
tom2drum, line 56: We already fetch the latest blocks on this page. Since the query key is different here, will there be two requests on the initial load?

Will the data be fetched on rollup instances where we don't display block information?

ui/home/Stats.tsx
tom2drum, line 1: I think we need to restructure this component.

Since updates to the latest block number can occur very frequently, the entire component will re-render, which is inefficient. It would be better to move the LatestBlock widget to a separate component and encapsulate all the necessary logic there. This way, the re-render will only affect that specific part of the tree.

tom2drum (changes requested):

ui/home/homeDataContext.tsx
```diff
@@ -0,0 +1,205 @@
...
+type LatestBatchSocketMessage = LatestBatchSocketEventMessage | SocketMessage.Unknown;
+
+type HomeDataContextValue = {
+  blocksQuery: HomeBlocksQueryResult;
+  latestBatchQuery: HomeLatestBatchQueryResult | undefined;
+};
+
+const HomeDataContext = React.createContext<HomeDataContextValue | null>(null);
+
+function useHomeBlocksData(): HomeBlocksQueryResult {
```
tom2drum: Please put these hooks into separate files.

ui/home/homeDataContext.tsx
```diff
@@ -0,0 +1,205 @@
...
+  }, [ queryClient, latestBatchSocketConfig ]);
+
+  const latestBatchChannel = useSocketChannel({
+    topic: latestBatchSocketConfig?.topic,
+    isDisabled: Boolean(latestBatchSocketDisabled),
+  });
+  useSocketMessage({
+    channel: latestBatchChannel,
+    event: latestBatchSocketConfig?.event,
+    handler: handleNewLatestBatchMessage as LatestBatchHandler,
```
tom2drum: I believe this type coercion can be safely removed because we specify the type for the entire object below.

ui/home/homeDataContext.tsx
```diff
@@ -0,0 +1,205 @@
...
+  return ctx;
+}
+
+export function useHomeBlocksQuery(): HomeBlocksQueryResult {
+  return useHomeDataContext().blocksQuery;
+}
+
+export function useHomeLatestBatchQuery(): HomeLatestBatchQueryResult | undefined {
+  return useHomeDataContext().latestBatchQuery;
+}
```
tom2drum: Not sure if these hooks are useful, since they are used one time in the code.

ui/home/Stats.tsx
```diff
@@ -117,19 +87,16 @@ const Stats = () => {
     return [
       latestBatchQuery?.data !== undefined && {
```
tom2drum, line 88: Let's check here also if rollup feature is enabled. The query has the placeholder data, so the condition almost always true.

ui/home/Stats.tsx
```diff
@@ -221,14 +188,22 @@ const Stats = () => {
...
+        if ('component' in item) {
+          return React.cloneElement(item.component as React.ReactElement<{ _last?: { gridColumn?: string } }>, { key: item.id, _last });
+        }
+
+        return (
+          <StatsWidget
+            key={ item.id }
+            { ...item }
+            isLoading={ isLoading }
+            _last={ _last }
```
tom2drum: Let's implement it the CSS way without cloning the component; we can simply define common properties for all items.

```suggestion
   _odd={{
      _last: { gridColumn: 'span 2' },
    }}
```

ui/home/utils.ts
```diff
@@ -1,15 +1,18 @@
...
-export interface HomeStatsItem extends StatsWidgetProps {
-  id: HomeStatsWidgetId;
-}
+export type HomeStatsComponentItem = { id: HomeStatsWidgetId; component: ReactElement };
+export type HomeStatsWidgetItem = StatsWidgetProps & { id: HomeStatsWidgetId; component?: undefined };
+
+export type HomeStatsItem = HomeStatsComponentItem | HomeStatsWidgetItem;
 
-export const isHomeStatsItemEnabled = (item: HomeStatsItem) => config.UI.homepage.stats.includes(item.id);
+export const isHomeStatsItemEnabled = <T extends { id: HomeStatsWidgetId }>(item: T) => config.UI.homepage.stats.includes(item.id);
```
tom2drum:
```suggestion
export const isHomeStatsItemEnabled = (item: { id: HomeStatsWidgetId }) => config.UI.homepage.stats.includes(item.id);
```

ui/home/utils.ts
```diff
@@ -1,15 +1,18 @@
...
 
-export const sortHomeStatsItems = (a: HomeStatsItem, b: HomeStatsItem) => {
+export const sortHomeStatsItems = <T extends { id: HomeStatsWidgetId }>(a: T, b: T) => {
```
tom2drum: Same suggestion

tom2drum (changes requested):
From the user's perspective, I think we need to update the total number of transactions as a new block arrives. The necessary data is included in the socket message payload. Otherwise, it seems strange that the latest block keeps changing while the transaction number remains unchanged, making it appear as if all new blocks are empty.

ui/home/Stats.tsx
maxaleks: If the rollup feature is not enabled, the `latestBatchQuery` will be undefined

ui/home/LatestBlockStatsWidget.tsx
```diff
@@ -0,0 +1,34 @@
...
+  className?: string;
+  isLoading: boolean;
+  fallbackValue: number | string | undefined;
+};
+
+const LatestBlockStatsWidget = ({ className, isLoading, fallbackValue }: Props) => {
+  const { blocksQuery } = useHomeDataContext();
+
+  const value = blocksQuery.data?.[0]?.height ?? fallbackValue;
+  if (value == null) {
```
tom2drum: Let's use strict comparison operator (`===`, I will add it to the rules later).
The typescript says that the value can be `number | string | undefined`, so comparing against `null` is confusing right now.

ui/home/Stats.tsx
tom2drum: Oh, I see 👍

tom2drum: Then let's do the same thing for `blocksQuery`: when it is not enabled let's return `undefined` from the hook instead of the pending query.

tom2drum (commented):
All good now 👍

ui/home/LatestBlockStatsWidget.tsx
tom2drum: https://github.com/blockscout/frontend/commit/54e31e40ca859c1ddccf6b227fd568333b4a434d

tom2drum (approved):

PR #3428, opened 2026-05-15
fix adblock detection and CSP for Sevio

maxaleks:

- replace url for adblock detection
- fix CSP for Sevio, added nonce

## Checklist for PR author

### DIFF
--- a/nextjs/csp/index.ts
+++ b/nextjs/csp/index.ts
@@ -7,9 +7,22 @@ import generateNftHtmlEmbedCspPolicy from './generateNftHtmlEmbedCspPolicy';
 
 const NFT_HTML_EMBED_PATH = '/nft-html-embed.html';
 
+let cspPolicies: { 'private': string; 'default': string } | undefined = undefined;
 let nftHtmlEmbedCsp: string | undefined = undefined;
 
+async function initializeCspPolicies() {
+  if (!cspPolicies) {
+    // Generate and cache both policies upfront
+    cspPolicies = {
+      'private': generateCspPolicy(true),
+      'default': generateCspPolicy(false),
+    };
+  }
+}
+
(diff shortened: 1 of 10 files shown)


nextjs/csp/index.ts
```diff
@@ -7,22 +7,9 @@ import generateNftHtmlEmbedCspPolicy from './generateNftHtmlEmbedCspPolicy';
...
-      'private': generateCspPolicy(true),
-      'default': generateCspPolicy(false),
-    };
-  }
-}
-
-export async function get(req?: NextRequest): Promise<string> {
-  await initializeCspPolicies();
-
+export async function get(req?: NextRequest, nonce?: string): Promise<string> {
```
tom2drum, line 23: What is the reason for deleting the initialize function and policy cache? If the nonce is not present, we should use the default policies, I believe.

proxy.ts
```diff
@@ -19,7 +20,15 @@ export async function proxy(req: NextRequest) {
     return accountResponse;
   }
 
-  const res = NextResponse.next();
+  const nonce = crypto.randomUUID().replaceAll('-', '');
```
tom2drum: If there is no servio ads on the instance, we don't need to generate nonce.

proxy.ts
```diff
@@ -19,7 +20,15 @@ export async function proxy(req: NextRequest) {
...
+  const requestHeaders = new Headers(req.headers);
```
tom2drum: Why do we need to copy all request headers? Why not just append the nonce header to `res` later, as is done for the other headers?

tom2drum (approved):

tom2drum (approved):

tom2drum (approved):

Issue #3566, opened 2026-07-15
Main page: improve initial loading performance


tom2drum:
## Problem

Profiling the main page of a production instance (desktop, fast network, clean browser profile) shows that the first content paints at ~1.1 s and the actual chain data appears only at ~2.5–3 s. Three structural issues contribute:

1. **Data fetching waits for the JS boot.** The main-page API requests start only after the whole client bundle is downloaded, evaluated, and the app is hydrated (~1 s in). Backend response time is therefore added *on top of* the boot instead of overlapping with it.

2. **Third-party SDKs sit in the critical bundle.** The analytics and error-monitoring SDKs (Mixpanel ~103 KB gz, Rollbar ~25 KB gz) are downloaded, evaluated, and initialized during boot, although nothing they do is needed before first paint.

3. **The wallet stack gates first render.** The web3 provider stack (AppKit + wagmi + viem and their dependencies, ~150 KB gz) wraps the whole app tree as a client-only dynamic component, so the first paint waits for its chunks to download and evaluate — even on pages with no wallet functionality in view.

Additionally, the `react-icons` dependency is pulled in for just two icons in the toolkit, which the app's own SVG sprite already covers.

## What can be done

- Start the main-page API requests from the HTML document itself so they run in parallel with the JS boot — time-to-content becomes `max(boot, backend)` instead of `boot + backend`.
- Defer loading of the analytics / error-monitoring SDKs until after first paint, with small queues so no events or error reports are lost.
- Defer the wallet stack: stop blocking first render on it, and load it on idle or on the routes/interactions that actually need it.
- Replace the `react-icons` usage with sprite icons and drop the dependency.

## Expected impact

| Metric (main page, desktop, fast network) | Current | Estimated after fixes |
|---|---|---|
| First / Largest Contentful Paint | ~1.1 s | ~0.7 s |
| First data request starts | ~1.1 s | ~0.15 s |
| Main content rendered | ~2.5–3 s | ~1.7–2 s |
| Critical-path JS (gzipped) | ~2.0 MB | ~1.7 MB |
| Total blocking time | ~0.65 s | ~0.4 s |

Numbers are from profiling a single instance and will vary with backend latency and hardware; mobile devices should see proportionally larger gains since most of the cost is CPU-bound script evaluation.

Related, but tracked separately: list-rendering performance work (#3501) that also improves the main page's data-render cost.

tom2drum:
## Final measurement (baseline vs final)

[...]

PR #3570, opened 2026-07-15
Main page: early-fetch primer for first-render API requests

tom2drum:

Part of #3566 — **subtask 1** of the [main-page loading-performance task](https://github.com/blockscout/frontend/blob/issue-3566-step-1/.agents/tasks/3566-main-page-loading-perf/spec.md), targeting the `issue-3566` feature branch (this is a stacked PR, not into `main`).

Fires an entry page's first-render API requests from an inline script in the HTML `<head>`, while the JS bundle is still downloading and evaluating, and consumes each primed response exactly once in the fetch layer — so backend latency overlaps the boot instead of stacking on top of it. It is purely a timing optimization: consume-once semantics plus a consume-time URL/header comparison that falls back to a normal fetch on any mismatch mean the primer can change *when* a request happens, never *whether* or *with what*.

Sub-spec with goal, inputs, design decisions, verification notes, and follow-ups: [`.agents/tasks/3566-main-page-loading-perf/subtasks/01-early-fetch-primer.md`](https://github.com/blockscout/frontend/blob/issue-3566-step-1/.agents/tasks/3566-main-page-loading-perf/subtasks/01-early-fetch-primer.md). Module architecture: [`src/server/primedRequests/CONTEXT.md`](https://github.com/blockscout/frontend/blob/issue-3566-step-1/src/server/primedRequests/CONTEXT.md).

### Proposed Changes

- **Generic per-page primer** (`src/server/primedRequests/`): a registry keyed by the Next.js page id maps each page to the first-render requests its widgets make. URL building, headers (including cookie-derived ones, resolved in the browser), CSP allowance, and fetch-layer consumption are all generic.
- **Deterministic inline script**: its `script-src` sha256 hash is computed once at server startup, so no request-specific value is baked in — route params, the active tab, and forwarded URL query filters are resolved in the browser from `location`.
- **Rolled out to** home, tx, address, token, token instance, block, user ops, tokens, and chain stats — multi-tab pages prime the default tab only. Added browser-side query-param forwarding so the tokens list primes with its URL filters (`q`, `type`, `sort`, `order`), byte-matched to the client's request.
- **Drift protection**: a colocated `*.primed.spec.tsx` per page asserts every primed request is one the real page fires on first render (byte-identical URL + headers); a completeness test fails if a registered page lacks its spec. Supported by a vitest harness that mounts whole pages and a `withEnvs` config-override utility shared with the Playwright suite.

No environment variable changes.

None. No behavior change by construction — every subsequent request for a primed resource (refetch, pagination, socket-driven) goes to the network as usual.

### Additional Information

Single production-run measurement (home page) recorded in the spec's impact table: first API request ~916 ms → ~90 ms, content rendered ~2534 ms → ~1307 ms. A full median re-measurement across the newly primed pages is a tracked follow-up before the parent `issue-3566` merges.

## Checklist for PR author


tom2drum:
bugbot run

PR #3596, opened 2026-07-28
Show transaction status, action, and timestamp in social link previews

### LINKED ISSUE #3593: Generate transaction OG title and description from transaction details
Implement transaction OG title and description generation.

Use this format:

```text
{chain_name} transaction {tx_hash_short} | Blockscout
{status} · {tx_action_with_ui_rounding} · {timestamp}
```

Details:
- Short transaction hash format: `0x04...3e1d`.
- Timestamp format: `MMM D, YYYY H:mm UTC`.
- Transaction action should match the Transaction details UI, including fallback behavior and amount rounding.

Live example:
https://robinhoodchain.blockscout.com/tx/0xdae46c25802fc94d2384d0ef0a9f89733ed350269d193426960b0abd7a4e671a

[Backend task](https://github.com/blockscout/blockscout/pull/14638)
**Robinhood Chain transaction 0xda...671a | Blockscout**
Success · Swap 2.92M SPERPS for 0.016 WETH · Jul 28, 2026 1:43 UTC

tom2drum:
## Description

Resolves #3593.

A shared transaction link produced a near-useless card: the OG title carried the full 66-character hash and
there was no `og:description` at all, so Telegram and X fell back to the generic page description. The card now
describes the transaction:

```text
Ethereum transaction 0xc1...cc2e | Blockscout
Success · Transfer 0.013 ETH to Fee Recipient: 0xadfb...e35 · Jul 29, 2026 14:04 UTC
```

The action reads exactly as the transaction page's subheading, including its fallback chain and amount
rounding, because both now render from the same code.

Reaching that needed two structural changes. The `og` block in the metadata template map became a real
template layer with `default`/`enhanced` variants (previously `og.description` was passed through raw and never
compiled), and `/tx/[hash]`'s `getServerSideProps` gained a bot-gated server-side fetch — crawlers don't run JS,
and `metadata.update()` only ever touches `<title>` and `<meta name="description">`.

Decisions worth knowing, all recorded in [the spec](.agents/tasks/3593-tx-og-title-description/spec.md):

- **SEO tags are untouched.** `<title>` keeps the full hash and `<meta name="description">` keeps the generic
  copy; only social-preview crawlers get the enhanced description. `generate()` has no notion of bot type — the
  presence of `apiData` is the only signal.
- **Enhanced data comes from a purpose-built endpoint.** `core:tx` supplies the status and timestamp the
  description cannot do without, and on a loaded instance it answered too slowly for a crawler (aborted on 6 of
  6 requests at a 2 s timeout). The Core API team added
  `/api/v2/transactions/:hash/preview` for this feature; against that same instance it answers in 0.51 s at p50
  and the card now enhances on 6 of 6. Raising the timeout instead was rejected: crawlers wait single-digit
  seconds, and a card that fails to render is worse than a plain one.
- **The summary stays a separate request.** Folding it into the preview response was offered and declined —
  `/summary` is by far the slowest call, so coupling them would turn every slow summary into a *lost*
  description instead of one that degrades to `0x9a...71a3 called transfer on FiatTokenProxy`.
- **All-or-nothing, by the existing `compileValue` contract.** Accepted losses: a pending transaction (no
  timestamp), an unresolvable action, and any failed, timed-out or 404 request — each falls back to today's card.
- **Nothing is requested where nothing can come of it.** An instance with the interpretation feature off, or
  running the Noves provider, makes no request at all: there is no action to be had, and the PM ruled that
  quietly pointing crawler traffic at a third party's slow API isn't ours to do.
- **The X card type now follows what the page has.** Every page claimed `summary_large_image`, but entity
  pages carry no image on purpose, and X answers a promised-but-missing image with the compact card and a grey
  placeholder where the image would be. The type is now `summary` when there is no image, so the card renders
  as title and description alone. Applies to every imageless route, not just transactions.
- **WhatsApp, Discord and LinkedIn were missing from bot detection** — a pre-existing gap that made them fall
  back to the generic description on *every* OG-enhanced route (address, token, NFT, stats), not just this one.
  Fixed here since WhatsApp is where it was noticed.

Also included: the currency rounding and summary-to-text rendering extracted from `TxInterpretation` so the
preview and the page cannot drift, and `getAddressName` extracted from `AddressEntity` so an address is labelled
the same way in both.

## Environment variables

None. The feature reuses the existing `NEXT_PUBLIC_OG_ENHANCED_DATA_ENABLED`, which defaults to on.

## Minimum API version

The enhanced description needs `GET /api/v2/transactions/:hash/preview`, added in
blockscout/blockscout#14638 with a follow-up fix in #14703. Both are merged but **not in a released version
yet** (latest release is v11.2.6), so the requirement is the first Core API release that contains them.

This is not a hard requirement: on an instance whose backend predates the endpoint the request fails and the
card keeps the generic description, exactly as it does today when a request times out. Deliberately no
fallback to `core:tx` — chaining a second attempt would spend a crawler's patience precisely on the instances
that are already slow.

None for deployments.

One deliberate output change: routes with no OG description template now emit `og:description` explicitly,
carrying the same text crawlers already inferred from `<meta name="description">`. This is visible in
`src/shell/metadata/__snapshots__/generate.spec.ts.snap` and was chosen over an implicit fallback so the tag is
always present. Every other route's `title` / `description` / `og:title` / `og:image` output is byte-identical.

## Additional information

- Verified end to end on two demos built from this branch, plus real cards in Telegram and X.
- `@blockscout/api-types` is pinned to a beta (`0.0.1-beta.eeda17a`) because the endpoint's schema is not in a
  stable release yet; it needs re-pinning to the stable version once the Core API release lands.
- The schemas were named `Preview` / `PreviewAddress` upstream, which said nothing about what they previewed
  once generated into a spec whose component names are global; renamed to `TransactionPreview` /
  `TransactionPreviewAddress` in blockscout/blockscout#14711 before anything depended on them.


tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `73171c99c382c3ff3e173f4784b2cd5d811f856c` → `9507cc0eba8f549e74a730865f9d07a8d805d559` |
| Round | 1 of 3 |
| Findings | 0 blocker · 5 major · 3 nit |
| By axis | Spec 3 · Standards 4 · Correctness 1 |
| Outcome | `needs-human` |

Whole-task review of #3593 (PR #3596).

src/server/PageMetadata.tsx
```diff
@@ -32,11 +32,20 @@ const PageMetadata = <Pathname extends Route['pathname']>(props: Props<Pathname>
...
-      { opengraph.imageUrl && <meta property="twitter:image" content={ opengraph.imageUrl }/> }
+      <meta property="twitter:domain" content={ config.app.host }/>
+      { opengraph.imageUrl ? (
+        <>
+          <meta name="twitter:card" content="summary_large_image"/>
+          <meta property="twitter:image" content={ opengraph.imageUrl }/>
+        </>
+      ) : (
+        <>
+          <meta name="twitter:card" content="summary"/>
```
tom2drum, line 45: **major** — Switches `twitter:card` to `summary` and injects a favicon `twitter:image` whenever `opengraph.imageUrl` is absent — site-wide, including every entity detail route.

Parent out of scope still says text-only / no image work (`og:image` — "Its current absence is deliberate"; `summary_large_image` was the reason). This was never an AC.

Drop these tags back to always-`summary_large_image` / image-only-when-present, or amend the parent AC if the X-card behaviour is intentional.

— Reviewed by Cursor Grok 4.5

src/server/utils/detectBotRequest.ts
```diff
@@ -36,6 +36,20 @@ export default function detectBotRequest(req: IncomingMessage): ReturnType {
     return { type: 'social_preview', bot: 'slack' };
   }
 
+  if (userAgent.toLowerCase().includes('whatsapp')) {
```
tom2drum, line 39: **major** — Adds WhatsApp / Discord / LinkedIn as `social_preview`, so every OG-enhanced gSSP (tx, token, …) now fetches for those UAs.

Parent FR only gates on existing `detectBotRequest` → `social_preview` and never expands the set; demo ACs named Telegram and X.

Revert the new matchers, or add an AC that names these bots.

— Reviewed by Cursor Grok 4.5

src/shell/metadata/types.ts
```diff
@@ -7,10 +7,13 @@ import type { MarketplaceDapp } from '@blockscout/admin-rs-types';
 import type { schemas } from '@blockscout/api-types';
 import type { LineChart } from '@blockscout/stats-types';
 
+import type { TxOgDescriptionParams } from 'src/slices/tx/utils/get-og-description-params';
```
tom2drum: **major** — `import type { TxOgDescriptionParams } from 'src/slices/tx/utils/get-og-description-params'` — `architecture.md`: "Cross-domain type imports go through `<slice|feature>/types/api.ts`, never deeper internal paths."

Move `TxOgDescriptionParams` to `src/slices/tx/types/api.ts` (or inline like address `ApiData`) and import from there.

— Reviewed by Cursor Grok 4.5

src/slices/tx/utils/get-og-description-params.ts
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: LicenseRef-Blockscout
+
+import type { schemas } from '@blockscout/api-types';
+import type { TxInterpretationResponse } from 'src/features/tx-interpretation/common/types/api';
+
+import addressToPlainText from 'src/features/tx-interpretation/common/utils/address-to-plain-text';
```
tom2drum, line 7: **major** — Slice util imports `addressToPlainText` / `summaryToPlainText` from the feature — `architecture.md`: "Compose at the page level when optional feature UI plugs into a core slice … not the other way around." Also Feature Envy: the util mostly manipulates the feature's data.

Keep status/timestamp derivation in the slice (or page); put interpretation→action text in the feature and compose at the page.

— Reviewed by Cursor Grok 4.5

src/slices/tx/utils/get-og-description-params.ts
```diff
@@ -0,0 +1,80 @@
...
+const TIMESTAMP_FORMAT = 'lll';
+
+export interface TxOgDescriptionParams {
+  tx_status: string;
+  tx_action: string;
+  tx_timestamp: string;
+}
+
+// `undefined` — as opposed to `null`, which is a pending transaction — means the field never arrived, so
+// there is no status to show. `fetchApi` hands back non-200 bodies as data, and this is what keeps a 404
```
tom2drum: **major** — Comment says "`fetchApi` hands back non-200 bodies as data" / keeps a 404 from reading as `Pending` — but `src/server/utils/fetchApi.ts` returns `undefined` on non-200 (since #3623); 404 never yields `{ message }` as `TransactionPreview`.

The matching test at `get-og-description-params.spec.ts` ("an error body, which `fetchApi` hands back as data") asserts that fictional contract.

Rewrite the comment for `undefined`-on-error; drop the error-body case (or assert `getOgDescriptionParams(undefined, …) === null` — already covered by "no transaction at all").

— Reviewed by Cursor Grok 4.5

src/features/tx-interpretation/common/utils/address-to-plain-text.ts
```diff
@@ -0,0 +1,15 @@
...
+import getAddressName from 'src/slices/address/utils/get-address-name';
+
+import shortenString from 'src/shared/texts/shorten-string';
+
+// What `truncation="constant"` resolves to in `AddressEntity`.
+const HASH_CHAR_NUMBER = 8;
+
+// How an address reads on the page, minus the display concerns that need a client: the proxy-implementation
+// tooltip and the bech32/Filecoin alt-hash, both driven by user settings unavailable server-side.
+export default function addressToPlainText(address: AddressNameSource) {
```
tom2drum, line 13: **nit** — New util ships with no `*.spec.ts` (hash fallback untested).

Subtask 02 AC: "a spec per new util". Add `address-to-plain-text.spec.ts`.

— Reviewed by Cursor Grok 4.5

src/features/tx-interpretation/common/utils/format-currency-value.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// Below this a two-decimal format would collapse to `0.00`, so significant digits are used instead.
+const SMALL_VALUE_THRESHOLD = 0.1;
+const SIGNIFICANT_DIGITS = 2;
+const DECIMAL_PLACES = 2;
+const THOUSANDS_THRESHOLD = 10 * THOUSAND;
+
+// Shared by the interpretation component and its plain-text renderer so an amount in a social preview
+// reads exactly as it does on the page.
+export default function formatCurrencyValue(value: string) {
```
tom2drum, line 16: **nit** — `export default function formatCurrencyValue(value: string)` (same pattern on `addressToPlainText`, `summaryToPlainText`, `getStatusText`, `getActionText`) — `typescript.md` Return types: "When declaring functions on the top-level of a module, declare their return types."

Add explicit return types (`: string`, `: string | undefined`, etc.).

— Reviewed by Cursor Grok 4.5

src/slices/address/utils/get-address-name.ts
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: LicenseRef-Blockscout
+
+import type { schemas } from '@blockscout/api-types';
+
+import { getTagName } from 'src/features/address-metadata/components/tag/utils';
+
```
tom2drum, line 6: **nit** — `Partial<Pick<…>> & { hash: string }` — `typescript.md` Interface extends: "ALWAYS prefer interfaces when modelling inheritance. The `&` operator has terrible performance…"

`export interface AddressNameSource extends Partial<Pick<…>> { hash: string }`

— Reviewed by Cursor Grok 4.5

src/server/PageMetadata.tsx
tom2drum: Fixed — as an amendment to the parent spec rather than a revert, since the behaviour is intentional. The PM asked for it after seeing real cards: X does not honour an omitted image, it reserves the slot and fills it with a grey placeholder. `9507cc0eb` therefore drops those routes to `summary` and points `twitter:image` at the instance icon; `og:image` stays unset so Telegram and WhatsApp keep the clean text-only card.

The parent spec now records this in a new "Follow-ups from the manual verification" section, and its `og:image` out-of-scope bullet carries the X exception instead of citing `summary_large_image` as settled (`806ce259d`).

src/server/utils/detectBotRequest.ts
tom2drum: Fixed the same way — spec amended, matchers kept. These were a bug fix, not a scope grab: the ACs named Telegram and X because that was what the demo had been checked in, and WhatsApp then turned out to show no preview at all. Discord and LinkedIn deliberately match the `…bot` suffix rather than the bare product name, because both also ship an in-app browser whose user agent carries that name — those are real visitors, not crawlers.

Recorded in the parent spec's "Follow-ups from the manual verification" section (`806ce259d`), including the consequence you flag: every OG-enhanced route now fetches for these three as well.

src/shell/metadata/types.ts
tom2drum: Fixed in `806ce259d` — `TxOgDescriptionParams` moved to `src/slices/tx/types/api.ts`, and both consumers (`src/shell/metadata/types.ts`, the util itself) import it from there.

src/slices/tx/utils/get-og-description-params.ts
tom2drum: Confirmed and fixed in `806ce259d`. `fetchApi` returns `undefined` on a non-200 since #3623, which landed on `main` and reached this branch through a merge mid-task — so the comment was describing a contract that had stopped holding under it.

The `undefined`-vs-`null` distinction stays, though: it is still load-bearing, just for a different reason. A 404, a failure, or a blown timeout now yields no body at all, and `tx?.status` on that is `undefined` — collapse it into `null` and every miss renders as `Pending`. The comment now says that, and the test asserting the old error-body shape is gone (the "no transaction at all" case already covers what remains).

src/slices/tx/utils/get-og-description-params.ts
tom2drum: Rejecting this one — the premise does not hold up.

The quoted rule is about **UI composition**: "the slice page renders the feature's component, not the other way around. Avoids cycles and keeps the feature opt-in." It governs where feature *components* get mounted, not whether slice code may call a feature's pure helpers. `architecture.md` has no rule against a slice importing from a feature, and in practice 160 files under `src/slices/` already do — including `src/slices/address/utils/get-address-name.ts:5`, which reaches into a feature's *component* utils, one level deeper than this does. The one type import here already goes through `common/types/api.ts` as required.

No cycle, either: nothing in `tx-interpretation/common` imports `slices/tx`, and the repo has no barrel files to create one indirectly.

On the Feature Envy reading: the output is OG metadata for the transaction page, which is the slice's concern, and the feature gate lives inside the util so it degrades to the method fallback on its own. Splitting status/timestamp from the action text would put a three-fragment assembly in the page for no gain — the page-level composition the rule protects is for optional UI, which this is not.

— resolve-review

src/features/tx-interpretation/common/utils/address-to-plain-text.ts
tom2drum: Deferred as a nit, not declined — the claim is right (it is the only util in that folder without a spec). Leaving this thread open so it stays visible; it is not this change's work.

src/features/tx-interpretation/common/utils/format-currency-value.ts
tom2drum: Deferred as a nit. Confirmed against `typescript.md`: "When declaring functions on the top-level of a module, declare their return types." Thread left open.

src/slices/address/utils/get-address-name.ts
tom2drum: Deferred as a nit. Confirmed against `typescript.md` — `interface extends` is preferred over `&`. Thread left open.

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `73171c99c382c3ff3e173f4784b2cd5d811f856c` → `b485abf0df8a55a4561d9ead87f91d0d4b9309cf` |
| Round | 2 of 3 |
| Findings | 0 blocker · 0 major open · 3 nit deferred |
| By axis | Spec 0 open · Standards 0 open (1 reject accepted) · Correctness 0 open |
| Outcome | `clear` |

## Arbitration

| Finding | Ruling |
| --- | --- |
| F1 PageMetadata twitter card / favicon image | `fixed-verified` — spec follow-ups + out-of-scope now own the behaviour |
| F2 WhatsApp / Discord / LinkedIn crawlers | `fixed-verified` — recorded in Follow-ups |
| F3 `TxOgDescriptionParams` cross-domain import | `fixed-verified` — now via `slices/tx/types/api.ts` |
| F4 slice → feature plain-text imports | `rejected-accepted` — compose rule is UI composition, not pure helpers |
| F5 `fetchApi` error-body comment/test | `fixed-verified` — comment + trap updated; dead test removed |
| F6–F8 nits | `deferred` (unchanged) |

No regressions from the fix commits. `b485abf0d` (expose `FAVICON_MASTER_URL` to presets) is adjacent tooling for the X card icon, not a fix regression.

— Reviewed by Cursor Grok 4.5

src/slices/tx/utils/get-og-description-params.ts
tom2drum: **rejected-accepted** (R2 arbitration) — Agree. The compose rule targets optional feature *UI* mounted from a slice page (“the slice page renders the feature's component”), not pure helpers. Slice→feature util imports for plain-text rendering do not violate it, and the Feature Envy split would not buy real composition.

— Reviewed by Cursor Grok 4.5

PR #3605, opened 2026-07-30
Move the block countdown page to the API v2 endpoint

### LINKED ISSUE #3583: Migrate block countdown from API v1 `getblockcountdown` to API v2
## Description

The frontend still uses the API v1 block countdown resource (`/api?module=block&action=getblockcountdown&blockno=…`, path plus search params) instead of the API v2 route (`/api/v2/blocks/:block_number/countdown`). The migration was not done earlier because the v2 endpoint was broken; a backend fix is in progress and should land in a patch release.

## Details

- Current (API v1): `https://eth.blockscout.com/api?module=block&action=getblockcountdown&blockno=25599448`
- Target (API v2): `https://eth.blockscout.com/api/v2/blocks/25599448/countdown`
- Backend fix for 422 on the v2 countdown endpoint: https://github.com/blockscout/blockscout/pull/14612
- API docs for this endpoint exist but were not generated yet (redacted-token issue); frontend work should wait on docs being available.

## Expected Behavior

- Use `/api/v2/blocks/:block_number/countdown` for block countdown.
- Drop the API v1 `getblockcountdown` usage once the v2 endpoint and docs are ready.

tom2drum:

Resolves #3583

The block countdown page was the app's last caller of the API v1 RPC endpoint. Its resource declared `path: '/api'` and passed `module=block&action=getblockcountdown&blockno=N` as query params — an RPC call wearing a REST resource's clothes. It now reads `GET /api/v2/blocks/:height/countdown`, and its payload type comes from the generated `@blockscout/api-types` package instead of a hand-written interface.

The v2 route has existed since blockscout/blockscout#12704 but answered 422 for valid input. blockscout/blockscout#14612 fixed that in backend v11.2.3 and renamed the response fields; blockscout/blockscout#14646 then settled the rest of the contract — real status codes for the non-success cases, and string-typed block numbers — and shipped in **v11.2.4**.

### Proposed Changes

- `core:block_countdown` moves out of the API v1 registry into `CORE_API_BLOCK_RESOURCES`, with `path: '/api/v2/blocks/:height/countdown'`. `:height` rather than the siblings' `:height_or_hash`, because this endpoint takes only a non-negative integer.
- The payload type is `paths['/api/v2/blocks/{block_number_param}/countdown']['get']`. The local `BlockCountdownResponse` interface is deleted — nothing outside the registry imported it.
- `src/api/resources/services/core/v1.ts` keeps only `graphql`. `CoreApiV1ResourcePayload` went with the countdown entry: `graphql` never had a payload branch, so the type was left vacuous.
- `BlockCountdown.tsx` reads the four renamed fields and drives its redirect off the response status: **404** ("block already mined") redirects to the block page, exactly as before; every other error reaches the error page.
- The Playwright mocks move to the v2 field names and a `pathParams` matcher. Values stay strings, so both screenshot baselines are unchanged.

No env var changes.

**The countdown page now requires backend v11.2.4 or newer.** On v11.2.3 the endpoint answers 200-with-a-message instead of 404/422 and returns the block numbers as JSON integers, so the page would neither render a countdown nor redirect correctly. No compatibility shim reads both contracts — instances older than v11.2.4 should not expect the countdown page to work.

One deliberate behavior change beyond that: an invalid height (non-numeric or negative) and an indexing chain now land on the error page. Under v1 both bounced the user to a block page that cannot exist.

### Additional Information

Verified against the `staging` preset running v11.2.4:

| Case | Response | Result |
| --- | --- | --- |
| `/block/countdown/99999999` | 200 | countdown renders |
| `/block/countdown/1` | 404 | redirects to `/block/1` |
| `/block/countdown/abc` | 422 | error page |

The 422-while-indexing and the 501 for disabled average block time are not reproducible on a healthy instance; both were verified against the controller and fallback-controller source.

The spec for this task lives in [`.agents/tasks/3583-block-countdown-api-v2/`](https://github.com/blockscout/frontend/tree/issue-3583/.agents/tasks/3583-block-countdown-api-v2) and is included in this PR.

## Checklist for PR author
- [ ] I added tests to cover any new functionality, following this [guide](./CONTRIBUTING.md#writing--running-tests) — the existing Playwright cases cover the rendering; the only new logic is a one-line status check, so a unit test would only re-assert an inline conditional.
- [x] If I have added a feature or functionality that is not privacy-compliant (e.g., tracking, analytics, third-party services), I have disabled it for private mode. — n/a
- [x] If I have added, changed, renamed, or removed an environment variable — n/a, no env var changes


src/api/resources/services/core/v1.ts
```diff
@@ -1,21 +1,11 @@
...
 export const CORE_API_V1_RESOURCES = {
   graphql: {
     path: '/api/v1/graphql',
   },
-  block_countdown: {
-    path: '/api',
-  },
 } satisfies Record<string, ApiResource>;
 
 export type CoreApiV1ResourceName = `core:${ keyof typeof CORE_API_V1_RESOURCES }`;
```
tom2drum, line 11: **nit** — possible Dead Scaffolding (`smells.md`): `export type CoreApiV1ResourceName = …` is unused after `CoreApiV1ResourcePayload` and its `core/index.ts` branch were removed.

Delete the type, or re-wire it into the core payload union if `graphql` still needs a typed payload slot.

— Reviewed by Composer

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `9afd8b689` → `1d6006222` |
| Round | 1 of 3 |
| By axis | Spec 0 · Standards 1 · Correctness 0 |
| Outcome | `clear` |

PR #3609, opened 2026-08-03
Deduplicate the local run scripts and fold the profiling build into `prod:preset`

tom2drum:

The local run scripts had grown by copy-paste. `tools/profiling/profile.preset.sh` was a copy of `tools/dev-server/prod.preset.sh` — the two were written in parallel and differed only in the build flags — and the three run scripts (`dev.preset.sh`, `dev.local.sh`, `prod.preset.sh`) each carried their own copy of the env-file layering, the asset regeneration block and the `dotenv` launch wrapper. Pure clean-up: no new capability beyond `--port` reaching `prod:preset`.

### Proposed Changes

**Profiling build folded into `prod:preset`** — the profiling copy is deleted and its one real difference is now a flag, so `pnpm prod:preset <alias> --profile` builds with `next build --webpack --profile`. There is no `profile:preset` package script anymore; `tools/profiling/CONTEXT.md` documents the invocation. Inheriting `prod.preset.sh`'s behaviour also fixes two things the profiling copy got wrong: `--skip-build` no longer re-fetches `.env.tmp` (it validates the previous run's output instead), and it derives `NEXT_PUBLIC_ICON_SPRITE_HASH` from the built sprite file rather than serving with it empty.

**Common steps extracted** — `tools/dev-server/run_steps.sh` holds `build_port_args`, `build_env_args`, `prepare_assets` and `run_with_envs`; each run script keeps only its argument parsing and the command it finally runs (288 lines across the three → 184, plus 118 shared). It is sourced rather than executed, because `prepare_assets` sources `build_sprite.sh` so the exported sprite hash reaches the launch step.

**`prod:preset` gains `--port`**, which the dev scripts already had. For it to mean anything on a `--skip-build` restart, `envs.js` generation moved from the build step to the start step — the browser fetches `envs.js` at runtime, so generating it at start is what lets a re-serve change ports.

**A naming convention, written down** — the repo already followed one implicitly: dotted names are entry points a human types (`dev.preset.sh`, `docker.preset.sh`), snake_case names are internals another script sources or invokes (`build_sprite.sh`, now `run_steps.sh`). `tools/dev-server/CONTEXT.md` states both rules and notes that `fetch.sh` predates them and is referenced by path from the Dockerfile.

`pnpm profile:preset` no longer exists — use `pnpm prod:preset <alias> --profile`. It's a local dev command only; nothing in CI, the Dockerfile, or the deploy path referenced it.

### Additional Information

No env variable changes. Verified by running each path against the `eth` and `multichain` presets: `dev:preset` on a custom port (`envs.js` picked up the port, sprite hash and pro-api flag present, HTTP 200), the multichain branch (config generated under `public/assets/multichain/`), a full `prod:preset` build + `next start`, and a `--skip-build` re-serve on a different port (`envs.js` regenerated with the new port, sprite hash derived from the built file, HTTP 200). Also `bash -n` on all four scripts, the sourced-only guard, every usage/validation path, and `pnpm lint:doc-links` + `pnpm lint:cspell`.

## Checklist for PR author


tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `269fae4155` → `4a96fd3e1a` |
| Round | 1 of 3 |
| Findings | 1 blocker · 1 major · 3 nit |
| By axis | Spec 0 · Standards 3 · Correctness 2 |
| Outcome | `blocked` |

tools/dev-server/prod.preset.sh
```diff
@@ -3,110 +3,99 @@
...
+        exit 1
+      fi
+      port="$2"; shift 2 ;;
+    --port=*)
+      port="${1#--port=}"; shift ;;
+    --skip-build)
+      skip_build=true; shift ;;
+    --profile)
+      profile=true; shift ;;
+    *)
```
tom2drum, line 41: **blocker** — `*) positional+=( "$1" ); shift ;;` treats a bare `--` as a positional. `pnpm prod:preset -- eth` forwards that `--` into the script (verified), so `${#positional[@]} -ne 1` prints usage and exits. The old last-wins `for arg` parser overwrote `--` with `eth` and succeeded.

Handle `--)` with `shift` (ignore) before collecting positionals — same for any other run script that adopted this parser.

— Reviewed by Composer

tools/dev-server/prod.preset.sh
```diff
@@ -3,110 +3,99 @@
...
-dotenv \
-  -v NEXT_PUBLIC_GIT_COMMIT_SHA=$(git rev-parse --short HEAD) \
-  -v NEXT_PUBLIC_GIT_TAG=$(git describe --tags --abbrev=0) \
-  -v NEXT_PUBLIC_ICON_SPRITE_HASH="${NEXT_PUBLIC_ICON_SPRITE_HASH}" \
-  "${env_args[@]}" \
-  -- bash -c 'next start -p $NEXT_PUBLIC_APP_PORT' |
+# generate envs.js and start the production server. Both steps belong here rather than in the
+# build: envs.js is read by the browser at runtime, so regenerating it now is what lets a
+# --skip-build run pick up a different --port, and the server itself reads the pro-api flag.
+run_with_envs 'source ./deploy/scripts/export_pro_api_flag.sh && ./deploy/scripts/make_envs_script.sh && next start -p $NEXT_PUBLIC_APP_PORT' |
```
tom2drum, line 96: **major** — `run_with_envs '... && ./deploy/scripts/make_envs_script.sh && next start ...' | pino-pretty` moved `make_envs_script.sh` out of the build's `|| exit 1` path into a pipeline with no `pipefail`. If `make_envs_script.sh` fails, the left-hand `bash -c` exits non-zero but the pipeline still exits 0 (pino-pretty's status).

Run make_envs via `run_with_envs ... || exit 1` before the piped `next start`, or `set -o pipefail` for this pipeline.

— Reviewed by Composer

.agents/AGENTS.md
```diff
@@ -37,7 +37,7 @@ Three ways to start the dev server:
 - `pnpm dev:local` — runs against a locally running Blockscout backend, using the committed `tools/dev-server/.env.localhost`.
 - `pnpm dev` — plain Next.js dev; needs a hand-written `.env.local` with the instance config.
 
-Both `dev:preset` and `dev:local` accept `--port <number>` to run on a port other than 3000 (it overrides `NEXT_PUBLIC_APP_PORT`, keeping the generated `envs.js` consistent).
+`dev:preset`, `dev:local` and `prod:preset` all accept `--port <number>` to run on a port other than 3000 (it overrides `NEXT_PUBLIC_APP_PORT`, keeping the generated `envs.js` consistent).
```
tom2drum: **nit** — Under "Three ways to start the **dev** server", the line now lists `prod:preset` among scripts that accept `--port`. `prod:preset` is not a fourth way to start the *dev* server; that sentence's invalidation trigger becomes "any run script's flags" rather than "dev-server start options" (prose-smells: Divergent Change / AGENTS.md carries pointers, never mechanics).

Keep `--port` on the two *dev* starters here; leave `prod:preset --port` to `tools/dev-server/CONTEXT.md` (already documents it).

— Reviewed by Composer

tools/dev-server/prod.preset.sh
```diff
@@ -3,110 +3,99 @@
...
 # Useful for performance measurements, where dev-mode overhead (React dev build, Turbopack
 # on-demand compile, StrictMode double-fetch) would skew the numbers.
+# The shared steps live in run_steps.sh.
 #
-# Usage: pnpm prod:preset <instance_alias> [--skip-build]
+# Usage: pnpm prod:preset <instance_alias> [--port <number>] [--skip-build] [--profile]
 #
 #   --skip-build  Start the server from the existing build (.next) without rebuilding.
 #                 Reuses the .env.tmp and public assets produced by the previous full run.
+#   --profile     Build a React-profileable production bundle (`next build --webpack --profile`):
```
tom2drum: **nit** — The new `--profile` header block restates the webpack/alias/minification rationale already owned by `tools/profiling/CONTEXT.md` "Running a profileable build" + Gotchas (prose-smells: Duplication / point, don't summarise).

Keep a one-line flag blurb; point at `tools/profiling/CONTEXT.md` for the rest.

— Reviewed by Composer

tools/dev-server/CONTEXT.md
```diff
@@ -17,15 +17,24 @@ unnecessary.
 
 ## Files
 
+Two naming rules hold across this directory (and `tools/scripts/`, `deploy/scripts/`):
```
tom2drum: **nit** — "Two naming rules hold across this directory (and `tools/scripts/`, `deploy/scripts/`)" puts a cross-tree convention in a directory CONTEXT. Renaming `deploy/scripts/build_sprite.sh` would force an edit here for a reason that is not "dev-server env fetch/run behaviour" (prose-smells: Divergent Change).

Scope the rule to `tools/dev-server/` only, or move the cross-cutting convention to a doc that owns all three trees.

— Reviewed by Composer

tools/dev-server/prod.preset.sh
tom2drum: Fixed in 125ffd265. Confirmed the claim first: `pnpm prod:preset -- eth --skip-build` did print usage, and pnpm does forward the separator. Added a `--) shift ;;` arm to all three run scripts — `dev.preset.sh` and `dev.local.sh` had the same wart pre-existing (their `*)` arms collected or rejected the `--`), so they get it too. `pnpm prod:preset -- eth --port abc` now reaches the port validator instead of failing the positional count, and `-- --skip-build` alone still errors on the missing alias.

tools/dev-server/prod.preset.sh
tom2drum: Fixed in 125ffd265 via `set -o pipefail` in `run_steps.sh`, with the reason in a comment there. Chose that over splitting `make_envs` into its own `|| exit 1` step because the split would source `export_pro_api_flag.sh` twice — two probes to `api.blockscout.com` on every prod start — and because pipefail also closes the identical hole in `dev.preset.sh` / `dev.local.sh`, whose `make_envs && next dev` pipelines have always masked failures.

Verified both directions (`bash -c "exit 7" | cat` → 7 with the lib sourced, 0 without) and re-ran `dev:preset` end to end, since `build_sprite.sh` is sourced into the same shell and pipes into `head -c 8`: sprite built, `envs.js` correct, HTTP 200.

.agents/AGENTS.md
tom2drum: Fixed in 125ffd265 — promoted from nit at the authors request. Reverted to naming only the two dev starters; `prod:preset --port` stays documented in `tools/dev-server/CONTEXT.md`, which is where the run-script flags belong.

tools/dev-server/prod.preset.sh
tom2drum: Fixed in 125ffd265 — promoted from nit. The header is now one line plus a pointer to `tools/profiling/CONTEXT.md`, which owns the rationale. Worth noting the duplication was created by this PR: the block was copied out of the deleted `profile.preset.sh`, and the same content had just been written into the profiling CONTEXT.

tools/dev-server/CONTEXT.md
tom2drum: Fixed in 125ffd265 — promoted from nit, with a narrowing rather than a removal. The rule is now scoped to this directory and the foreign filenames (`build_sprite.sh`, `download_assets.sh`, `docker.preset.sh`, `pw.docker.sh`) are gone — those were what created the coupling you describe. Both rules stay, because writing the convention down was the point of this part of the change and no doc currently owns all three trees.

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `269fae4155` → `125ffd2652` |
| Round | 2 of 3 |
| By axis | Spec — · Standards — · Correctness — (arbitration) |
| Outcome | `clear` |

Arbitration of Round 1: all five claimed fixes in `125ffd265` verified (`--)` in the three run scripts; `set -o pipefail` in `run_steps.sh`; AGENTS.md / `--profile` header / naming-rule scope). No fix-induced regressions.

— Reviewed by Composer

PR #3606, opened 2026-07-31
Add a code review step to the agent task workflow

tom2drum:
## Description and Related Issue(s)

Agent-written code went straight from "the agent says it's done" to a human reading the diff. This adds a review step in between, so an agent's work is critiqued before anyone looks at it — and so an unattended run can review, fix, and commit its own work without a developer in the loop for every leaf.

### Proposed Changes

Two new skills, deliberately split so the reviewer never fixes what it finds:

- **`review-changes`** — reviews a change on three axes (spec compliance, repo standards, correctness), each in its own fresh subagent context, then normalizes the results in one place: single severity per finding, one label per defect, and an anchor check that opens every cited `file:line` to drop hallucinated or stale claims. Output mode follows the tree state rather than a flag — an inline PR comment needs its line to exist in a pushed diff, so uncommitted work can only produce a `review.md` record. Supporting files carry the reusable parts: a 13-smell baseline adapted from Fowler for React/TS (`smells.md`), the record format (`review-template.md`), and the `gh` surface both skills share (`gh-commands.md`).
- **`resolve-review`** — adjudicates each finding and closes it out. Verdicts depend on who wrote the comment: this workflow's own findings can be rejected with a written reason, a bot's can be rejected outright, and a human's never are — they get `answered` instead, with the reasoning and the thread left open for the human to close.

Findings are `blocker` / `major` / `nit`, plus an orthogonal `needs-human` flag for anything turning on design intent the code can't settle. Disagreements go to arbitration rounds — a single fresh agent verifying fixes and ruling on rejections — bounded both by a 3-round cap and a no-repeat-rejection rule, so the loop terminates even against a stubborn re-rejector.

The rest wires it in:

- **`implement-task --auto`** — an unattended chain: review each leaf, resolve the findings, commit, take the next leaf. Typing the flag is what grants the run commit authority; it never amends or force-pushes, so unwinding a bad run is a plain `git reset`. It pushes only when finalizing the task's last leaf. Stop conditions are explicit (a `needs-human` finding, an unsettled dispute, a `[human]` subtask, a pending question, a red build), and every run closes with a catch-up summary that surfaces accepted rejections first — that's where an agent talks itself out of work while nobody is watching.
- **`[verify]` tag** — a third breakdown tag marking `[agent]` leaves whose result a human must verify in the running product. Such a leaf is left uncommitted even under `--auto`.
- **`code-reviewer` agent** (`.claude/agents/`) — a thin launcher so Claude Code can dispatch the review into a fresh context.
[...]

PR #3610, opened 2026-08-03
Tighten the agent doc-reference checker and fix the paths it was hiding

tom2drum:

Follow-up to #3606, which introduced `pnpm lint:doc-links` — the checker that resolves cross-references in the agent instruction surface, so a renamed file cannot silently leave an agent pointed at something that is no longer there.

That checker shipped with three deliberately broad exemptions, chosen to get it to zero false positives on its first run. This measures what each one was actually suppressing. All three turned out to be scoped far wider than the cases they existed for, and the widest of them was hiding real references.

### Proposed Changes

| Exemption | What it was actually buying | Now |
| --- | --- | --- |
| skip whole `*-template.md` files | nothing — suppressed zero findings | removed, +3 files checked |
| skip any line containing `e.g.` or `→` | 147 lines silenced to spare **2** genuine illustrations | scoped to the path, and only inside a table row for `→`; `e.g.` is no longer an exemption at all |
| ignore unanchored paths | 3 references naming a *kind* of file — and 14 naming a real file written short | shorthand is now reported; the 14 are written in full |

**11 references are newly checked** that the whole-line skip used to hide, among them `AGENTS.md`'s own pointers to `GLOSSARY.md`, `registry.json` and `tools/dev-server/CONTEXT.md`.

Shorthand detection is decidable rather than heuristic: a path matching exactly one tracked file names that file (`toolkit/theme/theme.ts` for `src/toolkit/theme/theme.ts`), while `types/api.ts` matching thirty of them asserts no single location worth checking. The split was clean across all 17 unanchored references. Files only — a directory such as `hooks/` is a convention every slice follows, and nothing distinguishes that from shorthand.

Three coverage gaps surfaced while measuring, all fixed here:

- The **10 `CONTEXT.md` files were never scanned**, though `prose-smells.md` and `.cursor/BUGBOT.md` both name them as instruction surface. They are clean as they stand.
- **Every rules file was checked twice**, through its `.cursor/rules/*.mdc` symlink, so each finding printed twice. Deduplicated by realpath: 57 to 49 unique files.
- **Anchored directory references** were skipped for want of a file extension.

Two documentation fixes, both needed to make the tightened rules pass:

- `.agents/rules/design-system.md` wrote all of its paths against an implicit `toolkit/` root that is really `src/toolkit/`, so not one of them resolved from anywhere. 22 paths anchored.
- `policies/app.ts` is now `src/server/csp/policies/app.ts` in the `add-env-var` skill.

One marker remains — a `<placeholder>` segment, or the output side of a `→` inside a table row. It is stated once, in a single string the **failure message** prints, so the convention reaches an author at the moment they trip it and there is no second copy to drift.

`e.g.` was dropped as a marker deliberately. Prose introducing a real file as an example is the common case by far, so exempting the form left live references unprotected against a later rename — the drift this script exists to catch. An example naming no real file gets a placeholder segment instead.

A path resolved against its own directory gets the same protection. Once broken, such a path is indistinguishable from one naming a *kind* of file, so a renamed neighbour used to fail silently; a path whose parent sits beside the file is now read as a reference to that neighbour, which is most of what a `CONTEXT.md` contains.

None for the app — this touches a lint script and the agent instruction files.

Worth knowing for in-flight branches: the checker now fails on things it used to let through. A path written short, a reference sharing a line with an `e.g.` or a prose `→`, a broken neighbour reference, or a broken reference inside a `CONTEXT.md` or a `*-template.md` will now fail `Checks` where it previously passed. A path that names no real file needs a `<placeholder>` segment.

### Additional Information

A green run proves nothing by itself, so each rule was verified to fire:

- A 20-case probe file — 10 references that must be reported, 10 that must stay silent. All 10 fired, all 10 stayed silent.
- The structural changes were proven by temporarily breaking a reference in a `CONTEXT.md`, in a `*-template.md`, and in a symlinked rules file. Each was reported, and the last exactly once.
- Every reference the removed exemptions had been hiding was broken in turn and confirmed to report, and the kind-of-file references (`types/api.ts`, `general/Page.tsx`) confirmed to stay silent when broken.
- Clean run green at 49 files; `eslint` and `cspell` clean.

Both probes were scratch files, deleted after use — there is no committed regression test for the script, which is the notable gap in this change. The checker is now complex enough (path-scoped exemptions, shorthand resolution, realpath deduplication) to deserve one.

Left alone deliberately: `.github` is not a root, since it holds no markdown at all — the `.github/*instructions*` glob in the review skill is forward-looking. Root-relative paths such as `/deploy/` stay unchecked.

## Checklist for PR author

### DIFF
--- a/.agents/skills/review-changes/prose-smells.md
+++ b/.agents/skills/review-changes/prose-smells.md
@@ -14,9 +14,9 @@ Three rules bind the baseline, mirroring `smells.md`:
   rule, suppress the smell that would contradict it.
 - **Always a judgement call.** Report a smell as a labelled possibility ("possible Sediment"), never as a
   violation.
-- **Skip what tooling enforces.** `lint:doc-links` already resolved every link, heading anchor and path —
-  including a path written short; cspell already ran. Their findings are facts, not smells — a reference that
-  does not resolve is never a finding here.
+- **Skip what tooling enforces.** `lint:doc-links` already resolved every link, heading anchor and path
+  reference, and flagged every path written short; cspell already ran. Their findings are facts, not smells —
+  neither a dead reference nor a shorthand path is ever a finding here.
 
 One defect, one label. Duplication and Shotgun Surgery describe the same mess from two directions — pick
 the sharper one.
--- a/.agents/skills/review-changes/prose-smells.md
+++ b/.agents/skills/review-changes/prose-smells.md
@@ -14,8 +14,9 @@ Three rules bind the baseline, mirroring `smells.md`:
   rule, suppress the smell that would contradict it.
 - **Always a judgement call.** Report a smell as a labelled possibility ("possible Sediment"), never as a
   violation.
-- **Skip what tooling enforces.** `lint:doc-links` already resolved every link, path reference and heading
-  anchor; cspell already ran. Their findings are facts, not smells — a broken link is never a finding here.
+- **Skip what tooling enforces.** `lint:doc-links` already resolved every link, heading anchor and path —
+  including a path written short; cspell already ran. Their findings are facts, not smells — a reference that
+  does not resolve is never a finding here.
 
 One defect, one label. Duplication and Shotgun Surgery describe the same mess from two directions — pick
 the sharper one.
--- a/tools/scripts/check-doc-links.mjs
+++ b/tools/scripts/check-doc-links.mjs
@@ -7,11 +7,15 @@
 // what a rule says, not on whether the rule's target still exists.
 //
 // These documents also carry paths that are *illustrations* rather than references — a template's output
-// column, an `e.g.`, a kind of file that lives in many slices. Checking those would produce noise that
-// trains everyone to ignore the checker, so an illustration is exempt wherever it cannot be mistaken for a
-// reference: a `<placeholder>` segment, an `(e.g. …)` parenthetical, or the right-hand side of a `→`. Each
-// exemption is scoped to the path itself rather than to its whole line, so a real reference standing beside
-// an illustration is still checked. Anything unmarked is read as a reference and has to resolve.
+// column, a kind of file that lives in many slices. Checking those would produce noise that trains everyone
+// to ignore the checker, so an illustration is exempt where it carries a mark that cannot be read as a
+// reference; `ILLUSTRATION_FORMS` below is the only statement of what those marks are. Each exemption is
+// scoped to the path itself rather than to its whole line, so a real reference standing beside an
+// illustration is still checked, and anything unmarked is read as a reference and has to resolve.
+//
+// An `e.g.` is deliberately *not* a mark. Prose that introduces a real file as an example is the common
+// case by far, and exempting it would leave those references unprotected against a later rename — the drift
+// this script exists to catch. An example that names no real file gets a `<placeholder>` segment instead.
 
 import { execFileSync } from 'node:child_process';
 import { existsSync } from 'node:fs';
@@ -88,11 +92,15 @@ function withoutFences(body) {
   });
 }
 
-// Blanks the marked-illustration regions and leaves the rest of the line checkable. A `→` runs to the end of
-// its table cell, because what follows the arrow is the hypothetical output of the pattern before it.
-const withoutIllustrations = (line) => line
-  .replace(/\([^)]*e\.g\.[^)]*\)?/g, '')
-  .replace(/→[^|]*/g, '');
+// Stated once, and printed on failure so the convention reaches an author at the moment they trip it rather
+// than in a document they would have to know to read.
+const ILLUSTRATION_FORMS = 'give it a <placeholder> segment, or place it after a → in a table row as the ' +
+  'output of the pattern before it';
+
+// Blanks the marked regions and leaves the rest of the line checkable. The arrow form is confined to table
+// rows, where a cell pairs a pattern with its filled-in output; in prose an arrow is ordinary punctuation
+// and the path after it is a reference like any other. The `<placeholder>` form is per-path, below.
+const withoutIllustrations = (line) => (/^\s*\|/.test(line) ? line.replace(/→[^|]*/g, '') : line);
 
 const isPlaceholder = (target) => /[<>{}*]|__/.test(target);
 
@@ -111,6 +119,15 @@ function resolves(realDir, target) {
 function shorthandFor(target) {
   const matches = tracked.filter((f) => f.endsWith(`/${ target }`));
   return matches.length === 1 ? matches[0] : undefined;
+}
+
+// A path whose parent directory sits beside the file was written relative to it — `components/Provider.tsx`
+// in a CONTEXT.md is a reference to a neighbour, and its absence is a break. Without this the whole relative
+// class goes unprotected: once broken, such a path is indistinguishable from `types/api.ts` naming a kind of
+// file, so a renamed neighbour would fail silently.
+function nearby(realDir, target) {
+  const parent = path.dirname(target);
+  return parent !== '.' && !target.endsWith('/') && existsSync(path.resolve(realDir, parent));
 }
 
 async function checkFile(fileRel, failures) {
@@ -136,7 +153,7 @@ async function checkFile(fileRel, failures) {
       const full = shorthandFor(target);
       if (full) {
         report(`${ target } is shorthand; write it in full: ${ full }`);
-      } else if (target.startsWith('./') || TOP_LEVEL.has(target.split('/')[0])) {
+      } else if (target.startsWith('./') || TOP_LEVEL.has(target.split('/')[0]) || nearby(realDir, target)) {
         report(`path reference does not exist: ${ target }`);
       }
     }
@@ -185,9 +202,7 @@ if (failures.length > 0) {
   // eslint-disable-next-line no-console
   console.error(
     `\n${ failures.length } unresolved reference(s) across ${ files.length } file(s).\n` +
-    'A path naming a shape rather than a file has to be marked as one, or it is read as a reference: give ' +
-    'it a <placeholder> segment, put it in an (e.g. …) parenthetical, or place it after a → as the output ' +
-    'of the pattern before it.',
+    `A path naming a shape rather than a file has to be marked as one, or it is read as a reference: ${ ILLUSTRATION_FORMS }.`,
   );
   process.exit(1);
 }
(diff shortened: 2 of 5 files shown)


tools/scripts/check-doc-links.mjs
```diff
@@ -86,33 +88,57 @@ function withoutFences(body) {
   });
 }
 
+// Blanks the marked-illustration regions and leaves the rest of the line checkable. A `→` runs to the end of
+// its table cell, because what follows the arrow is the hypothetical output of the pattern before it.
+const withoutIllustrations = (line) => line
+  .replace(/\([^)]*e\.g\.[^)]*\)?/g, '')
+  .replace(/→[^|]*/g, '');
```
tom2drum: **major** — `.replace(/→[^|]*/g, '')` blanks every path after `→`, including real references. `tools/dev-server/CONTEXT.md:12` — `(rewrite → \`src/pages/api/config.ts\`, …)` — drops the existing `src/pages/api/config.ts`. That contradicts L13–14 (“a real reference standing beside an illustration is still checked”) and the comment’s “table cell” framing: the strip is not limited to tables.

Scope `→` stripping to table cells (stop at `|` **and** require a table-row context), or only treat a path as illustration when it carries its own marker (`<placeholder>` / `(e.g. …)`).

— Reviewed by Composer

tools/scripts/check-doc-links.mjs
```diff
@@ -86,33 +88,57 @@ function withoutFences(body) {
   });
 }
 
+// Blanks the marked-illustration regions and leaves the rest of the line checkable. A `→` runs to the end of
+// its table cell, because what follows the arrow is the hypothetical output of the pattern before it.
+const withoutIllustrations = (line) => line
+  .replace(/\([^)]*e\.g\.[^)]*\)?/g, '')
```
tom2drum: **major** — `\([^)]*e\.g\.[^)]*\)?` makes the closing `)` optional, so an unclosed `(e.g.` wipes the rest of the line and any real backtick paths after it are never checked. The claimed scope is a closed `(e.g. …)` parenthetical.

Require the closing `)` (`\)` not `\)?`).

— Reviewed by Composer

.agents/skills/review-changes/prose-smells.md
```diff
@@ -14,8 +14,9 @@ Three rules bind the baseline, mirroring `smells.md`:
   rule, suppress the smell that would contradict it.
 - **Always a judgement call.** Report a smell as a labelled possibility ("possible Sediment"), never as a
   violation.
-- **Skip what tooling enforces.** `lint:doc-links` already resolved every link, path reference and heading
-  anchor; cspell already ran. Their findings are facts, not smells — a broken link is never a finding here.
+- **Skip what tooling enforces.** `lint:doc-links` already resolved every link, heading anchor and path —
```
tom2drum: **major** — possible Incomplete case split — “already resolved every … path — including a path written short” then narrows to “a reference that does not resolve is never a finding here”. The checker reports shorthand as “is shorthand; write it in full”, not as non-resolving; the third clause does not cover the case the second clause just added.

Drop the third clause, or name shorthand next to non-resolving refs as tooling facts Standards must not re-report.

— Reviewed by Composer

tools/scripts/check-doc-links.mjs
```diff
@@ -125,29 +151,44 @@ async function checkFile(fileRel, failures) {
...
+  for (const { file, line, message } of failures) {
     // eslint-disable-next-line no-console
-    console.error(`${ file }:${ line } — ${ reason }: ${ target }`);
+    console.error(`${ file }:${ line } — ${ message }`);
   }
   // eslint-disable-next-line no-console
-  console.error(`\n${ failures.length } unresolved reference(s) across ${ files.length } file(s).`);
+  console.error(
+    `\n${ failures.length } unresolved reference(s) across ${ files.length } file(s).\n` +
+    'A path naming a shape rather than a file has to be marked as one, or it is read as a reference: give ' +
```
tom2drum: **nit** — possible Duplication — the exemption triad (`<placeholder>`, `(e.g. …)`, `→`) is restated from the file header (L11–14) into the failure footer. Two copies in one file will drift.

One shared string for the triad; keep the header as why, reuse it in the footer.

— Reviewed by Composer

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `269fae415` → `d2e448517` |
| Round | 1 of 3 |
| Findings | 0 blocker · 3 major · 1 nit |
| By axis | Spec 0 · Standards 2 · Correctness 2 |
| Outcome | `blocked` |

One defect shared by the header claim and `withoutIllustrations` is reported once under Correctness.

tools/scripts/check-doc-links.mjs
tom2drum: **fix** — confirmed, and understated. Two live references were being blanked, not one:

| Line | Path | Resolved before the strip |
| --- | --- | --- |
| `tools/dev-server/CONTEXT.md:12` | `src/pages/api/config.ts` | yes |
| `src/features/connect-wallet/CONTEXT.md:29` | `components/Web3ProviderInner.tsx` | yes |

Took the first option: the strip is now gated on `/^\s*\|/`, so it applies inside a table row and nowhere else. That keeps the single genuine case — the page-template mapping in `add-new-page/SKILL.md:72` — and makes the comment true instead of aspirational, which was the actual defect.

Worth recording why this got through: the measurement behind the original claim ("exactly one line has an anchored path after a `→`") was run over `.agents`/`.claude`/`.cursor` only, *before* the CONTEXT.md files were added to the scanned set. Both instances live in the files that measurement never covered.

Restoring the checking turned out to be only half the fix — see the follow-up below.

Fixed in 4e47efd6c.

tom2drum: Follow-up on this thread, because the fix above was incomplete in a way worth flagging.

After gating the strip on table rows, I broke both references to confirm they were caught. Only one reported. `components/Web3ProviderInner.tsx` resolves *relative to its own directory* — and once broken it is indistinguishable from `types/api.ts` naming a kind of file, which the unanchored rule deliberately keeps silent. So the arrow fix restored checking, but that reference still had no drift protection.

A path whose parent directory sits beside the file is now read as a reference to that neighbour. That covers the whole relative class, which is most of what a `CONTEXT.md` contains. Verified: both arrow references now report when broken, and `types/api.ts` and `general/Page.tsx` stay silent when broken.

Fixed in 4e47efd6c.

tools/scripts/check-doc-links.mjs
tom2drum: **fix** — real, though latent rather than active: I audited all 15 lines carrying an unclosed `(e.g.` and none has a path after it on the same line, so nothing was being silenced today. The trap was genuine all the same, and the header did claim a closed parenthetical.

Fixed by removing the construct rather than requiring the `)`. The `(e.g. …)` exemption is gone entirely — prose introducing a *real* file as an example is the common case by far, so the form was costing drift protection on three live references (`tools/dev-server/` in `architecture.md`, two `config.ts` paths in `env-vars.md`) to spare one hypothetical path. The regex whose optional `)` you flagged no longer exists, so the failure mode cannot recur.

For the record, requiring `\)` would also have been safe — I checked the nested-paren case in `check-github-cli/SKILL.md:10`, where `[^)]*` stops at the markdown link's closing paren and the match still terminates correctly. Removal was simply the better trade.

Fixed in 4e47efd6c.

.agents/skills/review-changes/prose-smells.md
tom2drum: **fix**, on narrower grounds than stated. The mechanism in the finding is slightly off: a path written short *does* fail to resolve — that is exactly how the checker detects it, `resolves()` returning nothing before `shorthandFor()` runs — so the third clause did cover the case.

The wording was the real problem. "A reference that does not resolve" reads as *dangling*, which leaves a reviewer to guess whether "this path should be written in full" is a tooling fact or a Standards finding. Now stated as both cases explicitly:

> `lint:doc-links` already resolved every link, heading anchor and path reference, and flagged every path written short; cspell already ran. Their findings are facts, not smells — neither a dead reference nor a shorthand path is ever a finding here.

Fixed in 4e47efd6c.

tools/scripts/check-doc-links.mjs
tom2drum: **fix** — promoted from nit by the developer at Gate 1.

The triad is now a pair, since `e.g.` was dropped as a marker (see the thread on line 94), and it lives in one `ILLUSTRATION_FORMS` string that the failure message interpolates. The header comment keeps the *why* — including why an `e.g.` is deliberately not a marker — and points at the constant by name rather than restating what it contains, so a third form means editing one place.

Fixed in 4e47efd6c.

tools/scripts/check-doc-links.mjs
tom2drum: **R2 arbitrator:** `fixed` — `withoutIllustrations` only strips `→` on table rows; prose probe keeps `src/pages/api/config.ts`. `nearby()` covers the relative-neighbour gap the restore exposed. No regression.

— Reviewed by Composer

tools/scripts/check-doc-links.mjs
tom2drum: **R2 arbitrator:** `fixed` — the `(e.g. …)` exemption and its regex are gone, so the optional-`)` wipe cannot recur. Acceptable stronger fix than requiring `).

— Reviewed by Composer

.agents/skills/review-changes/prose-smells.md
tom2drum: **R2 arbitrator:** `fixed` — wording now names both cases: dead references and shorthand paths are tooling facts, not Standards smells.

— Reviewed by Composer

tools/scripts/check-doc-links.mjs
tom2drum: **R2 arbitrator:** `fixed` — single `ILLUSTRATION_FORMS` constant; header points at it; failure footer interpolates it.

— Reviewed by Composer

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `269fae415` → `4e47efd6c` |
| Round | 2 of 3 |
| By axis | Spec 0 · Standards 0 · Correctness 0 |
| Outcome | `clear` |

## Arbitration

| Finding | Ruling |
| --- | --- |
| F1 · → strip blanked prose refs | `fixed` — strip gated on table rows; prose paths kept; `nearby()` covers relative neighbours |
| F2 · optional `)` on `(e.g.` wipe | `fixed` — `(e.g. …)` exemption removed; failure mode gone |
| F3 · prose-smells shorthand case split | `fixed` — dead refs and shorthand both named as tooling facts |
| F4 · duplicated exemption triad | `fixed` — single `ILLUSTRATION_FORMS`; footer interpolates it |

PR #3625, opened 2026-08-10
Reduce noisy JavaScript errors reported to Rollbar

tom2drum:
## Description

Reduces noise in our Rollbar JavaScript error monitoring, where a large share of items came from third-party scripts, browser extensions, and abandoned wallet flows that never reach real users, drowning out real signal.

- **Capture only own-bundle uncaught errors.** The early window `error` listener forwarded *every* uncaught error to Rollbar, so a public page's browser extensions, injected wallet SDKs, userscript inline scripts, CDN assets, and opaque cross-origin `Script error.` events all became noise. Rollbar has no positive "only my code" allowlist, so we gate at the listener: forward an uncaught error only when the browser attributes it to a script under `/_next/`. Genuine page crashes are unaffected — they report through `AppErrorBoundary` as explicit `rollbar.critical` calls, not this listener.
- **Single-source the ignore rules across both Rollbar instances.** The error-page instance (`_error.tsx`) reused none of the browser instance's filtering. It now shares the window-free exception-class check and reports the page-crash error as a throwable so that check can read its exception class. It captures server-side uncaught exceptions — which carry none of the third-party `window` noise the browser instance fights — while leaving unhandled-rejection capture off.
- **Filter auto-translate DOM noise.** Browser auto-translate / DOM-mutating extensions relocate nodes React owns, so React's commit-phase `removeChild`/`insertBefore` fails with a `NotFoundError` (3000+ occurrences). Matched generically via the shared DOM-exception message tail.
- **Prune rules the origin gate made redundant** (injected-script / extension-origin / `Script error` / `ResizeObserver loop`), a net −165 lines.

Also includes two genuine fixes surfaced during triage: guarding the EIP-6963 `announceProvider` handler against a null `detail` (plus a capture-phase guard so wagmi's bundled `mipd` store is covered too), and reporting code-editor render-boundary crashes explicitly instead of relying on the silenced Monaco CDN noise.

## Environment variables

None.

## Minimum API version

None.

None. Note this changes *what* client errors are captured: after deploy, several currently-active items dominated by third-party scripts (notably `Script error.`) should stop arriving.

## Additional information

Filtering here suppresses at ingestion. Expected WalletConnect/AppKit connect-flow expiries (proposal / pairing / session request) are dropped via `ignoredMessages` code rules — they recur and match generically. Separately, a handful of low-count, vendored/environmental crash items (wagmi/AppKit, helia, phoenix) were muted directly in the Rollbar dashboard rather than given code rules, since a dependency bump may retire them.

### DIFF
--- a/src/features/connect-wallet/utils/install-eip6963-announce-guard.ts
+++ b/src/features/connect-wallet/utils/install-eip6963-announce-guard.ts
@@ -18,7 +18,7 @@ let isInstalled = false;
  * `stopImmediatePropagation()` the malformed event, neutralising it for every downstream consumer at
  * once. Idempotent.
  */
-export default function installEip6963AnnounceGuard(): void {
+export function installEip6963AnnounceGuard(): void {
   if (isInstalled || typeof window === 'undefined') {
     return;
   }
--- a/src/features/connect-wallet/utils/installEip6963AnnounceGuard.ts
+++ b/src/features/connect-wallet/utils/installEip6963AnnounceGuard.ts
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: LicenseRef-Blockscout
+
+// https://eips.ethereum.org/EIPS/eip-6963
+
+let isInstalled = false;
+
+/**
+ * Drops malformed `eip6963:announceProvider` events before any consumer's listener runs.
+ *
+ * EIP-6963 requires the event's `detail` to carry a provider `info`, but a non-compliant wallet
+ * extension can dispatch one with a null or partial `detail`. Consumers dereference `detail.info`
+ * without guarding — wagmi's vendored `mipd` store (`providerDetail.info.uuid`) most of all — and
+ * throw an uncaught "Cannot read properties of null (reading 'info')" we cannot fix inside the
+ * dependency.
+ *
+ * Listeners on `window` for a window-targeted event fire in registration order, so installing this
+ * before wagmi's config is created (which spins up the `mipd` store) lets it run first and
+ * `stopImmediatePropagation()` the malformed event, neutralising it for every downstream consumer at
+ * once. Idempotent.
+ */
+export default function installEip6963AnnounceGuard(): void {
+  if (isInstalled || typeof window === 'undefined') {
+    return;
+  }
+  isInstalled = true;
+
+  window.addEventListener('eip6963:announceProvider', (event) => {
+    const detail = (event as CustomEvent).detail;
+    if (detail === null || typeof detail !== 'object' || !('info' in detail) || !detail.info) {
+      event.stopImmediatePropagation();
+    }
+  }, true);
+}
--- a/src/services/rollbar/serverConfig.ts
+++ b/src/services/rollbar/serverConfig.ts
@@ -7,10 +7,11 @@ import config from 'src/config';
 import { isIgnoredExceptionClass } from './utils';
 
 /**
- * Rollbar options for the error-page instance in `src/pages/_error.tsx`. That instance runs on the
- * server (and on client-side error navigations) and reports only the explicit error handed to
- * `getInitialProps`; global capture stays off so it never re-reports, unfiltered, what the browser
- * instance already drops.
+ * Rollbar options for the error-page instance in `src/pages/_error.tsx`. It reports the explicit error
+ * handed to `getInitialProps`, plus server-side uncaught exceptions — `captureUncaught` is on because
+ * those page-crash errors surface nowhere else, and the server has none of the third-party `window`
+ * noise the browser instance has to filter. `captureUnhandledRejections` stays off, and the shared
+ * `isIgnoredExceptionClass` check drops the same DOM / Abort classes the browser instance drops.
  */
 export function buildServerConfig(accessToken: string): Configuration {
   return {
@@ -21,7 +22,7 @@ export function buildServerConfig(accessToken: string): Configuration {
       app_instance: config.services.rollbar.instance,
     },
     checkIgnore: (_isUncaught, _args, item) => isIgnoredExceptionClass(item),
-    captureUncaught: false,
+    captureUncaught: true,
     captureUnhandledRejections: false,
   };
 }
--- a/src/services/rollbar/serverConfig.ts
+++ b/src/services/rollbar/serverConfig.ts
@@ -8,10 +8,13 @@ import { isIgnoredExceptionClass } from './utils';
 
 /**
  * Rollbar options for the error-page instance in `src/pages/_error.tsx`. It reports the explicit error
- * handed to `getInitialProps`, plus server-side uncaught exceptions — `captureUncaught` is on because
- * those page-crash errors surface nowhere else, and the server has none of the third-party `window`
- * noise the browser instance has to filter. `captureUnhandledRejections` stays off, and the shared
- * `isIgnoredExceptionClass` check drops the same DOM / Abort classes the browser instance drops.
+ * handed to `getInitialProps`, plus server-side uncaught exceptions — `captureUncaught` is on there
+ * because those page-crash errors surface nowhere else, and the server has none of the third-party
+ * `window` noise the browser instance has to filter. It is gated to the server: this module also loads
+ * in the browser on a client-side error-page navigation, where an uncaught handler would bypass the
+ * queue's own-bundle origin gate and recapture the very noise we drop. `captureUnhandledRejections`
+ * stays off, and the shared `isIgnoredExceptionClass` check drops the same DOM / Abort classes the
+ * browser instance drops.
  */
 export function buildServerConfig(accessToken: string): Configuration {
   return {
@@ -22,7 +25,7 @@ export function buildServerConfig(accessToken: string): Configuration {
       app_instance: config.services.rollbar.instance,
     },
     checkIgnore: (_isUncaught, _args, item) => isIgnoredExceptionClass(item),
-    captureUncaught: true,
+    captureUncaught: typeof window === 'undefined',
     captureUnhandledRejections: false,
   };
 }
--- a/src/services/rollbar/serverConfig.ts
+++ b/src/services/rollbar/serverConfig.ts
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: LicenseRef-Blockscout
+
+import type { Configuration } from 'rollbar';
+
+import config from 'src/config';
+
+import { isIgnoredExceptionClass } from './utils';
+
+/**
+ * Rollbar options for the error-page instance in `src/pages/_error.tsx`. That instance runs on the
+ * server (and on client-side error navigations) and reports only the explicit error handed to
+ * `getInitialProps`; global capture stays off so it never re-reports, unfiltered, what the browser
+ * instance already drops.
+ */
+export function buildServerConfig(accessToken: string): Configuration {
+  return {
+    accessToken,
+    environment: config.services.rollbar.environment,
+    payload: {
+      code_version: config.services.rollbar.codeVersion,
+      app_instance: config.services.rollbar.instance,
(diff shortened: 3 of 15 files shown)


src/shared/code-editor/CodeEditor.tsx
```diff
@@ -285,6 +287,13 @@ const CodeEditor = ({ data, remappings, libraries, language, mainFile, contractN
     return <Center bgColor={ themeColors['editor.background'] } w="100%" h="100%" borderRadius="md">Oops! Something went wrong!</Center>;
   }, [ themeColors ]);
 
+  // Reported as a message rather than the raw error so it stays distinct from the CDN-load / worker
+  // noise that `isMonacoCdnError` drops: this fires only when the editor subtree actually crashed in
+  // render and the user saw the fallback — the signal worth keeping. The error rides along as data.
+  const handleEditorError = React.useCallback((error: Error) => {
+    rollbar?.error('Code editor failed to render', error);
```
tom2drum: **blocker** — `rollbar?.error('Code editor failed to render', error)` attaches the Error as a throwable, so Rollbar builds `body.trace` with frames. `isMonacoCdnError`'s frame check then drops the report whenever the stack touches the Monaco CDN — defeating the PR goal of reporting editor render crashes explicitly. The comment's "rides along as data" claim is wrong for Rollbar's API.

Match `AppErrorBoundary`: pass a plain object (`{ message, stack }`), not the Error instance.

— Reviewed by Cursor Grok 4.5

src/services/rollbar/serverConfig.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export function buildServerConfig(accessToken: string): Configuration {
+  return {
+    accessToken,
+    environment: config.services.rollbar.environment,
+    payload: {
+      code_version: config.services.rollbar.codeVersion,
+      app_instance: config.services.rollbar.instance,
+    },
+    checkIgnore: (_isUncaught, _args, item) => isIgnoredExceptionClass(item),
```
tom2drum, line 27: **major** — `checkIgnore: … isIgnoredExceptionClass(item)` never fires for how `_error.tsx` actually reports: `rollbar.error(err?.message ?? 'Unknown error', { cause, stack })` produces a message payload with no `body.trace.exception.class`. The PR's "shared window-free exception-class check" is therefore a no-op on the error-page instance (and there is no `ignoredMessages` for `'is not a child of this node'` either).

Pass `err` as the throwable, or share the DOM/Abort message ignores on this config too.

— Reviewed by Cursor Grok 4.5

src/features/connect-wallet/utils/installEip6963AnnounceGuard.ts
```diff
@@ -0,0 +1,33 @@
...
+ * without guarding — wagmi's vendored `mipd` store (`providerDetail.info.uuid`) most of all — and
+ * throw an uncaught "Cannot read properties of null (reading 'info')" we cannot fix inside the
+ * dependency.
+ *
+ * Listeners on `window` for a window-targeted event fire in registration order, so installing this
+ * before wagmi's config is created (which spins up the `mipd` store) lets it run first and
+ * `stopImmediatePropagation()` the malformed event, neutralising it for every downstream consumer at
+ * once. Idempotent.
+ */
+export default function installEip6963AnnounceGuard(): void {
```
tom2drum: **major** — `export default function installEip6963AnnounceGuard()` — utils must be named exports (`.agents/rules/typescript.md` → Default exports: "Everything else is a named export — hooks, utils, types, constants."). Peers in this folder already follow that.

Use `export function installEip6963AnnounceGuard` and named imports.

— Reviewed by Cursor Grok 4.5

src/features/connect-wallet/utils/install-eip6963-announce-guard.ts
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: LicenseRef-Blockscout
```
tom2drum, line 1: **major** — New util file is camelCase (`installEip6963AnnounceGuard.ts`). Breaks `.agents/rules/architecture.md` → Naming conventions: "Helper / util modules | kebab-case.ts" (peers: `wagmi-config.ts`, `public-client.ts`).

Rename to `install-eip6963-announce-guard.ts` (and the co-located `.spec.ts`).

— Reviewed by Cursor Grok 4.5

src/services/rollbar/utils.ts
```diff
@@ -68,8 +68,52 @@ export function getExceptionClass(item: Dictionary) {
...
+ * scripts, CDN/adblock load failures — are environmental and unactionable by us. Render-time failures
+ * still reach users through the editor's own ErrorBoundary, so dropping them here loses no signal.
+ *
+ * Still needed after the queue's own-bundle origin gate: a trace whose origin frame is our `/_next/`
+ * code but which only reaches the CDN bundle deeper in the stack passes that gate, and this catches it.
+ *
+ * Checks the message (worker failures arrive message-only, with no stack) and every frame (some
+ * traces enter through our own code and only reach the CDN bundle deeper in the stack).
+ */
+export function isMonacoCdnError(item: Dictionary) {
```
tom2drum: **nit** — `export function isMonacoCdnError(item: Dictionary)` has no return type. `.agents/rules/typescript.md` → Return types: top-level module functions declare return types (sibling `isIgnoredExceptionClass` already has `: boolean`).

Annotate `: boolean`.

— Reviewed by Cursor Grok 4.5

src/services/rollbar/utils.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import { describe, expect, it } from 'vitest';
+
+import { isMonacoCdnError, isIgnoredExceptionClass } from './utils';
+
+describe('isIgnoredExceptionClass', () => {
+  it('matches an AbortError so both Rollbar instances drop it', () => {
+    const item = {
+      body: { trace: { exception: { 'class': 'AbortError', message: 'signal is aborted without reason' } } },
+    } as unknown as Dictionary;
```
tom2drum: **nit** — Repeated `as unknown as Dictionary` with no safety comment. `.agents/rules/typescript.md` → Type assertions: "If a cast is unavoidable, add a comment explaining why it is safe."

One file-level note that partial Rollbar fixtures are force-typed as `Dictionary`, or a tiny fixture helper.

— Reviewed by Cursor Grok 4.5

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `d082c2cc` → `2af19527` |
| Round | 1 of 3 |
| Findings | 1 blocker · 3 major · 3 nit |
| By axis | Spec 1 · Standards 4 · Correctness 2 |
| Outcome | `blocked` |

## Not anchorable

- **nit · Spec** — PR Additional information says AppKit/wagmi items were muted in the Rollbar dashboard rather than given code rules, but `clientConfig` still adds `'Proposal expired' | 'Pairing expired' | 'Session request expired'`. Update the PR text to own these, or drop the code rules.
- **pre-existing (out of scope)** — `isIgnoredExceptionClass` drops every `NotFoundError` / `AbortError` by class alone (same breadth as before this PR). Narrowing to the DOM-exception message is a possible follow-up, not introduced here.

src/shared/code-editor/CodeEditor.tsx
tom2drum: Fixed in 4d9113c63 — `handleEditorError` now reports `{ cause, stack }` as plain data instead of the `Error` instance, so Rollbar builds no trace and `isMonacoCdnError`'s frame check can't drop it (matching `AppErrorBoundary`). Comment corrected.

src/services/rollbar/serverConfig.ts
tom2drum: Fixed in 4d9113c63 — `_error.tsx` now passes the `Error` as a throwable, so Rollbar builds a trace carrying `body.trace.exception.class` for the shared `isIgnoredExceptionClass` check to read. Server-side `captureUncaught` was also turned on, since page-crash errors weren't otherwise reported and the server has none of the third-party `window` noise the browser instance filters.

src/features/connect-wallet/utils/installEip6963AnnounceGuard.ts
tom2drum: Fixed in 4d9113c63 — now a named export (`export function installEip6963AnnounceGuard`); import site and spec updated.

src/features/connect-wallet/utils/install-eip6963-announce-guard.ts
tom2drum: Fixed in 4d9113c63 — renamed to `install-eip6963-announce-guard.ts` (and its `.spec.ts`); both import sites updated.

src/services/rollbar/utils.ts
tom2drum: Fixed in 4d9113c63 — annotated `: boolean`.

src/services/rollbar/utils.spec.ts
tom2drum: Fixed in 4d9113c63 — removed the `as unknown as Dictionary` coercions entirely: a plain object literal satisfies `Dictionary` (`{ [key: string]: unknown }`), so the double-cast and its now-unused import are gone.

src/services/rollbar/serverConfig.ts
tom2drum: Follow-up in e77233567 — `captureUncaught` is gated to `typeof window === 'undefined'`. This module also loads in the browser on a client-side error-page navigation; an unconditional uncaught handler there would bypass the queue's own-bundle origin gate and recapture the noise this PR drops, so capture is server-only.

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `2af19527` → `e77233567` (arbitration of Round 1) |
| Round | 2 of 3 |
| By axis | Spec 1 · Standards 0 · Correctness 0 |
| Outcome | `clear` |

## Arbitration

| Finding | Ruling |
| --- | --- |
| F1 blocker · CodeEditor Error-as-throwable | `fixed-accepted` — plain `{ cause, stack }` data; `isMonacoCdnError` cannot drop it |
| F2 major · server checkIgnore / `_error` shape | `fixed-accepted` — `rollbar.error(err ?? …)` builds `exception.class`; `captureUncaught` server-gated as documented |
| F3 major · default export | `fixed-accepted` — named export |
| F4 major · camelCase filename | `fixed-accepted` — `install-eip6963-announce-guard.ts` |
| F5 nit · `isMonacoCdnError` return type | `fixed-accepted` — `: boolean` |
| F6 nit · `as unknown as Dictionary` | `fixed-accepted` — casts removed |

No fix-introduced regressions. The server `captureUncaught: typeof window === 'undefined'` expansion is intentional (JSDoc why-comment) and out of bounds to re-litigate.

## Not anchorable

- **nit · Spec** (still open from R1) — PR Additional information still says AppKit/wagmi items were muted in the dashboard rather than given code rules, while `clientConfig` keeps `'Proposal expired' | 'Pairing expired' | 'Session request expired'`. Update the PR text to own these, or drop the code rules.

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `2af19527` → `e77233567` (arbitration correction) |
| Round | 2 of 3 |
| By axis | Spec 0 · Standards 0 · Correctness 0 |
| Outcome | `clear` |

Correction to the prior Round 2 body: the remaining Spec nit is **closed**. The PR description now owns the WalletConnect expiry `ignoredMessages` ("Expected WalletConnect/AppKit connect-flow expiries… are dropped via `ignoredMessages` code rules") and separately describes dashboard mutes for low-count items — so that R1 not-anchorable finding no longer holds.

All six Round 1 findings remain `fixed-accepted`. No open findings.

PR #3629, opened 2026-08-11
Restructure the product task workflow around vertical subtasks

tom2drum:
## Description

Refines the spec-driven product task workflow in `.agents/`. Documentation only — no application code is touched.

**A task that fits in one subtask no longer gets a spec.** It is grilled, implemented, and opened as a PR inside a single session, with the reasoning going into the PR description. A spec exists to hand work to a session that wasn't in the room; a task that never leaves the room doesn't need one. The `small / medium / large` vocabulary goes with it — a deferred subtask is marked by having only a `brief.md`, which is exactly what `large` used to record, and nothing mechanical ever read the label.

**A subtask is now a vertical slice**, bounded to one fresh context window and one commit, while the leaves inside it stay layer-shaped — one project skill each. The two levels answer different questions: what an agent can execute without judgement (a leaf) versus what a human can judge (a subtask). Recorded as [ADR 0002](.agents/adr/0002-layer-shaped-subtask-leaves.md), since it deviates from the tracer-bullet norm and looks wrong without its context.

That regrouping is what let the rest fall out:

- **Review and commit move from the leaf to the subtask.** Reviewing every leaf spent a subagent per axis on each step, and much of what it caught was churn the next leaf rewrote. A subtask's whole uncommitted diff is now one review and one commit.
- **Acceptance criteria replace the `[verify]` tag.** Each subtask spec carries a checklist; a `(human)` criterion is what pauses an unattended chain. The list is the gate, so no separate tag can drift from it — and the review's spec axis finally has a concrete target.
- **`Blocked by:` edges replace implicit ordering.** Numbers become identity rather than sequence, which is what lets a deferred subtask expand into appended siblings with no renumbering.
- **Nesting and the sub-branch / sub-PR scheme are gone.** They existed because a "big subtask" spanned several commits; with one commit per subtask that category no longer exists.

Skill-level changes:

- `to-spec` becomes a pure writer. Slack harvesting is dropped — reading replies is something the developer does in-session — and outreach moves to `grill-the-task`, where the questions are born and where a spec-less task needs it too.
- `grill-the-task` loses its size-classification step and gains a quiz loop over the breakdown (granularity, context-window fit, edges, prefactor-first), then forks: implement in-session, or hand off to `to-spec`.
- `review-changes` gates its spec axis on a spec existing — with none it runs two axes and says so, rather than inventing a standard to judge against.
- `create-pr` now always carries the *why*, sourced from the conversation, the issue, or the diff.

## Environment variables

None.

## Minimum API version

None.

None for any deployment. Within the workflow, specs written under the old shape (`Size` header row, `[verify]` tags, leaf-level breakdowns) are left untouched — a finished spec is a frozen record of what was decided then, and only new tasks follow the new structure.

## Additional information

`pnpm lint:doc-links` and `pnpm lint:cspell` both pass. The new `subtask-template.md` needs no symlink: `.claude/skills` is a directory symlink, so it is picked up automatically.

### DIFF
--- a/.agents/AGENTS.md
+++ b/.agents/AGENTS.md
@@ -68,6 +68,7 @@ Decisions with repo-wide consequences are recorded in `.agents/adr/`, named
 evidence and the trade-off, so it answers "why is it like this?" without a git archaeology session.
 
 - `0001-webpack-for-production-builds.md` — why production bundles are built with webpack while dev stays on Turbopack.
+- `0002-layer-shaped-subtask-leaves.md` — why a product task's subtasks cut vertically while the leaves inside them run along layers.
 
 Add a new record (next free number, and a line here) whenever a decision is expensive to rediscover:
 it constrains future work, was reached by measurement or an investigation worth not repeating, or
@@ -77,8 +78,9 @@ looks wrong without its context. Supersede rather than rewrite — flip the old 
 ## Product task workflow
 
 Product tasks (GitHub issues) are worked through a spec-driven workflow — interview, spec, agent
-implementation, code review. Specs accumulate in `.agents/tasks/` as a permanent record. See
-`./tasks/README.md` for the lifecycle, the skills that run it, and the spec conventions.
+implementation, code review. A task small enough to finish in its grilling session skips the spec
+entirely. Specs accumulate in `.agents/tasks/` as a permanent record. See `./tasks/README.md` for the
+lifecycle, the skills that run it, and the spec conventions.
 
 ## Editing this instruction set
 
--- a/.agents/AGENTS.md
+++ b/.agents/AGENTS.md
@@ -78,9 +78,8 @@ looks wrong without its context. Supersede rather than rewrite — flip the old 
 ## Product task workflow
 
 Product tasks (GitHub issues) are worked through a spec-driven workflow — interview, spec, agent
-implementation, code review. A task small enough to finish in its grilling session skips the spec
-entirely. Specs accumulate in `.agents/tasks/` as a permanent record. See `./tasks/README.md` for the
-lifecycle, the skills that run it, and the spec conventions.
+implementation, code review. Specs accumulate in `.agents/tasks/` as a permanent record. See
+`./tasks/README.md` for the lifecycle, the skills that run it, and the spec conventions.
 
 ## Editing this instruction set
 
--- a/.agents/adr/0002-layer-shaped-subtask-leaves.md
+++ b/.agents/adr/0002-layer-shaped-subtask-leaves.md
@@ -1,0 +1,57 @@
+# 0002 — subtasks cut vertically, leaves run along layers
+
+| | |
+| --- | --- |
+| Status | accepted |
+| Date | 2026-08-11 |
+| Deciders | @tom2drum |
+| Supersedes | — |
+
+## Decision
+
+**A subtask is a vertical slice; the leaves inside it are layer-shaped.**
+
+A subtask cuts a narrow but complete path through every layer it touches and is verifiable on its own. Its
+leaves do the opposite — each is one project skill's worth of work in one layer: `add-api-resource`, then
+`add-new-page`, then the `[human]` styling.
+
+```
+subtasks/01-cross-chain-list/     ← vertical: demoable, one context window, one commit
+  leaf 1 [agent] add-api-resource — declare the resource
+  leaf 2 [agent] add-new-page    — tab route + scaffold
+  leaf 3 [agent] wire the resource into the table
+  leaf 4 [human] style to mockup
+```
+
+The full model lives in "The subtask model" in `.agents/tasks/README.md`; this record holds only the
+reasoning, which that file should not have to carry.
+
+## Why
+
+The tracer-bullet norm says every unit of work should be a vertical slice, all the way down. Ours stops one
+level short, deliberately.
+
+**Layer-shaped leaves are what make execution mechanical.** The project skills are layer-shaped by
+construction — `add-api-resource` declares a resource, `add-new-page` scaffolds a route. A leaf that mapped
+one skill to one step lets `implement-task` execute it without deciding anything: open the skill, read the
+`inputs:` the grilling session already collected, run. Force a leaf to be vertical and it spans three
+skills, so the executor has to compose them itself — the interesting decisions move from the spec, where a
+human reviewed them, into an unattended run.
+
+**Vertical subtasks are what make review and verification meaningful.** An API resource reviewed alone
+cannot be judged against the thing that consumes it, and a scaffold with no data cannot be verified by
+looking at the running product. Grouping the leaves into a slice that renders gives both a real target: the
+review reads one coherent diff, and a `(human)` acceptance criterion has something to be true about.
+
+So the two levels answer two different questions. *What can an agent execute without judgement?* — a leaf.
+*What can a human judge?* — a subtask. Aligning both to the same axis would sacrifice one of them.
+
+## Consequences
+
+- Leaves are not run boundaries. A run executes a whole subtask; leaf checkboxes exist so a run interrupted
+  by a `[human]` leaf can resume, since nothing is committed until the subtask finishes.
+- The review unit is the subtask, so a leaf's code can be wrong for as long as it takes the slice to
+  finish. Accepted deliberately: reviewing every leaf spent three subagents per step, and most of what it
+  caught was churn the next leaf rewrote anyway.
+- Nesting is unnecessary. Work too big for one subtask becomes more subtasks with blocking edges between
+  them, never subtasks inside subtasks — which is what let the sub-branch and sub-PR machinery go.
--- a/.agents/adr/0002-layer-shaped-subtask-leaves.md
+++ b/.agents/adr/0002-layer-shaped-subtask-leaves.md
@@ -11,10 +11,6 @@
 
 **A subtask is a vertical slice; the leaves inside it are layer-shaped.**
 
-A subtask cuts a narrow but complete path through every layer it touches and is verifiable on its own. Its
-leaves do the opposite — each is one project skill's worth of work in one layer: `add-api-resource`, then
-`add-new-page`, then the `[human]` styling.
-
 ```
 subtasks/01-cross-chain-list/     ← vertical: demoable, one context window, one commit
   leaf 1 [agent] add-api-resource — declare the resource
@@ -23,8 +19,8 @@ subtasks/01-cross-chain-list/     ← vertical: demoable, one context window, on
   leaf 4 [human] style to mockup
 ```
 
-The full model lives in "The subtask model" in `.agents/tasks/README.md`; this record holds only the
-reasoning, which that file should not have to carry.
+"The subtask model" in `.agents/tasks/README.md` defines both levels and every rule that follows from them;
+this record holds only the reasoning, which that file should not have to carry.
 
 ## Why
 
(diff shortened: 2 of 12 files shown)


.agents/skills/grill-the-task/SKILL.md
```diff
@@ -107,25 +90,77 @@ Cover these domains:
...
+
+If the Slack MCP tools are unavailable, record the questions with owners anyway and tell the user to route
+them manually.
+
+Outreach is complete when every question has a recorded permalink — or an explicit note that the developer
+routes it manually.
+
+## Step 5 — Hand off
+
+**One subtask** — this session finishes the job. Create the feature branch (`issue-<number>` off `main`)
```
tom2drum: **blocker** — Step 5 only splits “One subtask” (implement + `create-pr`, no spec) vs “Several subtasks” (`to-spec`). Subtask mode is defined to produce a deferred folder’s `spec.md`, and `to-spec` says a subtask session hands off with that file to write — but Step 5 has no subtask-mode branch. Scoping one deferred subtask matches “One subtask” and wrongly implements/PRs instead of writing into the parent task.

Add a third handoff: subtask mode → always invoke `to-spec` (never the in-session implement path).

— Reviewed by Cursor Grok 4.5

.agents/skills/implement-task/SKILL.md
```diff
@@ -158,56 +160,58 @@ Nits never gate: `deferred` findings leave the `Outcome` `clear`.
 
 ### Step 8 — Update the spec
```
tom2drum, line 164: **blocker** — Step 8 checks the main-index box and sets Status `done` before Step 9’s `(human)` gate / commit. Index checkboxes are what `Blocked by:` and the frontier read (“subtasks that must be checked before it can start”), so dependents become eligible while the subtask is still unverified and uncommitted.

Check the index / set `done` only after commit (or after human verification confirms and the developer commits).

— Reviewed by Cursor Grok 4.5

.agents/skills/grill-the-task/SKILL.md
```diff
@@ -107,25 +90,77 @@ Cover these domains:
...
+
+Two things to look for while drafting it:
+
+- **Prefactor first.** *Make the change easy, then make the easy change.* When the current code fights the
+  feature, the first subtask reshapes it with no behaviour change — which also makes it the ideal opener for
+  an unattended chain, since it earns no `(human)` acceptance criteria and never pauses.
+- **Defer what can't be scoped.** A subtask blocked on a prototype, a spike, or an answer nobody has yet
+  gets a `brief.md` and no `spec.md`; a just-in-time subtask-mode session scopes it later.
+
+**Then front-load the executor skills' inputs — only when the breakdown has more than one subtask**, since
```
tom2drum: **major** — “front-load the executor skills' inputs — only when the breakdown has more than one subtask” / “A single-subtask task skips this entirely”. Subtask-mode scoping of one deferred leaf is not a single-subtask task — a later `implement-task` session must run blind — but the “more than one” guard skips `inputs:`.

Front-load whenever work will be executed in a later session (multi-subtask tasks and all subtask-mode runs); skip only task-mode one-subtask.

— Reviewed by Cursor Grok 4.5

.agents/skills/implement-task/SKILL.md
```diff
@@ -13,100 +13,99 @@ machine: each run starts a fresh session, picks up where the spec says work stop
...
 
 ### Step 2 — Reconcile the previous handoff
 
-If the previous subtask in the breakdown is `[human]` and still unchecked, ask the developer whether it's
-done before doing anything else — check it off if so, stop if it's still in progress (the order exists for a
-reason; don't leapfrog a pending style step unless the developer explicitly says the next subtask is
-independent).
+A run can arrive mid-subtask, because a `[human]` leaf inside one stops the chain. Before anything else:
+
+- **An unchecked `[human]` leaf in a subtask already in progress** — ask the developer whether it's done.
```
tom2drum, line 73: **major** — After checking off a `[human]` leaf, Step 2 only says “resume that subtask at its next unchecked leaf”. Default UI split ends on a `[human]` style leaf — there is no next leaf — and no branch says proceed to Steps 5–9 (verify/review/index/commit). Step 3’s finalize path for “final subtask ends on a `[human]` leaf” depends on that missing branch.

If no unchecked leaves remain after the handoff, continue at Step 5 for that subtask.

— Reviewed by Cursor Grok 4.5

.agents/adr/0002-layer-shaped-subtask-leaves.md
```diff
@@ -0,0 +1,57 @@
...
+| Status | accepted |
+| Date | 2026-08-11 |
+| Deciders | @tom2drum |
+| Supersedes | — |
+
+## Decision
+
+**A subtask is a vertical slice; the leaves inside it are layer-shaped.**
+
+A subtask cuts a narrow but complete path through every layer it touches and is verifiable on its own. Its
```
tom2drum: **major** — Possible Duplication — breaches `.agents/tasks/README.md` “This section is the only definition of the model — other files point here rather than restating it” and the ADR’s own “this record holds only the reasoning”: “A subtask cuts a narrow but complete path through every layer it touches and is verifiable on its own.” Same phrase also in `subtask-template.md` and the README.

Keep the bold Decision one-liner; delete the restating paragraph and point at README. Strip the same sentence from the template HTML comment.

— Reviewed by Cursor Grok 4.5

.agents/skills/to-spec/subtask-template.md
```diff
@@ -0,0 +1,69 @@
+# <NN> — <Subtask title>
+
+| | |
+| --- | --- |
+| Parent spec | `.agents/tasks/<dir>/spec.md` → link it as `../../spec.md`, subtask <NN> of #<issue> |
+| Status | `draft` \| `ready` \| `in progress` \| `done` |
```
tom2drum: **major** — Possible Incomplete case split — Status enum includes `ready`, but no skill transitions a subtask to it (`to-spec` only documents main-spec `draft`→`ready`; `implement-task` sets subtask `in progress`/`done` only).

Drop `ready` from the subtask template, or define who sets it and when.

— Reviewed by Cursor Grok 4.5

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `618a794f994db5df6b9a7579e45724de2cc9400b` → `9a9183a680d23a57c0342428f7c1d03f3d99215b` |
| Round | 1 of 3 |
| Findings | 2 blocker · 4 major · 2 nit |
| By axis | Spec skipped (no product-task spec) · Standards 3 · Correctness 4 |
| Outcome | `blocked` |

Instruction-surface PR only — Spec axis not dispatched.

.agents/AGENTS.md
```diff
@@ -77,8 +78,9 @@ looks wrong without its context. Supersede rather than rewrite — flip the old
 ## Product task workflow
 
 Product tasks (GitHub issues) are worked through a spec-driven workflow — interview, spec, agent
-implementation, code review. Specs accumulate in `.agents/tasks/` as a permanent record. See
-`./tasks/README.md` for the lifecycle, the skills that run it, and the spec conventions.
+implementation, code review. A task small enough to finish in its grilling session skips the spec
```
tom2drum: **nit** — Possible Shotgun Surgery / Divergent Change — prose-smells “AGENTS.md carries pointers, never mechanics”: “A task small enough to finish in its grilling session skips the spec entirely.” Restates README “Not every task needs a spec” instead of only pointing.

Delete that sentence; leave the pointer to `./tasks/README.md`.

— Reviewed by Cursor Grok 4.5

.agents/adr/0002-layer-shaped-subtask-leaves.md
```diff
@@ -0,0 +1,57 @@
...
+cannot be judged against the thing that consumes it, and a scaffold with no data cannot be verified by
+looking at the running product. Grouping the leaves into a slice that renders gives both a real target: the
+review reads one coherent diff, and a `(human)` acceptance criterion has something to be true about.
+
+So the two levels answer two different questions. *What can an agent execute without judgement?* — a leaf.
+*What can a human judge?* — a subtask. Aligning both to the same axis would sacrifice one of them.
+
+## Consequences
+
+- Leaves are not run boundaries. A run executes a whole subtask; leaf checkboxes exist so a run interrupted
```
tom2drum: **nit** — Possible Divergent Change — Consequences spell operational resume/commit rules (“Leaves are not run boundaries… leaf checkboxes exist so a run interrupted by a `[human]` leaf can resume”) that already live in README / `implement-task`. An ADR should keep the *why*; the workflow layer owns the mechanism.

Drop the operational resume bullets from the ADR; keep unique trade-off notes (review unit, no nesting) or point at README.

— Reviewed by Cursor Grok 4.5

.agents/skills/grill-the-task/SKILL.md
tom2drum: Fixed in 7f470d41f — Step 5 now has three branches, with subtask mode first: it always invokes `to-spec`, "however small the scoped subtask turns out to be", since the work is being written down for a later `implement-task` run. The remaining two are relabelled "Task mode, one subtask" / "Task mode, several subtasks" so neither can absorb a subtask-mode run.

.agents/skills/implement-task/SKILL.md
tom2drum: Fixed in 7f470d41f — the index checkbox now moves at commit time. Step 8 records only the subtask's own spec (leaf boxes, `Status: in progress`); the index box and `done` are set by Step 9 under `--auto`, and by the next run's Step 2 when the developer commits a verified subtask. Step 9's stop condition now reads "leave it uncommitted, its index box unchecked". The reasoning is stated in the skill so the ordering is not re-derived: a checked box means landed, and it is what releases dependents and finalizes the PR.

.agents/skills/grill-the-task/SKILL.md
tom2drum: Fixed in 7f470d41f — the gate is no longer subtask count but whether a later session executes the work: "every subtask-mode run, and any task-mode breakdown with more than one subtask". Only a task-mode single-subtask task skips it, since that session runs the executor skill itself.

.agents/skills/implement-task/SKILL.md
tom2drum: Fixed in 7f470d41f — Step 2 now branches after checking off the `[human]` leaf: continue at the next unchecked leaf, "or — when that leaf was the subtask's last, which the default UI split makes the common case — at Step 5", so the finished subtask still gets its verification, review, and commit.

.agents/adr/0002-layer-shaped-subtask-leaves.md
tom2drum: Fixed in 7f470d41f — the restating paragraph is gone from the ADR; the Decision is now the bold one-liner plus the diagram, closing with a pointer that names `.agents/tasks/README.md` as the definition of both levels. The same sentence is stripped from `subtask-template.md`, whose comment now points at "The subtask model" instead of restating it.

.agents/skills/to-spec/subtask-template.md
tom2drum: Fixed in 7f470d41f — `ready` dropped from the subtask template, leaving `draft | in progress | done`. Confirmed nothing set it: the only writer is `to-spec`, and that line is explicitly about the main spec ("once no `pending` question blocks the first subtask"). It also contradicted the model on purpose-built grounds — subtask readiness is derived from `Blocked by:` edges and question status, never stored.

.agents/AGENTS.md
tom2drum: Fixed in 7f470d41f (promoted from nit by the developer) — the sentence is removed; `AGENTS.md` carries only the pointer to `./tasks/README.md`.

.agents/adr/0002-layer-shaped-subtask-leaves.md
tom2drum: Fixed in 7f470d41f (promoted from nit by the developer) — Consequences now keep the trade-offs (review unit, no nesting) and drop the operational resume rules. The leaf bullet is reframed as the cost side of the review-unit trade-off, with the mechanism left to the workflow layer.

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `618a794f994db5df6b9a7579e45724de2cc9400b` → `7f470d41f830aca47bb41818e9e2aca15ef18862` |
| Round | 2 of 3 |
| By axis | Spec skipped · Standards n/a (arbitration) · Correctness n/a (arbitration) |
| Outcome | `clear` |

Arbitration of Round 1: all 8 claimed fixes verified in `7f470d41f` (subtask-mode handoff, index-at-commit, front-load gate, human-leaf → Step 5, ADR/template restatement, subtask `ready` dropped, AGENTS.md pointer-only, ADR Consequences trimmed). No regressions from the fix commit.

— Reviewed by Cursor Grok 4.5

PR #3635, opened 2026-08-14
Migrate the TAC operations views to Read API v2

### LINKED ISSUE #3627: Migrate TAC Operations UI to Read API v2
### Overview

The `tac-operation-lifecycle` service previously described an operation's lifecycle through a
single overloaded `type` field that mixed three unrelated concepts: the transfer **route**
(`TON_TAC_TON`, `TAC_TON`, `TON_TAC`), the **state** (`PENDING`, `ROLLBACK`), and a locally
derived **failure reason** (`INSUFFICIENT_FEE`), plus two catch-all values (`UNKNOWN`, `ERROR`).

That model has two defects the UI cannot work around:

1. **The route is unavailable until the operation completes.** A pending operation reports only
   `PENDING`, so the transfer direction cannot be shown while it matters most.
2. **A finalized failure without a rollback is indistinguishable from a success.** Both report
   the plain route value, so `type: "TON_TAC_TON"` has never meant "succeeded" — despite being
   read that way today.

The upstream Stage Profiler moved to a v2 contract that separates route, business outcome,
finality and rollback. The backend now consumes it, stores the facts separately, and exposes
them through a new **Read API v2** (backend PR:
https://github.com/blockscout/blockscout-rs/pull/1720, merged to `main`).

**Goal:** migrate the TAC operations list, details and by-transaction views to
`/api/v2/tac/...` and render the newly available lifecycle facts — independent status,
route-while-pending, rollback flag and failure reason — per the linked designs.

### To‑do / Implementation notes

- [ ] Point the API client at the v2 endpoints — paths, query params, pagination and search are
      **identical to v1**; only the operation object changes:
      - `GET /api/v2/tac/operations?q=&page_token=&page_items=`
      - `GET /api/v2/tac/operations/{operation_id}`
      - `GET /api/v2/tac/operations:byTx/{tx_hash}`
- [ ] Update types/schemas: four fields replace the single `type` (table below).
- [ ] Drive the status badge from `status` only — no UI element may infer an outcome from `type`.
- [ ] Drive the route icon/label from `type`, and render it for **pending** operations too, not
      only completed ones.
- [ ] Render `rollback: true` as a modifier on top of the failed state, **not** as a fourth
      status value. A `failed` operation may have `rollback` either `true` or `false`; both must
      render correctly.
- [ ] Display `error_reason` for failed operations, with a fallback string — the field is
      optional and legitimately absent in many failed states (see gotcha 2).
- [ ] Make the full per-stage failure text (`status_history[].note`) reachable from the details
      view.
- [ ] Handle `type: "UNKNOWN"` without layout breakage. This type means that the operation ID has been indexed, but the operation data has not been loaded yet. The same situation normally occurs when a newly created operation is being indexed and may persist for the first 1–2 seconds after it appears. API v1 behaved the same way. In the old UI, nothing was displayed in the operations table for the UNKNOWN type. In the new update, I suggest showing the `Pending` or `Fetching` status for such operations.
- [ ] Remove all code treating `PENDING`, `ROLLBACK`, `INSUFFICIENT_FEE` or `ERROR` as statuses —
      these values do not exist in v2.
- [ ] Regression-check search (`q`), pagination, sender rendering and the stage timeline. This is
      a re-binding, not a redesign of those parts.

### Operation object

| Field | Type | Required | Semantics |
|---|---|---|---|
| `operation_id` | `string` | yes | |
| `type` | `"UNKNOWN" \| "TON_TAC_TON" \| "TAC_TON" \| "TON_TAC"` | yes | **Route only.** Carries no outcome |
| `status` | `"pending" \| "success" \| "failed"` | yes | **Business outcome.** Lower-case values |
| `rollback` | `boolean` | yes | Whether a rollback occurred. Although technically this flag is independent of the `status` field, all Rollback operations should have the `failed` status. This should not be hardcoded in the UI, but it is worth using as a visual indicator of the indexer failure. |
| `timestamp` | `string` (RFC 3339, ms, UTC) | yes | e.g. `2026-08-07T12:34:56.789Z` |
| `sender` | `{ address: string, blockchain: "TAC" \| "TON" \| "UNKNOWN_BLOCKCHAIN" }` | no | |
| `error_reason` | `string` | no | Short failure label, e.g. `"Insufficient Fee"` |
| `status_history` | `V2OperationStage[]` | details endpoints only | Stage timeline (has no differences from v1) |

### Example response

Success operation

```json
{
    "items":
    [
        {
            "operation_id": "0xf01646ac36cbbcebd8a5ff09c300d5f3bebdd2fbe0135a377eaa485d9edcc670",
            "type": "TAC_TON",
            "status": "success",
            "rollback": false,
            "timestamp": "2026-08-11T02:31:29.000Z",
            "sender":
            {
                "address": "0x0dc5e4a7efe4d79371c88453bc8aee5bb8fb4f12",
                "blockchain": "TAC"
            },
            "error_reason": null
        },
        {
            "operation_id": "0xd93005d467abe050748b414acf5f0a62755a3af4efb63540e89f6c315ce214b8",
            "type": "TON_TAC_TON",
            "status": "failed",
            "rollback": true,
            "timestamp": "2026-08-11T02:52:33.000Z",
            "sender":
            {
                "address": "EQBW7ZPk4P8sUNVoQtrxJbvODj1fLYFD8IYCThSUtrO+bXTZ",
                "blockchain": "TON"
            },
            "error_reason": null
        },
        {
            "operation_id": "0xc950fbe77004b963342be7d70f2750d802212d6d3fbcf3eb0f691f2d7b55e778",
            "type": "TON_TAC",
            "status": "failed",
            "rollback": false,
            "timestamp": "2026-08-11T02:39:07.000Z",
            "sender": {
                "address": "EQDoF2OkxsI3gc5jAuxlqozN9H/SgEOUCopMa1yU4djLaXuL",
                "blockchain": "TON"
            },
            "error_reason": "Insufficient Fee"
        },
        {
            "operation_id": "0xf565590f3e6dceaf5dde0b5a6f72bf84e0043029bc7cff5999810e176ac61efe",
            "type": "TAC_TON",
            "status": "pending",
            "rollback": false,
            "timestamp": "2026-08-11T02:36:28.000Z",
            "sender":
            {
                "address": "0x0dc5e4a7efe4d79371c88453bc8aee5bb8fb4f12",
                "blockchain": "TAC"
            },
            "error_reason": null
        },
        {
            "operation_id": "0xc53ed4cb66dcb1aa796f1205147902753c8f2b21dfb6cd407128bdd5a552d7b9",
            "type": "TON_TAC_TON",
            "status": "failed",
            "rollback": true,
            "timestamp": "2026-08-11T02:18:18.000Z",
            "sender":
            {
                "address": "EQAGcbFm27NYbQXVBLkC/xiXN+jpRABo/cXB0ZXjbWhBo5lT",
                "blockchain": "TON"
            },
            "error_reason": "Insufficient Fee"
        }
    ],
    "next_page_params":
        {
            "page_token": 1754570110,
            "page_items": 50
        }
}
```

### Migration mapping

| v1 `type` | v2 equivalent |
|---|---|
| `PENDING` | `status: "pending"`; `type` may already carry a concrete route |
| `INSUFFICIENT_FEE` | `status: "failed"`, `error_reason: "Insufficient Fee"` |
| `ROLLBACK` | `status: "failed"`, `rollback: true`; the route is preserved in `type` |
| `TON_TAC_TON` / `TAC_TON` / `TON_TAC` | `type` = route; the outcome is now in `status` and **may be `failed`** |
| `UNKNOWN` | `type: "UNKNOWN"` |
| `ERROR` | Not produced in v2; such an operation reads as `pending` / `UNKNOWN` |

### Gotchas — deliberate backend decisions, not bugs

1. **`failed` does not wait for finality; `success` does.** An operation reads as `failed` as
   soon as the outcome is known, even while the indexer still polls it upstream. A successful
   but not-yet-final operation reads as `pending`. Observable transitions are therefore
   `pending → failed` or `pending → success`, and `success` is terminal. The asymmetry is
   intentional: a user cannot act on a failed operation, and "failed and pending at once" is not
   a state worth showing.
2. **`error_reason` may be missing on a failed operation.** It is published only when the stored
   value is at most 16 characters, because longer values are raw upstream payloads (serialized
   revert data, whole message bodies) rather than labels. Truncating them was explicitly
   rejected. Treat the field as optional in every failed state; the full text stays available
   per stage via `note`.
3. **`type: "UNKNOWN"` has two causes** — "not profiled yet" or "a route this API version does
   not know". They are not distinguishable and do not need to be distinguished in the UI.
4. **Legacy rows may report `success` for an operation that actually failed.** Operations indexed
   under the v1 upstream contract are mapped rather than hidden, and that contract could not
   express "finalized failure without rollback". This reproduces exactly what the current UI
   already shows, so it is not a regression. A background re-profiling worker converts these rows
   over time — transitional, no frontend handling required.
5. **Stage `timestamp` equals `null` when the stage has no transactions.** Pre-existing v1

### Mockups / Screenshots / Links

https://www.figma.com/design/1UWWsK0bg6ifzS9O1NLlo4/TAC-TON-TAC-operations?node-id=4001-37444&t=u4szFuUtOgKFb0P4-4

Image

Image

Image

Image

### Deployment info

The staging instance of `tac-operation-lifecycle` (SPB chain) was successfully updated and re-synced with the v2 stage-profiling upstream service. Full backward compatibility with the previous version has been preserved. (**NOTE:** This instance is not the same one used in production for the SPB testnet)

**API:** https://tac-spb-operation-lifecycle.k8s-dev.blockscout.com/
**Swagger:** https://github.com/blockscout/swaggers/blob/master/services/tac-operation-lifecycle/1.2.0/swagger.yaml
**Indexer PR:** https://github.com/blockscout/blockscout-rs/pull/1720

## Deployment info

- **No feature flag and no coordinated cutover.** Read API v1 is **byte-for-byte unchanged** — no
  path, message or enum number was modified, verified against the generated Swagger diff
  (additive only). Internally the v1 `type` is now computed per request instead of being
  materialized at index time, which is not observable from outside.
- **Ordering:** the backend must be deployed (and its DB migration applied) before the frontend
  switches to v2. v1 and v2 can then be served in parallel indefinitely.
- **No frontend env/config change** is required beyond the API base path already in use — the v2
  routes live on the same host as v1.
- **v1 sunset:** no date yet; to be agreed separately once the frontend has migrated. The v1
  client can stay in the codebase until then.
- **During rollout**, historical operations are gradually re-profiled into the v2 model by a
  background worker.

## Acceptance criteria

- All operation views read from `/api/v2/tac/...`.
- Status is rendered from `status`; no code path derives status from `type`.
- Route is rendered from `type` and is visible for `pending` operations.
- `rollback: true` is rendered as specified in the designs, for both rollback and non-rollback
  failures.
- `error_reason` is rendered when present, with a defined fallback when absent.
- `type: "UNKNOWN"` renders without layout breakage.
- Legacy status values (`PENDING`, `ROLLBACK`, `INSUFFICIENT_FEE`, `ERROR`) are gone from the
  frontend codebase.
- The details view exposes the per-stage `note` for failed stages.
- No regression in search (`q`), pagination, sender rendering or the stage timeline versus the
  current v1-backed views.

tom2drum:
## Description

Resolves #3627

The `tac-operation-lifecycle` service described an operation's whole lifecycle through one overloaded `type` field that mixed three unrelated things — the transfer **route** (`TON_TAC_TON`, `TAC_TON`, `TON_TAC`), the **state** (`PENDING`, `ROLLBACK`) and a locally derived **failure reason** (`INSUFFICIENT_FEE`). Two defects followed and could not be worked around in the UI: the route was unknowable until an operation completed, so the direction could not be shown while it mattered most; and a finalized failure without a rollback was indistinguishable from a success, so `type: "TON_TAC_TON"` never actually meant "succeeded" despite being read that way.

Read API v2 separates route, business outcome, finality and rollback. The operations list, the operation details page, the by-transaction block and the search surfaces now read `/api/v2/tac/...` and render the newly available facts:

- The status indicator is driven by `status` alone — no code path derives an outcome from `type` any more.
- The route comes from `type` and renders for **pending** operations too, not only completed ones.
- Status and route stay in **one combined tag** (a status icon and colour wrapping the route text) — a standing product decision; an earlier design split them into two fields and was rejected.
- `rollback: true` renders as a separate badge beside the status tag, never as a fourth status value. A `failed` operation may have `rollback` either `true` or `false`, and both render.
- `error_reason` appears in the status tag's tooltip when present, appended to the failure text; when absent the tooltip carries the plain failure text. The field is optional and legitimately missing in many failed states.
- `type: "UNKNOWN"` renders the status word in the neutral pending presentation with no route — it means the operation id is indexed but its data has not loaded yet.

Two decisions a reader would otherwise have to reverse-engineer:

- **`failed` is shown without hedging**, even though the indexer may later flip it to `success`. Showing a spinner was considered and rejected: the user can neither act on it nor know what they are waiting for.
- **The search result keeps a feature-owned type** instead of using the generated schema. Core's search union discriminates on `type`, and the proxied operation object carries its own `type` for the transfer route; `openapi-typescript` resolved the collision by overwriting the route enum with the discriminator literal `"tac_operation"` and flattening the object, losing the `tac_operation` wrapper and `priority`. The wire format is nested and does carry the route — core's own `search_controller_test.exs` asserts it — so the generated type is what is wrong, not the API. Worth fixing upstream; details in `.agents/tasks/3627-tac-operations-api-v2/subtasks/06-search-surfaces/notes.md`.

Plan and per-subtask records: [`.agents/tasks/3627-tac-operations-api-v2/spec.md`](.agents/tasks/3627-tac-operations-api-v2/spec.md).

## Environment variables

None.

## Minimum API version

- `tac-operation-lifecycle` **v1.2.0+** — the service release that serves Read API v2.
- Core **v11.2.8+** — the release whose `/api/v2/search` returns the v2 operation shape.

**This must ship together with the backends.** Two separate constraints:

- The **service** rollout is safe to do early: Read API v1 is byte-for-byte unchanged, so v1 and v2 can be served in parallel indefinitely. The service still has to reach the instances the frontends point at before this merges.
- **Core is not parallel-safe.** Core replaces its `/api/v1/tac/operations` call rather than serving both, so an instance running a core older than v11.2.8 would hand this frontend a v1 operation object that it no longer parses. There is deliberately no version guard or fallback.

## Additional information

Legacy values `PENDING`, `ROLLBACK`, `INSUFFICIENT_FEE` and `ERROR` are gone from every code path in `src/`, and the v1 client and its generated API docs are removed.

`TacOperation.pw.tsx`'s screenshot baselines were regenerated — its fixture is now a *failed* operation carrying an error reason, so the previous baselines no longer described the page. `TacOperationStatus.pw.tsx` and its five baselines were removed: the tag's text, tooltip, and status/route/rollback matrix are covered by `TacOperationStatus.spec.tsx` (Vitest) instead, which is cheaper than a screenshot per text permutation.


src/slices/search/pages/search-results/SearchResultListItem.tsx
```diff
@@ -259,7 +260,14 @@ const SearchResultListItem = ({ data, searchTerm, isLoading, addressFormat }: Pr
...
             </TacOperationEntity.Link>
-            <TacOperationStatus status={ data.tac_operation.type }/>
+            <TacOperationStatus
+              status={ data.tac_operation.status }
+              type={ data.tac_operation.type }
+              errorReason={ data.tac_operation.error_reason }
+              isRollback={ data.tac_operation.rollback }
+              isLoading={ isLoading }
+            />
+            { data.tac_operation.rollback && <Badge loading={ isLoading }>Rollback</Badge> }
```
tom2drum, line 270: **major** — `{ data.tac_operation.rollback && <Badge loading={ isLoading }>Rollback</Badge> }`: the search slice hand-renders a TAC-feature badge, and so does `SearchResultTableItem.tsx:388`. `src/slices/CONTEXT.md` (Child-slice ownership): "The slice that owns an entity owns its **rendering** — tables, lists, detail panels, types. Other slices or features that surface the entity … import those views; they never reimplement them."

The cause is that the rollback tag was never extracted. The task spec's *UI inventory* says "**The rollback tag** is its own component, rendered as a sibling — the details page title uses it without the status tag, so it cannot be baked in", and subtask 02's leaf 2 is ticked as "extract the rollback tag as its own component". At HEAD the literal is written out at seven call sites — `TacOperationDetails.tsx:64`, `TacOperationsListItem.tsx:50`, `TacOperationsTableItem.tsx:34`, `TxDetailsTacOperation.tsx:79`, `SearchBarSuggestTacOperation.tsx:32`, this line, and `SearchResultTableItem.tsx:388` — plus an eighth copy inside `TacOperationTag.tsx:20`. A `TacOperationRollbackTag` existed earlier on this branch and was removed in 64e7870.

**Fix.** Reinstate the feature-owned rollback tag and render it at all eight sites, so no slice carries tac markup.

— Reviewed by Cursor agent (`review-changes`)

src/features/chain-variants/tac/utils/tac-operation.ts
```diff
@@ -4,55 +4,89 @@ import * as tac from '@blockscout/tac-operation-lifecycle-types';
...
+ */
+export function getTacOperationStatusTooltip(
+  status: tac.V2OperationStatus,
+  errorReason: string | null | undefined,
+  isRollback: boolean | undefined,
+): string | null {
+  if (status !== tac.V2OperationStatus.failed) {
+    return null;
+  }
+  if (isRollback) {
```
tom2drum, line 56: **major** (needs-human) — `if (isRollback) { return ROLLBACK_TOOLTIP; }` discards `error_reason` for every rollback. Requirement 7: "`error_reason` is rendered inside the status tag's tooltip when present, appended to the failure text", and subtask 02's ticked criterion repeats it. A `failed` operation with `rollback: true` and `error_reason: "Insufficient Fee"` now shows only the generic rollback sentence, and `TacOperationStatus.spec.tsx:57` pins that as intended ("prefers the rollback text over an error reason").

The spec also puts the rollback wording on the rollback tag rather than the status tag — subtask 02's human criterion covers "the rollback tag's placement and tooltip copy" — so threading `isRollback` through the status tooltip splits one concept across two components. Related gap from the same early return: a rollback on a `success` or `pending` operation gets no tooltip at all, which the *Data & API* table warns against ("Independent of `status` … Do not hardcode that relationship").

**Fix.** Keep the status tooltip as failure-plus-reason, move `ROLLBACK_TOOLTIP` onto the rollback tag, and extend the tooltip matrix to rollback × non-failed. Where the rollback copy belongs is a product call, hence needs-human.

— Reviewed by Cursor agent (`review-changes`)

src/features/chain-variants/tac/mocks/operations.ts
```diff
@@ -1,53 +1,56 @@
 import * as tac from '@blockscout/tac-operation-lifecycle-types';
 
-export const tacOperation: tac.OperationDetails = {
+export const tacOperation: tac.V2OperationDetails = {
   operation_id: '0x35f5d9c2bf07477ede48935c7130945faf17a3e5f69a7d20ce3725676513095c',
-  type: tac.OperationType.TON_TAC_TON,
+  type: tac.V2OperationType.TON_TAC_TON,
+  status: tac.V2OperationStatus.failed,
+  rollback: false,
```
tom2drum, line 7: **nit** — `rollback: false` here, and in `stubs.ts:9`, `stubs.ts:21` and `mocks/search.ts:10`: no fixture or stub anywhere sets `rollback: true`, so the one branch every surface gained in this PR (requirement 5, "`rollback: true` renders as a separate tag beside the status tag") is exercised by nothing — no Vitest render, no screenshot, no placeholder state.

Separately, making this shared details mock `status: failed` with an `error_reason` is what forces `TacOperation.pw.tsx:36-40` to spread and override it back for the pending case.

**Fix.** Add a rollback fixture, or keep the base mock a plain success and derive the failed and rollback variants from it.

— Reviewed by Cursor agent (`review-changes`)

src/features/chain-variants/tac/components/TacOperationStatus.spec.tsx
```diff
@@ -0,0 +1,111 @@
+// @vitest-environment jsdom
+
+import React from 'react';
+
+import * as tac from '@blockscout/tac-operation-lifecycle-types';
+
+import { afterEach, describe, expect, it } from 'vitest';
+import { cleanup, render, screen } from 'vitest/lib';
+
+import { FAILURE_TOOLTIP, getTacOperationStatusText, getTacOperationStatusTooltip, ROLLBACK_TOOLTIP } from '../utils/tac-operation';
```
tom2drum, line 10: **nit** — two of the three `describe` blocks (`status text`, `status tooltip`) exercise `../utils/tac-operation`, not the component this file is named after. `.agents/rules/tests-unit.md`: "Test files must be named `*.spec.ts` or `*.spec.tsx` and placed alongside the code they test." The repo's convention is a spec beside the util (e.g. `src/slices/address/utils/get-address-name.spec.ts`), and this is the only component spec in `src/features/**` that reaches into a sibling `utils/` module.

**Fix.** Move the two util `describe`s into `utils/tac-operation.spec.ts`, leaving the `rendering` block here.

— Reviewed by Cursor agent (`review-changes`)

tom2drum (commented):
## Review — whole PR

| | |
| --- | --- |
| Reviewed | `92aa834c7` (merge-base with `main`) → `8912ab86c` |
| Round | 1 — whole-PR review; subtasks 01 and 06 were reviewed separately in their own records |
| Findings | 0 blocker · 2 major · 2 nit |
| By axis | Spec 2 · Standards 1 · Correctness 1 |
| Outcome | `blocked` — two majors open, one of them needing a product call |

Both majors sit in the same area: the rollback tag was never extracted into the component the spec asks for, and the rollback wording ended up inside the status tooltip where it displaces `error_reason`. The two nits are fixture and test-placement hygiene.

CI owns the checks; this review does not re-run or report them.

## Not anchorable

**nit** (needs-human) — `src/features/chain-variants/tac/components/TacOperationStatus.pw.tsx` is deleted along with its five baselines, so the shared status tag now has no visual coverage in any state. The parent spec's *Tests* section keeps "Screenshot cases stay minimal — one per visual variant", and subtask 02's leaf 5 is ticked as "Extend the `TacOperationStatus.pw.tsx` cases to one per visual variant". The PR description argues the Vitest spec replaces "a screenshot per text permutation" — which the spec already agreed with — but that is a different claim from dropping the per-visual-variant cases. If the deletion is deliberate, the two ticked leaves (02.5 here, 02.2 in the major below) should be corrected so the permanent record doesn't claim work the tree doesn't contain.

— Reviewed by Cursor agent (`review-changes`)

src/slices/search/pages/search-results/SearchResultListItem.tsx
tom2drum: **Rejected.** The finding's premise is the spec, and the spec is what's wrong here. "The rollback tag is its own component" and subtask 02's leaf 2 both specify a *how*, not a *what* — implementation detail that shouldn't have been written into a spec in the first place. The product requirement is that `rollback: true` renders as a separate tag beside the status tag; whether that's a shared component or a one-line `Badge` is an implementation choice, made deliberately in 64e7870.

On the ownership rule specifically: `<Badge>Rollback</Badge>` uses a shared toolkit primitive with a literal label. `src/slices/CONTEXT.md` is about not reimplementing an entity's **views** — "tables, lists, detail panels, types" — and a one-word badge isn't one. The TAC entity views the search rows do surface (`TacOperationEntity`, `TacOperationStatus`) are imported from the feature, as the rule requires.

Follow-up: the spec's *UI inventory* line and subtask 02's leaf 2 will be corrected rather than the code.

src/features/chain-variants/tac/utils/tac-operation.ts
tom2drum: **Rejected.** The precedence is intentional, not an oversight. A rollback is the more significant fact about an operation than any `error_reason` string, so when both are present the rollback wording wins the status tooltip — and the tooltip belongs on the status/route tag, where the user is already looking. `TacOperationStatus.spec.tsx:57` encodes that on purpose.

Requirement 7 in the parent spec ("appended to the failure text") is the thing that's out of date; it predates the rollback tag existing as a separate signal. The same applies to the suggestion to move `ROLLBACK_TOOLTIP` onto the rollback tag.

Follow-up: the spec will be corrected to match.

src/features/chain-variants/tac/mocks/operations.ts
tom2drum: **Deferred** — a fair catch, but not this change's work. No fixture sets `rollback: true` anywhere, so the branch is genuinely uncovered; picking it up means a rollback fixture plus a render assertion, which is worth doing on its own rather than bolted onto this PR. Leaving the thread open so it stays visible.

src/features/chain-variants/tac/components/TacOperationStatus.spec.tsx
tom2drum: **Deferred** — correct about the convention, but cosmetic here, and not this change's work. Splitting the two util `describe`s into `utils/tac-operation.spec.ts` is a clean follow-up. Leaving the thread open so it stays visible.

PR #3636, opened 2026-08-14
Keep scam tokens out of the address token list when they arrive over the socket

tom2drum:
## Description

With "Hide scam tokens" on, the address token list could still show scam tokens a few seconds after load, without any API request.

The REST response is filtered by the backend (the `show-scam-tokens` header is omitted, so scam tokens are dropped). The websocket can carry no headers, so `updated_token_balances_*` pushes the address's balances unfiltered — and `useFetchTokens` merges those straight into the same React Query cache entry the ERC-20 tab reads, re-adding what the backend had removed.

Socket payloads are now filtered client-side when the setting is on, mirroring what the address token-transfers list already does (`useAddressTokenTransfersSocket`).

Reproduced and verified on OP Sepolia dev: the address receives a scam USDC balance push every few seconds; before the fix the row appeared and the widget count went 2 → 3, after the fix the list stays filtered across 40+ pushes.

## Environment variables

None

## Minimum API version

None

None

## Additional information

Reported in Slack; the same guard applies to every token type this hook handles (ERC-721/1155/404 + custom types), not just ERC-20.

### DIFF
--- a/src/slices/token/pages/address/useFetchTokens.ts
+++ b/src/slices/token/pages/address/useFetchTokens.ts
@@ -13,9 +13,12 @@ import useApiQuery, { getResourceKey } from 'src/api/hooks/useApiQuery';
 import useSocketChannel from 'src/api/socket/useSocketChannel';
 import useSocketMessage from 'src/api/socket/useSocketMessage';
 
+import { useAppContext } from 'src/shell/app/context';
+
 import { useMultichainContext } from 'src/features/multichain/context';
 
 import config from 'src/config';
+import * as cookies from 'src/shared/storage/cookies';
 
 import type { TokenEnhancedData } from './utils';
 import { calculateUsdValue } from './utils';
@@ -24,6 +27,8 @@ interface Props {
   hash?: string;
   enabled?: boolean;
 }
+
+const EMPTY_TOKEN_BALANCES: Array<schemas['TokenBalance']> = [];
 
 const tokenBalanceItemIdentityFactory = (match: schemas['TokenBalance']) => (item: schemas['TokenBalance']) => ((
   match.token && item.token &&
@@ -40,6 +45,12 @@ const socketEventForTokenType = (tokenTypeId: string): string => {
 const additionalTypes = config.slices.token.additionalTypes;
 
 export default function useFetchTokens({ hash, enabled }: Props) {
+  const { cookies: appCookies } = useAppContext();
+
+  // the socket carries no request headers, so the backend pushes balances unfiltered —
+  // the scam filter it applies to the REST response has to be repeated here
+  const shouldHideScamTokens = config.slices.token.hideScamTokensEnabled && !(cookies.get(cookies.NAMES.SHOW_SCAM_TOKENS, appCookies) === 'true');
+
   const erc20query = useApiQuery('core:address_tokens', {
     pathParams: { hash },
     queryParams: { type: [ 'ERC-20' ] },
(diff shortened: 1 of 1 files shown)


tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `fe1b07122fad462df6a0dce1021904040f8afc16` → `52caa38001683c96c2ce66e5d6660932d1db504e` |
| Round | 1 of 3 |
| Findings | 0 blocker · 2 major · 1 nit |
| By axis | Spec (skipped — no task spec) · Standards 1 · Correctness 2 |
| Outcome | `needs-human` |

## Not anchorable

None.

src/slices/token/pages/address/useFetchTokens.ts
```diff
@@ -85,17 +94,21 @@ export default function useFetchTokens({ hash, enabled }: Props) {
...
+        const updatedData = tokenBalances.find(tokenBalanceItemIdentityFactory(currentItem));
         return updatedData ?? currentItem;
       }) || [];
 
       const extraItems = prevData?.next_page_params ?
         [] :
-        payload.token_balances.filter((socketItem) => !items.some(tokenBalanceItemIdentityFactory(socketItem)));
+        tokenBalances.filter((socketItem) => !items.some(tokenBalanceItemIdentityFactory(socketItem)));
 
       if (!prevData) {
```
tom2drum, line 122: **major** — `if (!prevData) { return { items: extraItems, next_page_params: null }; }` — with hide on, an all-scam socket payload filters to `tokenBalances=[]`, so `extraItems` is empty and this still seeds the React Query cache as a successful empty page. That can clear the loading path before REST lands (`isPending` flips once data exists). Pre-filter, the same race seeded scam rows; the new filter turns it into a false empty success.

If `!prevData`, return `undefined` (skip the write), or only seed when `extraItems.length > 0`.

— Reviewed by Cursor Grok 4.5

src/slices/token/pages/address/useFetchTokens.ts
```diff
@@ -85,17 +94,21 @@ export default function useFetchTokens({ hash, enabled }: Props) {
...
+      payload.token_balances.filter((item) => item.token?.reputation !== 'scam') :
+      payload.token_balances;
+
     queryClient.setQueryData(queryKey, (
       prevData: paths['/api/v2/addresses/{address_hash_param}/tokens']['get'] | undefined,
     ) => {
       const items = prevData?.items.map((currentItem) => {
-        const updatedData = payload.token_balances.find(tokenBalanceItemIdentityFactory(currentItem));
+        const updatedData = tokenBalances.find(tokenBalanceItemIdentityFactory(currentItem));
         return updatedData ?? currentItem;
```
tom2drum: **major** — `return updatedData ?? currentItem` after `payload.token_balances.filter((item) => item.token?.reputation !== 'scam')` — scam rows are stripped from the merge input only. An identity already in `prevData.items` whose socket update is scam keeps the old row (`updatedData` is undefined). Hide is incomplete for "in cache already / reputation flipped to scam on the socket".

When hiding, drop `currentItem`s that match a scam row in the raw payload, then merge the non-scam balances.

— Reviewed by Cursor Grok 4.5

src/slices/token/pages/address/useFetchTokens.ts
```diff
@@ -40,6 +43,12 @@ const socketEventForTokenType = (tokenTypeId: string): string => {
 const additionalTypes = config.slices.token.additionalTypes;
 
 export default function useFetchTokens({ hash, enabled }: Props) {
+  const { cookies: appCookies } = useAppContext();
+
+  // the socket carries no request headers, so the backend pushes balances unfiltered —
+  // the scam filter it applies to the REST response has to be repeated here
+  const shouldHideScamTokens = config.slices.token.hideScamTokensEnabled && !(cookies.get(cookies.NAMES.SHOW_SCAM_TOKENS, appCookies) === 'true');
```
tom2drum, line 52: **nit** — possible Duplicated Code: `config.slices.token.hideScamTokensEnabled && !(cookies.get(cookies.NAMES.SHOW_SCAM_TOKENS, appCookies) === 'true')` is identical to `useAddressTokenTransfersSocket.ts:61`.

Extract a shared helper under `src/slices/token/` and call it from both socket hooks.

— Reviewed by Cursor Grok 4.5

src/slices/token/pages/address/useFetchTokens.ts
tom2drum: Fixed in `5f9c33e1f`. When `prevData` is undefined and there is nothing to add, the updater now returns `undefined`, which React Query treats as a no-op — so an all-scam payload no longer seeds an empty page and ends the loading state of a request still in flight.

src/slices/token/pages/address/useFetchTokens.ts
tom2drum: Fixed in `5f9c33e1f`. Cached rows whose identity matches a scam entry in the raw payload are now dropped before the merge, so a token flagged as scam while the page is open disappears instead of keeping its stale row.

src/slices/token/pages/address/useFetchTokens.ts
tom2drum: Deferred — the duplication is real and a shared helper is the right call, but it also rewrites `useAddressTokenTransfersSocket`, which is outside the scope of this fix. Leaving the thread open so it is not lost.

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `fe1b07122fad462df6a0dce1021904040f8afc16` → `5f9c33e1f56a76032ce8677ce2e09d2e1dac3800` |
| Round | 2 of 3 |
| Findings | 0 blocker · 0 major · 1 nit (deferred) |
| By axis | Spec (skipped) · Standards 1 deferred · Correctness 0 open |
| Outcome | `clear` |

## Arbitration

- **F1 (empty seed on `!prevData`)** — fixed. `extraItems.length === 0` → `return undefined` (RQ setQueryData no-op).
- **F2 (stale cached scam row)** — fixed. `scamBalances` filter drops matching cached identities before merge.
- **F3 (duplicated `shouldHideScamTokens`)** — deferred-ok. Shared helper would rewrite `useAddressTokenTransfersSocket`; out of scope for this PR.

## Not anchorable

None.

src/slices/token/pages/address/useFetchTokens.ts
tom2drum: **Arbitration (R2)** — fixed. Confirmed `if (!prevData) { if (extraItems.length === 0) return undefined; }` at the current HEAD; React Query treats `undefined` from the updater as a no-op, so an all-scam filtered payload no longer seeds an empty success page.

— Reviewed by Cursor Grok 4.5

src/slices/token/pages/address/useFetchTokens.ts
tom2drum: **Arbitration (R2)** — fixed. Confirmed `prevData?.items.filter(...!scamBalances.some...).map(...)` drops cached identities that match scam rows in the raw payload before merge.

— Reviewed by Cursor Grok 4.5

src/slices/token/pages/address/useFetchTokens.ts
tom2drum: **Arbitration (R2)** — deferred-ok. Duplication still present; extracting a shared helper would also rewrite `useAddressTokenTransfersSocket`, which is outside this PR’s scope. Thread stays open as deferred.

— Reviewed by Cursor Grok 4.5

PR #3637, opened 2026-08-15
Replace the Slack-only issue skill with a generic create-issue workflow

tom2drum:
## Description

The old skill only knew how to turn a Slack thread into a GitHub issue. Filing from an agent session, meeting notes, or a Notion/Fireflies pointer meant ignoring that workflow or stretching Slack-specific steps.

`create-issue` takes whatever the user pointed at (or this conversation), files a thin public-safe issue, and stops. Slack URL decoding stays as a gotcha note; other sources use connected tools or a paste fallback. Issues in `blockscout/frontend` and `blockscout/blockscout` are added to their project boards with Status left unset for weekly triage. Type (`Bug` / `Task` / `Feature`) is inferred and confirmed; effort is not guessed.

## Environment variables

None

## Minimum API version

None

None. The Slack-only skill name is gone — invoke `create-issue` (or ask to file an issue in ordinary language).

## Additional information

None

Made with [Cursor](https://cursor.com)


.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+
+If the repo is in the table, add the new issue to the specified project board:
+
+| Repo | Owner | Project number |
+|---|---|---|
+| `blockscout/frontend` | `blockscout` | 6 |
+| `blockscout/blockscout` | `blockscout` | 8 |
+
+```bash
+gh project item-add <number> --owner <owner> --url <issue-url>
```
tom2drum, line 96: **blocker** — `gh project item-add` needs the `project` OAuth scope, but the only auth gate is line 33's `check-github-cli`, which stops at `gh auth status`. That check passes today while this step cannot: this repo's `gh auth status` reports `Token scopes: 'gist', 'read:org', 'read:packages', 'repo', 'workflow'` — no `project`. `gh issue create --help` says it outright: "Adding an issue to projects requires authorization with the `project` scope."

So the capability added by f289b661a fails on every run, and line 97 ("A board failure does not undo the issue") silently absorbs it — the issue gets filed, the board step quietly never works, and nobody notices.

**Suggested fix.** Before the `item-add` block, require `gh auth status` to list a `project` scope and tell the user to run `gh auth refresh -s project` if it is absent.

— Reviewed by Claude Opus 5

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+```bash
+gh issue create \
+  --repo <owner>/<name> \
+  --title "<title>" \
+  --body-file <path> \
+  --type "<Bug|Task|Feature>" \
+  --label "<label>"
+```
+
+Omit `--label` when there are none. Repeat `--label` for each label. Always pass `--type`. Write the body to a temp file so shell escaping cannot mangle it.
```
tom2drum: **major** — "Always pass `--type`" is unconditional, but line 35 lets the repository be any `owner/name` the user types. `--type` resolves a name against the *owner's* issue types; `Bug`/`Task`/`Feature` are only verified enabled for the `blockscout` org. Against a repo under another owner, or one with issue types disabled, `gh issue create --type "Task"` errors *before* the issue is created — unlike the board step, this failure loses the whole issue, and there is no fallback.

**Suggested fix.** Make `--type` conditional — pass it for `blockscout/*` repos, and on a type-resolution error retry the same `gh issue create` without `--type`.

— Reviewed by Claude Opus 5

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+
+**Repository.** Recommend one: this workspace's `origin` when it is a `blockscout/*` repo and the topic fits; otherwise a best-guess `blockscout/*` repo from the topic. Skip fork remotes. If they say it is the wrong place, list `blockscout` source repos (`gh repo list blockscout --source --no-archived --limit 100 --json name,description`) or take an `owner/name` they type.
+
+**Type.** Infer one of `Bug`, `Task`, or `Feature` from the topic: unexpected broken behavior → `Bug`; new user-facing capability → `Feature`; otherwise `Task`.
+
+**Labels.** `gh label list --repo <owner>/<name> --json name,description --limit 100`. Pick labels that match the topic; zero is fine.
+
+**Title.** Imperative mood ("Fix X", "Add support for Y").
+
+**Body.** Neutral third-person technical language that can stand on a public tracker. From this conversation, take the ask plus concrete technical facts already established (errors, affected surface) — not the debugging transcript or references to the existing code. Omit empty sections:
```
tom2drum: **major** — "From this conversation, take the ask plus concrete technical facts already established" contradicts Step 1, which resolves the source to a *fetched pointer* instead of the conversation — line 19: "Use only that pointer, or this conversation."

Followed literally when a Slack URL was given, the body gets drafted from the chat rather than the thread that was just read. This is leftover phrasing from the pre-6ed565924 Slack-only skill, where the two always coincided.

**Suggested fix.** Replace "From this conversation" with "From the source", so it tracks whatever Step 1 resolved.

— Reviewed by Claude Opus 5

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Step 2 — Pick the topic
+
+Done when exactly one topic is selected, or the run has stopped.
+
+- Several issue-worthy topics → list them in one short round and wait. "All of them" means a separate issue per topic: confirm and create each before starting the next.
+- File when the problem or request can be stated in a couple of sentences. Missing repro, acceptance criteria, or technical detail is fine — omit those sections later.
+- No request or problem in the source, or the subject itself cannot be named → stop and tell them the issue cannot be created.
+
+## Step 3 — Draft
```
tom2drum, line 31: **major** — Step 3 is the longest step — it produces repo, type, labels, title and body — yet it carries no completion criterion, while Steps 1 and 2 both open with "Done when …". `writing-for-agents/SKILL.md`: "Every step ends on a **completion criterion** — the condition that tells the agent the work is done." Also `prose-smells.md`, "Uncheckable completion criterion".

Without one, the agent can enter Step 4 with a half-drafted body or unpicked labels.

**Suggested fix.** Add "Done when repository, type, labels, title and body are all decided." under the heading.

— Reviewed by Claude Opus 5

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
+---
+name: create-issue
```
tom2drum, line 2: **nit** · needs-human — Two linked changes to how this skill is reached:

1. The deleted skill set `disable-model-invocation: true`; the new frontmatter drops it, buying permanent context load. No skill in `.agents/` invokes `create-issue`. `SKILL-MECHANICS.md`: "Pick model-invocation only when the agent must reach the skill on its own, or another skill must. If it only ever fires by hand, make it user-invoked and pay no context load." Precedent is mixed — `create-pr` and `to-spec` are model-invoked.
2. Given it is now always loaded, the description spends four trigger names on one branch — "…from a discussion, Slack link, meeting notes, or a document." Step 1 treats discussion / meeting notes / document identically (one branch: "The user named or linked a source"); only Slack has a distinct path. "file or create" are synonyms.

**Suggested fix.** Confirm the invocation-mode switch is intended. If not, restore `disable-model-invocation: true`. Either way, collapse the description to the two real branches — e.g. "Use when the user asks to file an issue from a source they name (Slack thread, notes, document) or from this conversation."

— Reviewed by Claude Opus 5

tom2drum (commented):
# Review — source-agnostic `create-issue` skill

| | |
| --- | --- |
| Reviewed | `fe1b07122` → `7aaded0c` |
| Round | 1 of 3 |
| Findings | 1 blocker · 3 major · 5 nit |
| By axis | Spec — (skipped) · Standards 5 · Correctness 4 |
| Outcome | `needs-human` |

**Spec axis skipped** — this change has no task spec (no `.agents/tasks/` folder for this branch), so it ran on two axes. Standards used `prose-smells.md`, the instruction-surface baseline, since the diff is entirely under `.agents/`.

Two findings are marked **needs-human** — they turn on intent the code cannot settle: whether making this skill model-invokable was deliberate, and whether project numbers 6 and 8 are the right boards.

— Reviewed by Claude Opus 5

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+  --label "<label>"
+```
+
+Omit `--label` when there are none. Repeat `--label` for each label. Always pass `--type`. Write the body to a temp file so shell escaping cannot mangle it.
+
+If the repo is in the table, add the new issue to the specified project board:
+
+| Repo | Owner | Project number |
+|---|---|---|
+| `blockscout/frontend` | `blockscout` | 6 |
```
tom2drum, line 92: **nit** · needs-human — Project numbers `6` (`blockscout/frontend`) and `8` (`blockscout/blockscout`) are hardcoded and could not be verified from here: `gh project list --owner blockscout` fails on the missing `read:project` scope, and `/orgs/blockscout/projects/{6,8}` return 404 unauthenticated, which is indistinguishable from private. A wrong number silently files onto another team's board — and the scope gap in the blocker above means nobody would find out by running it.

**Suggested fix.** Have someone with `read:project` run `gh project list --owner blockscout` and confirm both numbers and their titles.

— Reviewed by Claude Opus 5

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+## Additional Context
+
+[Any other relevant technical information]
+```
+
+**Public-safe.** People unnamed and unattributed; no links to the private source (Slack, Notion, Fireflies, internal docs); no client names; no unreleased dates or roadmap.
+
+## Step 4 — Confirm, create, stop
+
+Present repository, type, title, body, labels (or "None"), and project board (or "None") and wait. Apply requested edits and re-confirm.
```
tom2drum, line 73: **nit** — This line orders the agent to present "project board (or \"None\")" and wait — but the table that defines which repos even have a board sits at lines 88–91, after the create command. The agent is told to confirm a value whose mechanism it has not yet read. Scattering, against `writing-for-agents`: "Co-location … Keep a concept's definition, rules, and caveats under one heading rather than scattered."

**Suggested fix.** Move the board table and the "Unlisted repos skip the board" line above the confirm sentence.

— Reviewed by Claude Opus 5

.agents/skills/create-issue/slack-thread.md
```diff
@@ -0,0 +1,28 @@
...
+- **channel_id** — the segment starting with `C` (e.g. `C04XXXX5DAT`).
+- **message_ts** — take the `p`-prefixed number, drop the `p`, insert a dot before the last 6 digits. `p1709834567890123` → `1709834567.890123`.
+
+If the URL cannot be parsed, ask for `channel_id` and `message_ts`.
+
+## Read the thread
+
+```
+Tool: slack_read_thread
+Server: plugin-slack-slack
```
tom2drum: **nit** — `Server: plugin-slack-slack` pins a Cursor-plugin server name carried over from the deleted skill; it does not match the Slack MCP server actually connected here. Sediment — a line true of a previous design. The tool name `slack_read_thread` is the durable fact; the server name will keep drifting.

**Suggested fix.** Drop the `Server:` line and name only the tool.

— Reviewed by Claude Opus 5

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+name: create-issue
+description: >-
+  Create a GitHub issue from the available context. Use when they ask
+  to file or create an issue from a discussion, Slack link, meeting notes, or
+  a document.
+---
+
+# Create issue
+
+Turn a provided information into a thin, public-safe GitHub issue. Then stop.
```
tom2drum: **nit** — "Turn a provided information into a thin, public-safe GitHub issue." is ungrammatical, as is line 21's "For reading Slack threads use [`slack-thread.md`](slack-thread.md) instruction."

**Suggested fix.** "Turn the provided material into…" and "To read a Slack thread, follow [`slack-thread.md`](slack-thread.md)."

— Reviewed by Claude Opus 5

.agents/skills/create-issue/SKILL.md
tom2drum: Confirm that it is intentional. And agree to collapse the descriptioin.

.agents/skills/create-issue/SKILL.md
tom2drum: Confirm that the IDs are correct.

.agents/skills/create-issue/SKILL.md
tom2drum: Fixed in 664694fab. Before `item-add` the skill now requires a `project` scope in `gh auth status`, and asks for `gh auth refresh -s project` if it is missing.

.agents/skills/create-issue/SKILL.md
tom2drum: Fixed in 664694fab. `--type` is passed only for `blockscout/*`; a type-resolution error retries the same create without `--type`.

.agents/skills/create-issue/SKILL.md
tom2drum: Fixed in 664694fab. The body is drafted from the source Step 1 resolved, not from this conversation.

.agents/skills/create-issue/SKILL.md
tom2drum: Fixed in 664694fab. Step 3 now opens with "Done when repository, type, labels, title and body are all decided."

.agents/skills/create-issue/slack-thread.md
tom2drum: Fixed in 664694fab. Dropped the `Server:` line; only the tool name remains.

.agents/skills/create-issue/SKILL.md
tom2drum: Deferred. The board table stays after create; not this change's work.

.agents/skills/create-issue/SKILL.md
tom2drum: Model-invocation stays — that switch was intentional so "file an issue from this chat" is reachable in ordinary language. Description collapsed to the two real branches (named source vs this conversation) in 664694fab.

.agents/skills/create-issue/SKILL.md
tom2drum: Fixed in 664694fab. "Turn the provided material…" and "To read a Slack thread, follow …".

tom2drum (commented):
# Review — round 2 (arbitration)

| | |
| --- | --- |
| Reviewed | `fe1b07122` → `664694fab` |
| Round | 2 of 3 |
| New findings | 0 blocker · 0 major · 1 nit |
| By axis | Arbitration (fresh agent, ruling on the round-1 record) |
| Outcome | `clear` |

An arbitration round verifies the round-1 fixes rather than re-reviewing the change, so it surfaces no new findings outside the fixes themselves.

## Fixes verified — 7 of 7

| | Finding | Ruling |
| --- | --- | --- |
| F1 | blocker · `project` scope gate | **verified** — line 88 sits before the `item-add` block and "wait until it does" is a hard stop, not advice |
| F2 | major · conditional `--type` | **verified** — line 86 is reachable, and the unconditional block above it reads as a template-plus-caveat, matching how `--label` is already handled |
| F3 | major · "From the source" | **verified** — line 45 now tracks whatever Step 1 resolved |
| F4 | major · Step 3 completion criterion | **verified** — line 33 matches the "Done when …" shape of Steps 1 and 2 and enumerates Step 3's actual outputs |
| F5 | nit · invocation mode + description | **verified** — model-invocation confirmed intentional; description now names exactly Step 1's two branches |
| F8 | nit · stale `Server:` line | **verified** — only `Tool: slack_read_thread` remains |
| F9 | nit · grammar | **verified** |

## F6 — closed, IDs confirmed

This was `needs-human` in round 1 only because the token could not list projects. It can now. `gh project list --owner blockscout` returns **6 = "Front-end tasks"** (230 items) and **8 = "Blockscout Elixir Board"** (1213 items). Both numbers are correct for the repos the table maps them to.

## F7 — deferral stands, reasoning corrected

The reject read "Deferred. The board table stays after create; **not this change's work**." The scope half of that is wrong: the board table was added by `f289b661a`, inside this same PR, so it is squarely in the reviewed diff — and the scattering is real, since line 73 asks the agent to present "project board (or 'None')" seventeen lines before the table that defines it.

The deferral itself survives on a different and legitimate ground: `review-changes` makes `deferred` the default for a `nit`, which is never auto-fixed. Outcome unchanged — just not for the stated reason.

## One regression from the fixes

See the inline comment on line 73. It follows from F2's fix and is a `nit`, so it does not block.

— Reviewed by Claude Opus 5

.agents/skills/create-issue/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+## Additional Context
+
+[Any other relevant technical information]
+```
+
+**Public-safe.** People unnamed and unattributed; no links to the private source (Slack, Notion, Fireflies, internal docs); no client names; no unreleased dates or roadmap.
+
+## Step 4 — Confirm, create, stop
+
+Present repository, type, title, body, labels (or "None"), and project board (or "None") and wait. Apply requested edits and re-confirm.
```
tom2drum, line 73: **nit** · regression from F2's fix — Making `--type` conditional on `blockscout/*` left the surrounding text still treating it as universal:

- line 33 requires "type … decided" for every repo
- line 39 infers one unconditionally
- this line has the agent present "repository, **type**, title, body" for confirmation

So for a non-`blockscout/*` repo the user confirms a type that line 86 then silently never sends. Harmless to the issue itself, but it is the document disagreeing with itself.

**Suggested fix.** Present type only when the repo is `blockscout/*` (or mark it "not applicable"), and qualify line 39's inference the same way.

— Reviewed by Claude Opus 5

Issue #646, opened 2023-02-22
"Parent hash" from block page has broken link


vbaranov:
Steps to reproduce:
- open any block page
- try to open parent block from "Parent hash"
- `/block` page will be opened instead of corresponding block

PR #3641, opened 2026-08-17
Add a skill that turns Slack config requests into exact env vars for DevOps

tom2drum:
## Description

Config requests in Slack carry intent ("turn on X", "new logo"); DevOps needs exact `KEY=value` lines for a named instance or a vague class. This adds `resolve-config-request`, an operational skill that does that conversion and posts the result to the DevOps requests channel.

The general path reads the thread, pins each ask to a documented variable, diffs against live config when the target is nameable, validates the overlay with `envs-validator`, and drafts a Russian DevOps message for approval before send. Designer-owned chrome (logo, favicon, theme, banners) takes a disclosed skin branch with an unconditional demo; fetched JSON configs (chains menu, footer, marketplace, …) get a `frontend-configs` PR and no demo.

Slack thread parsing moves to `.agents/slack-thread.md` so `create-issue` and this skill share one parser. Channel IDs stay in `TEAM.md`.

## Environment variables

None.

## Minimum API version

None.

None.

## Additional information

The grilling brief that produced the skill is left uncommitted under `.agents/handoffs/` — it is a design record, not something the agent loads.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.agents/skills/resolve-config-request/SKILL.md
+++ b/.agents/skills/resolve-config-request/SKILL.md
@@ -24,7 +24,7 @@ No commit without the user's explicit confirmation in this conversation.
 
 ## Scope
 
-**In:** variables in `docs/ENVS.md`, plus start-time variables like `FAVICON_MASTER_URL`.
+**In:** variables in `docs/ENVS.md`, including start-time ones such as `FAVICON_MASTER_URL`.
 
 **Out, with the exit:**
 
@@ -82,12 +82,12 @@ Post the approved message to `blockscout-devops-requests` — resolve the channe
 Every deployed instance exposes `{ envs: { …all NEXT_PUBLIC_* } }` at `<url>/node-api/config`. The CLI:
 
 ```bash
-tools/dev-server/fetch.sh <alias> --out=/tmp/instance.env
+tools/dev-server/fetch.sh <alias> --omit-local-envs --out=/tmp/instance.env
 ```
 
 Aliases live in `tools/dev-server/registry.json`. How fetch and env layering work: `tools/dev-server/CONTEXT.md`.
 
-Limits: registry alias only (otherwise GET `/node-api/config` yourself); `/node-api/config` is `NEXT_PUBLIC_*` only, so start-time variables like `FAVICON_MASTER_URL` cannot be checked this way.
+Limits: registry alias only (otherwise GET `/node-api/config` yourself); `/node-api/config` is `NEXT_PUBLIC_*` only, so start-time variables like `FAVICON_MASTER_URL` cannot be checked this way; `fetch.sh` drops `ignoredEnvs` and `deprecatedEnvs` — for those, GET `/node-api/config`.
 
 ## Validation
 
@@ -95,16 +95,21 @@ Limits: registry alias only (otherwise GET `/node-api/config` yourself); `/node-
 
 Overlay our change on the fetched env. dotenv-cli: the **first** `-e` file wins (`tools/dev-server/CONTEXT.md`). For a key to drop, strip it from the fetched file before validating.
 
+From `deploy/tools/envs-validator/`, the preamble in `test.sh` (collect placeholders, copy `test/assets`, build), then:
+
 ```bash
-pnpm --filter envs-validator build
-dotenv -e /tmp/change.env -e /tmp/instance.env -- pnpm --filter envs-validator validate
+pnpm exec dotenv -e /tmp/change.env -e /tmp/instance.env -- pnpm run validate
 ```
 
 Schema organisation: `deploy/tools/envs-validator/CONTEXT.md`. This is what catches missing companion variables (`.when(...)`), forbidden combinations, and malformed nested JSON — a documentation check cannot.
 
 ## Fetched configs
 
-JSON the instance **fetches** at startup — not instance chrome, no demo. Work in the `frontend-configs` checkout. Follow the `check-github-cli` skill. Confirm with the user before the commit and the PR. The `create-pr` skill's frontend template, ENVs label, and issue-from-branch steps do not apply.
+JSON the instance **fetches** at startup — not instance chrome, no demo.
+
+Checkout: a workspace folder named `frontend-configs` or `blockscout_frontend_configs`. If none, stop and ask the user to add it.
+
+Follow the `check-github-cli` skill. Confirm with the user before the commit and the PR. The `create-pr` skill's frontend template, ENVs label, and issue-from-branch steps do not apply.
 
 | Directory | Variable |
 | --- | --- |
--- a/.agents/skills/resolve-config-request/SKILL.md
+++ b/.agents/skills/resolve-config-request/SKILL.md
@@ -1,0 +1,141 @@
+---
+name: resolve-config-request
+description: Turn a Slack request into the exact env vars to set on a live instance, and hand them to DevOps.
+disable-model-invocation: true
+---
+
+# Resolve a config request
+
+Convert **intent → variables**. The request already carries the intent; this skill never originates it.
+
+## Sure
+
+Only send a DevOps message we are **sure** of. Every other rule serves that.
+
+- **Stop** on ambiguity. If the variable cannot be pinned, or the value is not in the documented set of allowed values, stop and tell the user.
+- **The user is the face.** Uncertainty and clarifying questions go to the user. Requester-facing posts are only the handover reply after an approved send, and the demo link on the skin branch.
+- **Intent is given.** Any question that turns on what someone *wants* is a stop, not a judgement call.
+- **Operator.** `docs/ENVS.md` is the manual. The docs pin legal values, not the consumers. A value the docs cannot pin is a documentation gap, fixed as documentation.
+- **Off.** Prefer a documented off-value over unsetting. Unset only when the docs show the default *is* the desired state.
+- **Mirror** the requester's targeting language. Unstated is not undeterminable — the DevOps bot resolves instances.
+- **Drift.** `NEXT_PUBLIC_HOMEPAGE_HERO_BANNER_CONFIG` and `NEXT_PUBLIC_COLOR_THEME_OVERRIDES` ship with a `frontend-configs` file change in the same run; take the skin branch.
+
+No commit without the user's explicit confirmation in this conversation.
+
+## Scope
+
+**In:** variables in `docs/ENVS.md`, plus start-time variables like `FAVICON_MASTER_URL`.
+
+**Out, with the exit:**
+
+- **Not configurable** — needs a code change. Stop and offer the `create-issue` skill. The channel carries these; they look like ordinary requests.
+- **Retired variable named** — read `docs/DEPRECATED_ENVS.md` and propose the replacement.
+- **Outside the frontend, or mixed with a non-configurable ask** — do the configurable frontend part only, and tell the user what was left out.
+- **CDN-only trees** in `frontend-configs` with no `docs/ENVS.md` variable (`multisearch/`, `token-icons/`, `nft-images/`, `explorer-logos/`, `meta-suites-logos/`) — stop and tell the user.
+
+## Steps
+
+### 1. Read the thread
+
+Follow `.agents/slack-thread.md`. Threads are typically in Russian.
+
+**Done when:** every message and attachment in the thread is in hand.
+
+### 2. Map the ask to variables
+
+Using `docs/ENVS.md` as the operator manual, pin each change to a variable and a documented value. **Stop** if it cannot be pinned.
+
+**Skin.** Read [`SKIN-REQUESTS.md`](SKIN-REQUESTS.md) when a designer owns the change: this instance's logo, icon, favicon, OG image, colour theme, hero banner, homepage highlights, or navigation promo banner. That branch runs to its own stop; it returns here at step 3. A chains-menu or footer icon that merely *lives* in `network-icons/` / `footer-icons/` is not that.
+
+**Fetched config.** A chains menu, footer, marketplace, widgets, or cross-chain JSON — and icons those files reference — is a `frontend-configs` PR without a demo. See **Fetched configs** below; do that PR before step 3. Skip when the skin branch is already running — those files ride in that PR.
+
+**Done when:** every asked change is a `(variable, value)` pair pinned by the docs (and any non-skin configs PR is merged to `main`), or the run has stopped.
+
+### 3. Read current state (nameable instance)
+
+When the target is a nameable instance, fetch its live config, diff against the proposed change, drop no-ops, and identify anything that needs removing. Sending a variable already at that value is how these requests lose credibility.
+
+When the target is a vague class ("everywhere", "all testnets"), there is no single current state — go set-only.
+
+**Done when:** the message contains only real changes, or the target is a class and the message is set-only.
+
+### 4. Validate
+
+Run the startup validator against the live env plus our change, before drafting. See **Validation** below.
+
+**Done when:** the validator accepts the overlay (or the only remaining vars are start-time ones it cannot see).
+
+### 5. Draft the DevOps message
+
+Compose it per **The DevOps message** below. Show it to the user and wait.
+
+**Done when:** the user has approved the exact text.
+
+### 6. Send, then hand over
+
+Post the approved message to `blockscout-devops-requests` — resolve the channel ID from `.agents/TEAM.md`. Then reply in the original thread — same language as the thread — with a link to that message, saying the request has been handed over and is waiting on DevOps.
+
+**Done when:** both posts exist and the original-thread reply carries the DevOps link.
+
+## Reading an instance's env
+
+Every deployed instance exposes `{ envs: { …all NEXT_PUBLIC_* } }` at `<url>/node-api/config`. The CLI:
+
+```bash
+tools/dev-server/fetch.sh <alias> --out=/tmp/instance.env
+```
+
+Aliases live in `tools/dev-server/registry.json`. How fetch and env layering work: `tools/dev-server/CONTEXT.md`.
+
+Limits: registry alias only (otherwise GET `/node-api/config` yourself); `/node-api/config` is `NEXT_PUBLIC_*` only, so start-time variables like `FAVICON_MASTER_URL` cannot be checked this way.
+
+## Validation
+
+**Default alias: `eth`.** One representative instance is enough, including when the target is a vague class.
+
+Overlay our change on the fetched env. dotenv-cli: the **first** `-e` file wins (`tools/dev-server/CONTEXT.md`). For a key to drop, strip it from the fetched file before validating.
+
+```bash
+pnpm --filter envs-validator build
+dotenv -e /tmp/change.env -e /tmp/instance.env -- pnpm --filter envs-validator validate
+```
+
+Schema organisation: `deploy/tools/envs-validator/CONTEXT.md`. This is what catches missing companion variables (`.when(...)`), forbidden combinations, and malformed nested JSON — a documentation check cannot.
+
+## Fetched configs
+
+JSON the instance **fetches** at startup — not instance chrome, no demo. Work in the `frontend-configs` checkout. Follow the `check-github-cli` skill. Confirm with the user before the commit and the PR. The `create-pr` skill's frontend template, ENVs label, and issue-from-branch steps do not apply.
+
+| Directory | Variable |
+| --- | --- |
+| `configs/featured-networks/` | `NEXT_PUBLIC_FEATURED_NETWORKS` |
+| `configs/footer-links/` (+ `configs/footer-icons/`) | `NEXT_PUBLIC_FOOTER_LINKS` |
+| `configs/marketplace/` and siblings (`marketplace-categories/`, `marketplace-subgraph-links/`, `marketplace-logos/`, `marketplace-security-reports/`) | `NEXT_PUBLIC_MARKETPLACE_CONFIG_URL`, `NEXT_PUBLIC_MARKETPLACE_CATEGORIES_URL`, `NEXT_PUBLIC_MARKETPLACE_GRAPH_LINKS_URL` |
+| `configs/widgets/` | `NEXT_PUBLIC_ADDRESS_3RD_PARTY_WIDGETS_CONFIG_URL` |
+| `configs/cross-chain/` | `NEXT_PUBLIC_ZETACHAIN_SERVICE_CHAINS_CONFIG_URL` |
+
+Hosted icons for **inlined** lists (the env holds the JSON; the file is the URL inside it): `configs/ide-icons/` → `NEXT_PUBLIC_CONTRACT_CODE_IDES`; `configs/nft-marketplace-logos/` → `NEXT_PUBLIC_VIEWS_NFT_MARKETPLACES`; `configs/multichain-balance/` → `NEXT_PUBLIC_MULTICHAIN_BALANCE_PROVIDER_CONFIG`.
+
+After merge to `main`, confirm each raw URL returns 200. If the instance already has that URL, the DevOps ask is a restart to re-fetch — no new `KEY=value`. If the URL is new, or the value is inlined, it goes in the block as usual.
+
+## The DevOps message
+
+Russian, informal. Target in the requester's own words — not registry aliases. Sets in one fenced `KEY=value` block; keys to drop as names only. Mention the requester with `<@U…>` from the thread author's `user_id` (`.agents/TEAM.md` → How to address).
+
+````
+Привет! 🐈
+Для eth.blockscout.com нужно поставить:
+```
+NEXT_PUBLIC_NAVIGATION_HIGHLIGHTED_ROUTES=['/accounts']
+```
+cc <@U024DUPJG3A>
+````
+
+Unset-only: `нужно убрать:` + names in the fence, no `=value`. Mixed: both sentences, set first. Restart-only (fetched file, URL already on the instance): `нужно перезапустить фронт, чтобы подтянуть обновлённый конфиг.` — no fence.
+
+Omit: that we validated; a backlink to the source thread. A set or drop fence already implies the restart. Caveats as bullets after the fence, only when there is a real one (don't-remove-X, companion variables).
+
+## Value formatting
+
+- **`rgba()`, never hex** — a `#` in an env var breaks bash.
+- **JSON values use single quotes** (`'{"a":1}'`) so they paste into a shell or `.env` file. The validator's `replaceQuotes` converts them.
--- a/.agents/skills/resolve-config-request/SKILL.md
+++ b/.agents/skills/resolve-config-request/SKILL.md
@@ -85,7 +85,7 @@ Every deployed instance exposes `{ envs: { …all NEXT_PUBLIC_* } }` at `<url>/n
 tools/dev-server/fetch.sh <alias> --omit-local-envs --out=/tmp/instance.env
 ```
 
-Aliases live in `tools/dev-server/registry.json`. How fetch and env layering work: `tools/dev-server/CONTEXT.md`.
+Step 3 diffs this file. Aliases live in `tools/dev-server/registry.json`. How fetch and env layering work: `tools/dev-server/CONTEXT.md`.
 
 Limits: registry alias only (otherwise GET `/node-api/config` yourself); `/node-api/config` is `NEXT_PUBLIC_*` only, so start-time variables like `FAVICON_MASTER_URL` cannot be checked this way; `fetch.sh` drops `ignoredEnvs` and `deprecatedEnvs` — for those, GET `/node-api/config`.
 
@@ -93,12 +93,17 @@ Limits: registry alias only (otherwise GET `/node-api/config` yourself); `/node-
 
 **Default alias: `eth`.** One representative instance is enough, including when the target is a vague class.
 
-Overlay our change on the fetched env. dotenv-cli: the **first** `-e` file wins (`tools/dev-server/CONTEXT.md`). For a key to drop, strip it from the fetched file before validating.
+`--omit-local-envs` drops `localEnvs` (`tools/dev-server/envs-rules.json`); the schema requires `NEXT_PUBLIC_APP_HOST`. Copy the fetched file and add those keys, then overlay our change on the copy. dotenv-cli: the **first** `-e` file wins (`tools/dev-server/CONTEXT.md`). For a key to drop, strip it from the copy before validating.
+
+```bash
+cp /tmp/instance.env /tmp/instance.validate.env
+jq -r '.localEnvs | to_entries[] | "\(.key)=\(.value)"' tools/dev-server/envs-rules.json >> /tmp/instance.validate.env
+```
 
 From `deploy/tools/envs-validator/`, the preamble in `test.sh` (collect placeholders, copy `test/assets`, build), then:
 
 ```bash
-pnpm exec dotenv -e /tmp/change.env -e /tmp/instance.env -- pnpm run validate
+pnpm exec dotenv -e /tmp/change.env -e /tmp/instance.validate.env -- pnpm run validate
 ```
 
 Schema organisation: `deploy/tools/envs-validator/CONTEXT.md`. This is what catches missing companion variables (`.when(...)`), forbidden combinations, and malformed nested JSON — a documentation check cannot.
--- a/.agents/skills/resolve-config-request/SKIN-REQUESTS.md
+++ b/.agents/skills/resolve-config-request/SKIN-REQUESTS.md
@@ -1,0 +1,81 @@
+# Skin requests
+
+The branch taken when a designer owns the change: this instance's logo, icon, favicon, OG image, colour theme, hero banner, homepage highlights, or navigation promo banner. Screenshots do that badly and a live **demo** does it well. The demo is unconditional — even a plain logo swap earns one (that is where dark-mode variants and viewBox cropping go wrong).
+
+Variables: `NEXT_PUBLIC_NETWORK_LOGO` / `_DARK`, `NEXT_PUBLIC_NETWORK_ICON` / `_DARK`, `FAVICON_MASTER_URL`, `NEXT_PUBLIC_OG_IMAGE_URL`, `NEXT_PUBLIC_COLOR_THEME_OVERRIDES`, `NEXT_PUBLIC_HOMEPAGE_HERO_BANNER_CONFIG`, `NEXT_PUBLIC_HOMEPAGE_HIGHLIGHTS_CONFIG`, `NEXT_PUBLIC_NAVIGATION_PROMO_BANNER_CONFIG`.
+
+Work in the workspace's `frontend-configs` checkout. Follow the `check-github-cli` skill before any `gh` step.
+
+Non-skin variables on the same request wait and ride in the phase-2 DevOps message. Non-skin `frontend-configs` files on the same request go in this PR.
+
+## Drift
+
+The `configs/hero-banner/` and `configs/color-themes/` JSON files are the editable source of truth for variables whose values are **inlined** into the env var. They are not fetched at runtime. To change one colour, read the current file, patch it, and regenerate the string.
+
+Emitting `NEXT_PUBLIC_HOMEPAGE_HERO_BANNER_CONFIG` or `NEXT_PUBLIC_COLOR_THEME_OVERRIDES` without updating the file in the same change lets the two drift: the next colour tweak regenerates from a stale base and silently reverts everything since. Those variables always take this branch, even for a one-colour change.
+
+Regenerate with the configs repo's converter:
+
+```bash
+node ./tools/json-converter/index.js <path-to-file>
+```
+
+## Colour sources
+
+Two configs, overlapping vocabulary; the request usually covers only one of them in text. Reconcile both before writing anything.
+
+- **Message text → hero banner.** background, text colour, button default/hover, light and dark.
+- **Figma sheet → colour theme overrides.** These tokens typically exist *only* in Figma.
+
+A `background` in the message text is not the theme's `bg.primary`.
+
+Read fills off a rendered frame (screenshot). Bound-variable lookup returns design-system defaults for per-instance raw fills. To tell which tokens are actually part of the request, diff each swatch against `DEFAULT_THEME_COLORS` in `src/toolkit/theme/foundations/colors.ts`: a token sitting at the default is not part of the request; one that differs is.
+
+**If Figma is unreachable, stop.** A half-applied skin reaching DevOps is worse than a blocked request.
+
+Hero banner values are `Array<string | undefined>` with index `[0]` = light, `[1]` = dark (`[1] || [0]` fallback).
+
+## Assets
+
+Optimise SVGs with the configs repo's `svgo.config.cjs` — it strips dimensions, keeps `viewBox`, and applies `prefixIds` per file so light and dark variants do not collide. Favicon: square PNG, ≥180×180. OG image: 1200×600.
+
+Directories: `configs/network-logos/`, `configs/network-icons/`, `configs/favicons/`, `configs/og-images/`, `configs/hero-banner/`, `configs/network-skins/` (hero backgrounds), `configs/color-themes/`, `configs/homepage-highlights/`. Navigation promo banner is inlined and has no file in this repo.
+
+While the configs PR is open, asset URLs are the **PR branch's** raw GitHub URLs, not `main`.
+
+## Demo mechanics
+
+`.env.extra` is committed and is the **only** channel that carries env overrides into a review deploy. An instance not in the registry also needs a temporary `tools/dev-server/registry.json` entry. Both are committed files in the **frontend** repo and must be reverted in phase 2. Env layering and the preset add/remove procedure: `tools/dev-server/CONTEXT.md`.
+
+Hostname: `review-<branch-slug>.k8s-dev.blockscout.com`. Follow the `deploy-demo` skill.
+
+Verify the theme by **computed values** — read the resolved CSS custom properties in both light and dark. The first click after opening the theme settings pop-over is swallowed; click again. A demo also confirms two things a local run cannot: that the favicon bundle is generated at container start from `FAVICON_MASTER_URL`, and that the envs-validator accepts the inline JSON blobs.
+
+## Phase 1 — demo
+
+No approval gate on the branch, the configs PR, the demo, or the message that carries the demo link: the PR is open but unmerged, so nothing can break a production instance. The requester (designer) is the gatekeeper for phase 2; the user relays that approval. There is no automated watch on the thread — the user monitors and continues this session.
+
+1. Produce the assets and JSON configs from the thread and Figma.
+2. Open a PR on `frontend-configs` (target per that repo's README). Confirm with the user before the commit and the PR. The `create-pr` skill's frontend template, ENVs label, and issue-from-branch steps do not apply.
+3. Point the demo at the PR-branch raw URLs (see **Demo mechanics**). Deploy it.
+4. Verify on the live demo, then post the demo link to the requester. **Stop.**
+
+**Done when:** the configs PR is open, the demo is live, the requester has the link, and this run has stopped.
+
+## Phase 2 — ship
+
+Starts only when the user relays the designer's approval.
+
+1. Merge the configs PR. Swap every branch URL to `main` and confirm each returns 200 — that is when instances can fetch them.
+2. For inlined vars (hero banner, colour theme), regenerate the strings from the merged files (the converter above). URL vars (homepage highlights, asset URLs): the env value is the `main` raw URL. Navigation promo: the inlined string produced in phase 1.
+3. Run the parent skill from **Read current state** through **Send, then hand over**.
+4. **Teardown** — checkable, they fail silently when skipped:
+   - temporary frontend branch deleted, local and remote
+   - `tools/dev-server/registry.json` reverted
+   - `.env.extra` reverted
+   - demo destroyed (`.github/workflows/cleanup.yml`; hostname returns 404)
+   - `git status` clean
+
+The review-image cleanup is broken ([frontend#3638](https://github.com/blockscout/frontend/issues/3638)); hostname 404 is the bar, not image deletion.
+
+**Done when:** the DevOps message is sent, the handover reply is posted, and every teardown box above is true.
--- a/.agents/skills/resolve-config-request/SKIN-REQUESTS.md
+++ b/.agents/skills/resolve-config-request/SKIN-REQUESTS.md
@@ -29,9 +29,9 @@ Two configs, overlapping vocabulary; the request usually covers only one of them
 
 A `background` in the message text is not the theme's `bg.primary`.
 
-Read fills off a rendered frame (`get_screenshot`). Bound-variable lookup (`get_variable_defs`) returns design-system defaults for per-instance raw fills. To tell which tokens are actually part of the request, diff each swatch against `DEFAULT_THEME_COLORS` in `src/toolkit/theme/foundations/colors.ts`: a token sitting at the default is not part of the request; one that differs is.
+Read fills off a rendered frame (`get_screenshot`). Bound-variable lookup (`get_variable_defs`) returns design-system defaults for per-instance raw fills. Both need the Figma plugin in Claude Code or Cursor. To tell which tokens are actually part of the request, diff each swatch against `DEFAULT_THEME_COLORS` in `src/toolkit/theme/foundations/colors.ts`: a token sitting at the default is not part of the request; one that differs is.
 
-**If `get_screenshot` / `get_variable_defs` fail, or Figma is unreachable, stop.** A half-applied skin reaching DevOps is worse than a blocked request.
+**If those tools are absent or fail, or Figma is unreachable, stop.** A half-applied skin reaching DevOps is worse than a blocked request.
 
 Hero banner values are `Array<string | undefined>` with index `[0]` = light, `[1]` = dark (`[1] || [0]` fallback).
 
--- a/.agents/skills/resolve-config-request/SKIN-REQUESTS.md
+++ b/.agents/skills/resolve-config-request/SKIN-REQUESTS.md
@@ -1,10 +1,10 @@
 # Skin requests
 
-The branch taken when a designer owns the change: this instance's logo, icon, favicon, OG image, colour theme, hero banner, homepage highlights, or navigation promo banner. Screenshots do that badly and a live **demo** does it well. The demo is unconditional — even a plain logo swap earns one (that is where dark-mode variants and viewBox cropping go wrong).
+A live **demo** is the review surface — screenshots miss dark-mode variants and viewBox cropping. The demo is unconditional — even a plain logo swap earns one.
 
 Variables: `NEXT_PUBLIC_NETWORK_LOGO` / `_DARK`, `NEXT_PUBLIC_NETWORK_ICON` / `_DARK`, `FAVICON_MASTER_URL`, `NEXT_PUBLIC_OG_IMAGE_URL`, `NEXT_PUBLIC_COLOR_THEME_OVERRIDES`, `NEXT_PUBLIC_HOMEPAGE_HERO_BANNER_CONFIG`, `NEXT_PUBLIC_HOMEPAGE_HIGHLIGHTS_CONFIG`, `NEXT_PUBLIC_NAVIGATION_PROMO_BANNER_CONFIG`.
 
-Work in the workspace's `frontend-configs` checkout. Follow the `check-github-cli` skill before any `gh` step.
+Checkout and commit/PR rules: parent **Fetched configs**. Follow the `check-github-cli` skill before any `gh` step.
 
 Non-skin variables on the same request wait and ride in the phase-2 DevOps message. Non-skin `frontend-configs` files on the same request go in this PR.
 
@@ -29,9 +29,9 @@ Two configs, overlapping vocabulary; the request usually covers only one of them
 
 A `background` in the message text is not the theme's `bg.primary`.
 
-Read fills off a rendered frame (screenshot). Bound-variable lookup returns design-system defaults for per-instance raw fills. To tell which tokens are actually part of the request, diff each swatch against `DEFAULT_THEME_COLORS` in `src/toolkit/theme/foundations/colors.ts`: a token sitting at the default is not part of the request; one that differs is.
+Read fills off a rendered frame (`get_screenshot`). Bound-variable lookup (`get_variable_defs`) returns design-system defaults for per-instance raw fills. To tell which tokens are actually part of the request, diff each swatch against `DEFAULT_THEME_COLORS` in `src/toolkit/theme/foundations/colors.ts`: a token sitting at the default is not part of the request; one that differs is.
 
-**If Figma is unreachable, stop.** A half-applied skin reaching DevOps is worse than a blocked request.
+**If `get_screenshot` / `get_variable_defs` fail, or Figma is unreachable, stop.** A half-applied skin reaching DevOps is worse than a blocked request.
 
 Hero banner values are `Array<string | undefined>` with index `[0]` = light, `[1]` = dark (`[1] || [0]` fallback).
 
@@ -45,7 +45,7 @@ While the configs PR is open, asset URLs are the **PR branch's** raw GitHub URLs
 
 ## Demo mechanics
 
-`.env.extra` is committed and is the **only** channel that carries env overrides into a review deploy. An instance not in the registry also needs a temporary `tools/dev-server/registry.json` entry. Both are committed files in the **frontend** repo and must be reverted in phase 2. Env layering and the preset add/remove procedure: `tools/dev-server/CONTEXT.md`.
+`.env.extra` is committed and is the **only** channel that carries env overrides into a review deploy. An instance not in the registry: add it to `tools/dev-server/registry.json`, run `pnpm presets:sync` (writes `.github/workflows/deploy-review.yml` and `.vscode/tasks.json`), commit all four **on the demo branch**. They die with that branch. Procedure: `tools/dev-server/CONTEXT.md`.
 
 Hostname: `review-<branch-slug>.k8s-dev.blockscout.com`. Follow the `deploy-demo` skill.
 
@@ -53,10 +53,10 @@ Verify the theme by **computed values** — read the resolved CSS custom propert
 
 ## Phase 1 — demo
 
-No approval gate on the branch, the configs PR, the demo, or the message that carries the demo link: the PR is open but unmerged, so nothing can break a production instance. The requester (designer) is the gatekeeper for phase 2; the user relays that approval. There is no automated watch on the thread — the user monitors and continues this session.
+No **designer** gate on the branch, the configs PR, the demo, or the demo-link message — the PR is open but unmerged, so nothing can break a production instance. The user's commit confirmation still applies. The requester (designer) is the gatekeeper for phase 2; the user relays that approval. There is no automated watch on the thread — the user monitors and continues this session.
 
 1. Produce the assets and JSON configs from the thread and Figma.
-2. Open a PR on `frontend-configs` (target per that repo's README). Confirm with the user before the commit and the PR. The `create-pr` skill's frontend template, ENVs label, and issue-from-branch steps do not apply.
+2. Open a PR on `frontend-configs` (target `main` branch). Commit/PR confirmation and `create-pr` exceptions: parent **Fetched configs**.
 3. Point the demo at the PR-branch raw URLs (see **Demo mechanics**). Deploy it.
 4. Verify on the live demo, then post the demo link to the requester. **Stop.**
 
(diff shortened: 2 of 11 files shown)


.agents/skills/resolve-config-request/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Validation
+
+**Default alias: `eth`.** One representative instance is enough, including when the target is a vague class.
+
+Overlay our change on the fetched env. dotenv-cli: the **first** `-e` file wins (`tools/dev-server/CONTEXT.md`). For a key to drop, strip it from the fetched file before validating.
+
+```bash
+pnpm --filter envs-validator build
+dotenv -e /tmp/change.env -e /tmp/instance.env -- pnpm --filter envs-validator validate
```
tom2drum: **blocker** — The validation recipe cannot run in a checkout, so step 4 (`Run the startup validator … before drafting`) is a gate the agent cannot pass. `pnpm --filter envs-validator build` + `validate` exits 1 at `checkPlaceholdersCongruity` with `🚨 Unable to read placeholders file` — `deploy/tools/envs-validator/.env.registry` and `.env` are generated and gitignored. Once those are generated, any of the 8 `envsWithJsonConfig` vars present in a fetched env (every instance sets `NEXT_PUBLIC_FEATURED_NETWORKS`) then fails on a missing `public/assets/configs/featured_networks.json`. Both verified empirically. Separately, `dotenv` is not on `PATH` — it lives only in `node_modules/.bin`.

**Fix.** Mirror `deploy/tools/envs-validator/test.sh`: from the tool dir, run `../../scripts/collect_envs.sh ../../../docs/ENVS.md` and `cp -r test/assets ./public/` first, and invoke it as `./node_modules/.bin/dotenv …`.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
```diff
@@ -0,0 +1,81 @@
...
+
+Optimise SVGs with the configs repo's `svgo.config.cjs` — it strips dimensions, keeps `viewBox`, and applies `prefixIds` per file so light and dark variants do not collide. Favicon: square PNG, ≥180×180. OG image: 1200×600.
+
+Directories: `configs/network-logos/`, `configs/network-icons/`, `configs/favicons/`, `configs/og-images/`, `configs/hero-banner/`, `configs/network-skins/` (hero backgrounds), `configs/color-themes/`, `configs/homepage-highlights/`. Navigation promo banner is inlined and has no file in this repo.
+
+While the configs PR is open, asset URLs are the **PR branch's** raw GitHub URLs, not `main`.
+
+## Demo mechanics
+
+`.env.extra` is committed and is the **only** channel that carries env overrides into a review deploy. An instance not in the registry also needs a temporary `tools/dev-server/registry.json` entry. Both are committed files in the **frontend** repo and must be reverted in phase 2. Env layering and the preset add/remove procedure: `tools/dev-server/CONTEXT.md`.
```
tom2drum: **major** — "An instance not in the registry also needs a temporary `tools/dev-server/registry.json` entry. Both are committed files" understates what the demo needs. The review deploy's `envs_preset` is a **generated enum** in `.github/workflows/deploy-review.yml` (`# presets:start — generated from tools/dev-server/registry.json (run \`pnpm presets:sync\`)`). Without running `pnpm presets:sync` and committing that file, the new alias is not dispatchable — and the phase-2 teardown list never reverts it, so the enum entry leaks into `main`.

**Fix.** Add `pnpm presets:sync` to phase 1 step 3, and add `.github/workflows/deploy-review.yml` reverted to the teardown checklist.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
```diff
@@ -0,0 +1,81 @@
...
+Starts only when the user relays the designer's approval.
+
+1. Merge the configs PR. Swap every branch URL to `main` and confirm each returns 200 — that is when instances can fetch them.
+2. For inlined vars (hero banner, colour theme), regenerate the strings from the merged files (the converter above). URL vars (homepage highlights, asset URLs): the env value is the `main` raw URL. Navigation promo: the inlined string produced in phase 1.
+3. Run the parent skill from **Read current state** through **Send, then hand over**.
+4. **Teardown** — checkable, they fail silently when skipped:
+   - temporary frontend branch deleted, local and remote
+   - `tools/dev-server/registry.json` reverted
+   - `.env.extra` reverted
+   - demo destroyed (`.github/workflows/cleanup.yml`; hostname returns 404)
```
tom2drum: **major** — The teardown order strands the demo. It puts "temporary frontend branch deleted, local and remote" before "demo destroyed (`.github/workflows/cleanup.yml`; hostname returns 404)", but that workflow only fires on `pull_request` closed/merged — and phase 1 opens no *frontend* PR — or on `workflow_dispatch`, whose `appName: review-$GITHUB_REF_NAME_SLUG` requires dispatching **on the branch ref**. Delete the branch first and there is no ref left to dispatch on, so the checklist's last box can never go true.

**Fix.** Reorder: dispatch `cleanup.yml` on the branch (`gh workflow run cleanup.yml --ref <branch>`), confirm the 404, *then* delete the branch.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
```diff
@@ -0,0 +1,81 @@
...
+
+Hostname: `review-<branch-slug>.k8s-dev.blockscout.com`. Follow the `deploy-demo` skill.
+
+Verify the theme by **computed values** — read the resolved CSS custom properties in both light and dark. The first click after opening the theme settings pop-over is swallowed; click again. A demo also confirms two things a local run cannot: that the favicon bundle is generated at container start from `FAVICON_MASTER_URL`, and that the envs-validator accepts the inline JSON blobs.
+
+## Phase 1 — demo
+
+No approval gate on the branch, the configs PR, the demo, or the message that carries the demo link: the PR is open but unmerged, so nothing can break a production instance. The requester (designer) is the gatekeeper for phase 2; the user relays that approval. There is no automated watch on the thread — the user monitors and continues this session.
```
tom2drum: **major** — "No approval gate on the branch, the configs PR, the demo, or the message that carries the demo link" contradicts line 59 ("Confirm with the user before the commit and the PR") and `SKILL.md:23` ("No commit without the user's explicit confirmation in this conversation"). An agent reading this section first will commit and open the PR unprompted.

**Fix.** Scope this sentence to what it means — no *designer* approval gate — and leave the commit/PR confirmation standing.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+Post the approved message to `blockscout-devops-requests` — resolve the channel ID from `.agents/TEAM.md`. Then reply in the original thread — same language as the thread — with a link to that message, saying the request has been handed over and is waiting on DevOps.
+
+**Done when:** both posts exist and the original-thread reply carries the DevOps link.
+
+## Reading an instance's env
+
+Every deployed instance exposes `{ envs: { …all NEXT_PUBLIC_* } }` at `<url>/node-api/config`. The CLI:
+
+```bash
+tools/dev-server/fetch.sh <alias> --out=/tmp/instance.env
```
tom2drum: **major** — `tools/dev-server/fetch.sh <alias> --out=…` does not return the instance's real env. Per `tools/dev-server/CONTEXT.md`, `fetch.ts` "drops `ignoredEnvs` + `deprecatedEnvs`, applies/omits `localEnvs`" — so without `--omit-local-envs` it substitutes dev values (`NEXT_PUBLIC_APP_HOST=localhost`, `_PORT`, `_ENV`, …), and it drops deprecated keys entirely. Step 3's "fetch its live config, diff against the proposed change, drop no-ops" therefore misreports those keys as unset — including exactly the case Scope's "Retired variable named" branch exists to handle.

**Fix.** Add `--omit-local-envs`, and note that dropped/deprecated keys must be read from `GET /node-api/config` directly.

— Reviewed by Claude Opus 5

.agents/slack-thread.md
```diff
@@ -1,11 +1,9 @@
-# Slack thread pointer
+# Read a Slack thread
 
-Parse a Slack thread URL and read it with the Slack MCP. If the tool is missing, unauthenticated, or errors, return to the parent skill's paste fallback — Slack is not a hard gate.
+Parse a Slack thread URL and read the full conversation, including replies.
```
tom2drum: **major** — The move out of `create-issue/` dropped the failure branch. The original opened with "If the tool is missing, unauthenticated, or errors, return to the parent skill's paste fallback — Slack is not a hard gate." `create-issue` still carries its own paste fallback, but `resolve-config-request` SKILL.md:40 ("Follow `.agents/slack-thread.md`") has none — so a Slack outage has no defined exit in a skill where every other branch names one, and the run dead-ends on a step-1 `Done when` it cannot satisfy.

**Fix.** Restore the one-line failure branch here, or give the exit at `SKILL.md:40`.

— Reviewed by Claude Opus 5

.agents/slack-thread.md
```diff
@@ -25,3 +23,7 @@ Arguments:
 ```
 
 If the thread has more than 200 messages, paginate with `cursor` until the full conversation is read.
+
+Attachments often carry the actual content. The Slack connector reports metadata only — download each file.
```
tom2drum: **major** — Rule without a mechanism: "The Slack connector reports metadata only — download each file." No tool, script, or command for that download is named anywhere on the repo's instruction surface, while `slack_read_thread` two lines up gets a full argument block. Step 1 of the new skill depends on attachments (designer assets, Figma sheets), so this is the one instruction an agent cannot follow.

**Fix.** Name the download mechanism with its arguments, as the `slack_read_thread` block does, or state the paste fallback when there is none.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+```bash
+pnpm --filter envs-validator build
+dotenv -e /tmp/change.env -e /tmp/instance.env -- pnpm --filter envs-validator validate
+```
+
+Schema organisation: `deploy/tools/envs-validator/CONTEXT.md`. This is what catches missing companion variables (`.when(...)`), forbidden combinations, and malformed nested JSON — a documentation check cannot.
+
+## Fetched configs
+
+JSON the instance **fetches** at startup — not instance chrome, no demo. Work in the `frontend-configs` checkout. Follow the `check-github-cli` skill. Confirm with the user before the commit and the PR. The `create-pr` skill's frontend template, ENVs label, and issue-from-branch steps do not apply.
```
tom2drum: **major** — Rule without a mechanism: "Work in the `frontend-configs` checkout" (and `SKIN-REQUESTS.md:7`, "the workspace's `frontend-configs` checkout") never says where that checkout is or how to obtain it. Both `frontend-configs` branches — the fetched-configs PR and the whole skin phase 1 — hang off a location the agent has to guess.

**Fix.** Give the path once (or the clone probe), in one of the two files, and point at it from the other.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+
+Follow `.agents/slack-thread.md`. Threads are typically in Russian.
+
+**Done when:** every message and attachment in the thread is in hand.
+
+### 2. Map the ask to variables
+
+Using `docs/ENVS.md` as the operator manual, pin each change to a variable and a documented value. **Stop** if it cannot be pinned.
+
+**Skin.** Read [`SKIN-REQUESTS.md`](SKIN-REQUESTS.md) when a designer owns the change: this instance's logo, icon, favicon, OG image, colour theme, hero banner, homepage highlights, or navigation promo banner. That branch runs to its own stop; it returns here at step 3. A chains-menu or footer icon that merely *lives* in `network-icons/` / `footer-icons/` is not that.
```
tom2drum, line 48: **major** — Duplication (`prose-smells.md`). The skin trigger list is verbatim in the pointer and in its target: here, "when a designer owns the change: this instance's logo, icon, favicon, OG image, colour theme, hero banner, homepage highlights, or navigation promo banner" against `SKIN-REQUESTS.md:3`, "The branch taken when a designer owns the change: this instance's logo, icon, favicon, …". The PR gate is duplicated the same way — `SKILL.md:107` and `SKIN-REQUESTS.md:59` both carry "Confirm with the user before the commit and the PR. The `create-pr` skill's frontend template, ENVs label, and issue-from-branch steps do not apply." Two copies of a routing rule is two things to keep in sync.

**Fix.** Keep the trigger list in the pointer only; keep the PR-gate sentence in one file and point at it from the other.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+
+Only send a DevOps message we are **sure** of. Every other rule serves that.
+
+- **Stop** on ambiguity. If the variable cannot be pinned, or the value is not in the documented set of allowed values, stop and tell the user.
+- **The user is the face.** Uncertainty and clarifying questions go to the user. Requester-facing posts are only the handover reply after an approved send, and the demo link on the skin branch.
+- **Intent is given.** Any question that turns on what someone *wants* is a stop, not a judgement call.
+- **Operator.** `docs/ENVS.md` is the manual. The docs pin legal values, not the consumers. A value the docs cannot pin is a documentation gap, fixed as documentation.
+- **Off.** Prefer a documented off-value over unsetting. Unset only when the docs show the default *is* the desired state.
+- **Mirror** the requester's targeting language. Unstated is not undeterminable — the DevOps bot resolves instances.
+- **Drift.** `NEXT_PUBLIC_HOMEPAGE_HERO_BANNER_CONFIG` and `NEXT_PUBLIC_COLOR_THEME_OVERRIDES` ship with a `frontend-configs` file change in the same run; take the skin branch.
```
tom2drum, line 21: **nit** · needs-human — Duplication (`prose-smells.md`). "**Drift.** `NEXT_PUBLIC_HOMEPAGE_HERO_BANNER_CONFIG` and `NEXT_PUBLIC_COLOR_THEME_OVERRIDES` ship with a `frontend-configs` file change in the same run; take the skin branch" restates a routing decision already made twice — step 2 (line 48) routes colour theme and hero banner to the skin branch, and `SKIN-REQUESTS.md:11-15` carries the full drift argument. Whether the **Sure** list is the right place for a third copy is a judgement about what that list is for.

**Fix.** Drop the bullet, or reduce it to the *why* that step 2 omits.

— Reviewed by Claude Opus 5

tom2drum (commented):
# Review — resolve-config-request skill

| | |
| --- | --- |
| Reviewed | `e0d8d66` → `695f76f` |
| Round | 1 of 3 |
| Findings | 1 blocker · 8 major · 5 nit |
| By axis | Spec — (no spec; skipped) · Standards 7 · Correctness 7 |
| Outcome | `blocked` |

Two axes ran. There is no spec for this change (it was done outside the task workflow, no `.agents/tasks/` folder), so the spec axis was not dispatched — standards and correctness carried the review.

The standards axis used `prose-smells.md` (the diff touches only the instruction surface plus `cspell.jsonc`). Checks were not run: PR mode leaves those to the Checks workflow.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
```diff
@@ -0,0 +1,81 @@
...
+Two configs, overlapping vocabulary; the request usually covers only one of them in text. Reconcile both before writing anything.
+
+- **Message text → hero banner.** background, text colour, button default/hover, light and dark.
+- **Figma sheet → colour theme overrides.** These tokens typically exist *only* in Figma.
+
+A `background` in the message text is not the theme's `bg.primary`.
+
+Read fills off a rendered frame (screenshot). Bound-variable lookup returns design-system defaults for per-instance raw fills. To tell which tokens are actually part of the request, diff each swatch against `DEFAULT_THEME_COLORS` in `src/toolkit/theme/foundations/colors.ts`: a token sitting at the default is not part of the request; one that differs is.
+
+**If Figma is unreachable, stop.** A half-applied skin reaching DevOps is worse than a blocked request.
```
tom2drum: **nit** · needs-human — Rule without a mechanism: "Read fills off a rendered frame (screenshot). Bound-variable lookup returns design-system defaults for per-instance raw fills" and "**If Figma is unreachable, stop.**" both assume a Figma access path that is never named, so "unreachable" is not an observable state. Whether this belongs here or is assumed session context is a call about the skill's audience.

**Fix.** Name the Figma tool or probe once, so the stop condition is checkable.

— Reviewed by Claude Opus 5

cspell.jsonc
```diff
@@ -34,7 +34,9 @@
         // Posthog project key
         "phc_\\w+",
         // GitHub account names
-        "@(\\w|-)+"
+        "@(\\w|-)+",
+        // Russian copy in agent skills (DevOps / Slack templates)
+        "[А-Яа-яЁё]+"
```
tom2drum, line 39: **nit** — `"[А-Яа-яЁё]+"` is dead config. cspell already ignores Cyrillic; removing this `ignoreRegExp` still yields 0 issues on both new files (verified). It also has a cost — the pattern silences any future Cyrillic typo repo-wide, so a real one in a DevOps template would go unreported.

**Fix.** Drop the rule and its comment.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+
+No commit without the user's explicit confirmation in this conversation.
+
+## Scope
+
+**In:** variables in `docs/ENVS.md`, plus start-time variables like `FAVICON_MASTER_URL`.
```
tom2drum: **nit** — "**In:** variables in `docs/ENVS.md`, **plus** start-time variables like `FAVICON_MASTER_URL`" reads as two disjoint sets, but `FAVICON_MASTER_URL` is itself in `docs/ENVS.md`. What actually makes it special is that `/node-api/config` cannot see it — which line 90 already says.

**Fix.** "variables in `docs/ENVS.md`, including start-time ones such as `FAVICON_MASTER_URL`".

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+- **Outside the frontend, or mixed with a non-configurable ask** — do the configurable frontend part only, and tell the user what was left out.
+- **CDN-only trees** in `frontend-configs` with no `docs/ENVS.md` variable (`multisearch/`, `token-icons/`, `nft-images/`, `explorer-logos/`, `meta-suites-logos/`) — stop and tell the user.
+
+## Steps
+
+### 1. Read the thread
+
+Follow `.agents/slack-thread.md`. Threads are typically in Russian.
+
+**Done when:** every message and attachment in the thread is in hand.
```
tom2drum, line 42: **nit** — Duplication (`prose-smells.md`): "**Done when:** every message and attachment in the thread is in hand." is the same sentence as `.agents/slack-thread.md:29`, which the step above it just told the agent to follow.

**Fix.** Drop this one — the shared doc is the authoritative site for it.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKILL.md
tom2drum: Fixed in 4e4c2d02a. Validation now runs from `deploy/tools/envs-validator/` with the `test.sh` preamble (placeholders + `test/assets` + build), then `pnpm exec dotenv` for the overlay.

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
tom2drum: Fixed in 4e4c2d02a. A new alias still needs `pnpm presets:sync` and the four files committed **on the demo branch** so the review deploy can dispatch it. No revert: they die with that branch.

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
tom2drum: Fixed in 4e4c2d02a. Teardown is now: `gh workflow run cleanup.yml --ref <branch>`, confirm 404, then delete the branch.

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
tom2drum: Fixed in 4e4c2d02a. Scoped to no **designer** gate; the user's commit confirmation still applies.

.agents/skills/resolve-config-request/SKILL.md
tom2drum: Fixed in 4e4c2d02a. Fetch now passes `--omit-local-envs`. Dropped `ignoredEnvs` / `deprecatedEnvs` are read from `GET /node-api/config`.

.agents/slack-thread.md
tom2drum: Fixed in 4e4c2d02a. If `slack_read_thread` is missing, unauthenticated, or errors, ask the user to paste the thread; stop if they decline.

.agents/slack-thread.md
tom2drum: Fixed in 4e4c2d02a. Connector still only yields `url_private`. If Files Reader is installed, GET that URL with the app token; otherwise paste or upload, and stop if the user declines.

.agents/skills/resolve-config-request/SKILL.md
tom2drum: Fixed in 4e4c2d02a. Checkout is a workspace folder named `frontend-configs` or `blockscout_frontend_configs`. If none, stop and ask the user to add it. Skin points at the same rule.

.agents/skills/resolve-config-request/SKILL.md
tom2drum: Fixed in 4e4c2d02a. Trigger list stays in the pointer. Commit/PR confirmation and the `create-pr` exceptions live once under Fetched configs; skin points there.

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
tom2drum: Fixed in 4e4c2d02a. Named `get_screenshot` and `get_variable_defs`; stop if they fail or Figma is unreachable.

cspell.jsonc
tom2drum: Rejected. The first commit of this PR failed the pre-commit cspell hook on `Привет`, `нужно`, `поставить`, and the rest of the Russian template. The `ignoreRegExp` is why that hook now passes; dropping it would re-break commits.

.agents/skills/resolve-config-request/SKILL.md
tom2drum: Fixed in 4e4c2d02a. Now: variables in `docs/ENVS.md`, including start-time ones such as `FAVICON_MASTER_URL`.

.agents/skills/resolve-config-request/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+Post the approved message to `blockscout-devops-requests` — resolve the channel ID from `.agents/TEAM.md`. Then reply in the original thread — same language as the thread — with a link to that message, saying the request has been handed over and is waiting on DevOps.
+
+**Done when:** both posts exist and the original-thread reply carries the DevOps link.
+
+## Reading an instance's env
+
+Every deployed instance exposes `{ envs: { …all NEXT_PUBLIC_* } }` at `<url>/node-api/config`. The CLI:
+
+```bash
+tools/dev-server/fetch.sh <alias> --omit-local-envs --out=/tmp/instance.env
```
tom2drum, line 85: **blocker** · regression — Adding `--omit-local-envs` fixed step 3's diff and broke step 4's validation, because one fetched env now serves both.

That flag drops the five `localEnvs` keys (`tools/dev-server/envs-rules.json`): `NEXT_PUBLIC_APP_PROTOCOL`, `_HOST`, `_PORT`, `_ENV`, `NEXT_PUBLIC_API_WEBSOCKET_PROTOCOL`. But `deploy/tools/envs-validator/schema.ts:52` has `NEXT_PUBLIC_APP_HOST: yup.string().required()`. So the documented step-4 overlay on this file exits 1 — `🚨 ENVs validation failed … NEXT_PUBLIC_APP_HOST is a required field` — where the same overlay without the flag reports `👍 All good!`. Step 4's `Done when` ("the validator accepts the overlay") is unreachable again, one gate later than round 1.

The two steps genuinely want different files: step 3 needs the *real* remote values (hence the flag), step 4 needs a *complete* env the schema will accept.

**Fix.** Keep `--omit-local-envs` for the step-3 diff, and have Validation add the four `NEXT_PUBLIC_APP_*` keys back (or fetch a second copy without the flag purely for the overlay). Whichever way, say which file each step reads.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
```diff
@@ -0,0 +1,79 @@
...
+Hostname: `review-<branch-slug>.k8s-dev.blockscout.com`. Follow the `deploy-demo` skill.
+
+Verify the theme by **computed values** — read the resolved CSS custom properties in both light and dark. The first click after opening the theme settings pop-over is swallowed; click again. A demo also confirms two things a local run cannot: that the favicon bundle is generated at container start from `FAVICON_MASTER_URL`, and that the envs-validator accepts the inline JSON blobs.
+
+## Phase 1 — demo
+
+No **designer** gate on the branch, the configs PR, the demo, or the demo-link message — the PR is open but unmerged, so nothing can break a production instance. The user's commit confirmation still applies. The requester (designer) is the gatekeeper for phase 2; the user relays that approval. There is no automated watch on the thread — the user monitors and continues this session.
+
+1. Produce the assets and JSON configs from the thread and Figma.
+2. Open a PR on `frontend-configs` (target `main` branch). Commit/PR confirmation and `create-pr` exceptions: parent **Fetched configs**.
```
tom2drum, line 59: **major** · regression — "Open a PR on `frontend-configs` (target `main` branch)" replaced a correct pointer ("target per that repo's README") with a hardcoded branch that is wrong for half of this very skill's files.

That repo's README has two procedures. "How to update existing configs (e.g for marketplace, for network menu)" branches from **`dev`**, PRs to `dev`, validates on the development stand, then `dev`→`main`. Only "How to update existing configs for network assets (logos, icons)" branches from and PRs to `main`.

A skin request spans both: `network-logos/`, `network-icons/`, `favicons/`, `og-images/` are the assets path, but `hero-banner/`, `color-themes/`, `homepage-highlights/` are JSON configs and belong on `dev`. Targeting `main` for those skips the dev-stand validation the README puts there.

**Fix.** Restore the pointer — "target per that repo's README" — or, if the split is worth stating here, name both: assets → `main`, JSON configs → `dev`.

— Reviewed by Claude Opus 5

.agents/slack-thread.md
```diff
@@ -25,3 +23,10 @@ Arguments:
 ```
 
 If the thread has more than 200 messages, paginate with `cursor` until the full conversation is read.
+
+Attachments often carry the actual content. The Slack connector reports metadata only (`url_private`).
+
+- **Files Reader** installed ([workspace app directory](https://blockscout.slack.com/marketplace/A0BMY22GBQR-files-reader)): GET each `url_private` with that app's token (`Authorization: Bearer`).
+- Otherwise, or if the GET fails: ask the user to paste or upload each file. Stop if they decline.
```
tom2drum: **major** · needs-human — The paste fallback on line 30 does close the dead end, so the blocking half of the round-1 finding is fixed. Two things remain, and both turn on knowledge I do not have:

1. **The Files Reader branch still names no token source.** "GET each `url_private` with that app's token (`Authorization: Bearer`)" — where does an agent get that token? Every other mechanism in these two files names its source. As written this branch is unfollowable, so in practice it always falls through to paste.

2. **"The Slack connector reports metadata only" may no longer be true.** A tool named `slack_read_file` exists on the same connector as `slack_read_thread` and takes a file id. I have not exercised it on a real attachment, so I cannot say it returns usable content — but if it does, this whole bullet is unnecessary and the instruction is actively pointing away from the working path.

**Fix.** Try `slack_read_file` on a real thread attachment. If it works, replace the Files Reader bullet with it. If it does not, name where the token comes from — or drop the branch and keep paste as the only path.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
```diff
@@ -0,0 +1,79 @@
...
+Two configs, overlapping vocabulary; the request usually covers only one of them in text. Reconcile both before writing anything.
+
+- **Message text → hero banner.** background, text colour, button default/hover, light and dark.
+- **Figma sheet → colour theme overrides.** These tokens typically exist *only* in Figma.
+
+A `background` in the message text is not the theme's `bg.primary`.
+
+Read fills off a rendered frame (`get_screenshot`). Bound-variable lookup (`get_variable_defs`) returns design-system defaults for per-instance raw fills. To tell which tokens are actually part of the request, diff each swatch against `DEFAULT_THEME_COLORS` in `src/toolkit/theme/foundations/colors.ts`: a token sitting at the default is not part of the request; one that differs is.
+
+**If `get_screenshot` / `get_variable_defs` fail, or Figma is unreachable, stop.** A half-applied skin reaching DevOps is worse than a blocked request.
```
tom2drum: **nit** · needs-human — The fix named `get_screenshot` and `get_variable_defs`, which are Figma Dev Mode MCP tool names. No Figma MCP server is configured in this repo, so as the instruction stands the stop condition on line 34 keys off two tools an agent here cannot call — which is the outcome round 1 warned was worse than the original gap.

This needs you: if that server is present at user scope in the sessions where you run this skill, the names are right and only their origin is missing. If it is not reliably present, the stop condition should be phrased against something observable.

**Fix.** Say once where the Figma tools come from ("requires the Figma Dev Mode MCP server"), so "they fail" is distinguishable from "they were never there".

— Reviewed by Claude Opus 5

tom2drum (commented):
# Review — resolve-config-request skill

| | |
| --- | --- |
| Reviewed | `695f76f` → `4e4c2d0` (round 1 was `e0d8d66` → `695f76f`) |
| Round | 2 of 3 — arbitration |
| Findings | 1 blocker · 2 major · 2 nit (3 nits carried, deferred) |
| Outcome | `blocked` |

Arbitration only — the 14 round-1 findings re-checked against the fixes, plus regressions the fixes themselves introduced. No fresh axis pass.

## Rulings

| Round-1 finding | Ruling |
| --- | --- |
| Validator recipe unrunnable (`SKILL.md:99-101`) | ✅ **verified** — ran it: `collect_envs.sh` + `cp -r test/assets ./public/` + build all exit 0, `pnpm exec dotenv` resolves, validator reports `All good!` |
| `presets:sync` / demo preset (`SKIN-REQUESTS.md:48`) | ✅ **verified** — `sync-preset-lists.mjs` writes exactly `deploy-review.yml` + `.vscode/tasks.json`; `deploy-review.yml` is `workflow_dispatch`-only and a dispatch reads its input enum from the dispatched ref, so committing on the demo branch works and needs no revert |
| Teardown order strands the demo | ✅ **verified** — `cleanup.yml`'s `workflow_dispatch` + `appName: review-$GITHUB_REF_NAME_SLUG` make dispatch-then-delete the only order that resolves the slug |
| "No approval gate" contradiction | ✅ **verified** — scoped to *designer* gate; no longer contradicts `SKILL.md:23` |
| `fetch.sh` misreports live env (`SKILL.md:85`) | ❌ **incomplete, and it introduced a blocker** — see the inline comment |
| Slack failure branch dropped | ✅ **verified** — `slack-thread.md:3` names a real exit; `SKILL.md:40` inherits it |
| Attachment download names no mechanism | ⚠️ **needs-human** — the paste fallback closes the dead end, but two points remain; see the inline comment |
| `frontend-configs` checkout unnamed | ✅ **verified** — both named folder names resolve, and the "stop and ask" branch covers absence |
| Duplication (trigger list, PR gate) | ✅ **verified** — trigger list only at `SKILL.md:48`; the commit/PR + `create-pr` sentence appears once, with `SKIN-REQUESTS.md:7` and step 2 pointing at it |
| Figma mechanism unnamed (nit) | ⚠️ **needs-human** — see the inline comment |
| `FAVICON_MASTER_URL` wording (nit) | ✅ **verified** |
| **cspell dead config (nit) — rejected** | ✅ **reject accepted.** Your reject is right and my round-1 finding was wrong. I re-tested it: with `"[А-Яа-яЁё]+"` replaced by a never-matching pattern, the hook's own invocation (`npx cspell --config cspell.jsonc --file-list stdin --no-must-find-files`) reports **12 issues** on `SKILL.md` — `Привет`, `нужно`, `поставить`, `убрать`, `перезапустить`, `фронт`, `чтобы`, `подтянуть`, `обновлённый`, `конфиг`. The rule is load-bearing exactly as you said. Closing as `rejected-accepted`; `cspell.jsonc` restored, tree clean. |

**Carried, still deferred** (nits, never auto-fixed, no reply needed): the `Drift` bullet duplication at `SKILL.md:21`, and the duplicated `Done when:` at `SKILL.md:42` / `slack-thread.md:32`.

— Reviewed by Claude Opus 5

.agents/slack-thread.md
tom2drum: Fixed in 110ad3668. `slack_read_file` is not on this connector. Attachments now go through the `slack-file` skill (script + Keychain service `slack-files-token`). Paste/upload remains the fallback if that skill stops.

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
tom2drum: Rejected. That README split is outdated. Configs PRs go to `main` directly — the hardcoded target is intentional.

.agents/skills/resolve-config-request/SKILL.md
tom2drum: Fixed in 110ad3668. Step 3 still diffs `/tmp/instance.env` (`--omit-local-envs`). Validation copies it, appends `localEnvs` from `envs-rules.json`, and overlays on that copy so `NEXT_PUBLIC_APP_HOST` is present.

cspell.jsonc
tom2drum: **Reject accepted.** You are right and my finding was wrong — recording the ruling on the thread so the record is self-describing, not only in the round-2 review body.

I retested it: with `"[А-Яа-яЁё]+"` replaced by a never-matching pattern, the pre-commit hook's own invocation — `npx cspell --config cspell.jsonc --file-list stdin --no-must-find-files` — reports **12 issues** on `SKILL.md`: `Привет`, `нужно`, `поставить`, `убрать`, `перезапустить`, `фронт`, `чтобы`, `подтянуть`, `обновлённый`, `конфиг`. Exit 1. The `ignoreRegExp` is load-bearing exactly as you said, and my round-1 claim that it was dead config was simply wrong. `cspell.jsonc` restored, tree clean.

Closing as `rejected-accepted`.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKILL.md
tom2drum: **Ruling: fix incomplete** (round 2) — recording it here so the thread carries its own outcome.

`--omit-local-envs` did fix step 3's diff, but it stripped `NEXT_PUBLIC_APP_HOST`, which `deploy/tools/envs-validator/schema.ts:52` marks `.required()` — so step 4 broke. That follow-on is tracked on its own thread rather than here: see the round-2 blocker comment on `SKILL.md:85`.

— Reviewed by Claude Opus 5

.agents/skills/slack-file/scripts/slack-file
```diff
@@ -0,0 +1,135 @@
...
+  meta=$(slack_curl "https://slack.com/api/files.info?file=$id")
+  if ! api_ok "$meta"; then
+    echo "slack-file: $id: $(api_err "$meta")" >&2
+    status=1
+    continue
+  fi
+  name=$(jq -r '.file.name' <<<"$meta")
+  url=$(jq -r '.file.url_private_download // .file.url_private' <<<"$meta")
+  out="$dir/${id}-${name//[^A-Za-z0-9._-]/_}"
+  slack_curl -L -o "$out" "$url"
```
tom2drum: **major** — The download leg writes failure bodies to disk and reports success, so `SKILL.md:46` ("If the command fails: ask the user to paste or upload each file") can never fire.

`slack_curl -L -o "$out" "$url"` has no `-f` and no status check. Without `-f`, curl exits **0** on an HTTP 403/404 — expired token, missing `files:read`, no channel access — having written the error JSON or Slack's HTML login page into `$out`. The loop then `echo "$out"` and the script exits `0`. The agent Reads a login page believing it is the attachment, which is worse than a clean failure: it corrupts step 1 of `resolve-config-request` silently.

The same script already gets this right one leg up — line 78 uses `curl -sS -f -F …` for the upload. Only the download is missing it.

`-L` compounds it. Since curl 7.58 a custom `Authorization` header is stripped on redirect to a **different host**, so any hop (a login redirect, an `external_url`-backed file) re-requests unauthenticated and lands exactly in the case above.

**Fix.** Add `-f` and route the failure into the loop's existing error path — note `set -euo pipefail` means a bare `-f` would abort the whole run on the first bad id instead of continuing, so it needs to be explicit:

```bash
if ! slack_curl -fL -o "$out" "$url"; then
  echo "slack-file: $id: download failed" >&2
  rm -f "$out"
  status=1
  continue
fi
```

And either use `--location-trusted` so the header survives a hop, or drop `-L` so a redirect surfaces as a failure rather than a bad file.

— Reviewed by Claude Opus 5

tom2drum (commented):
# Review — resolve-config-request skill

| | |
| --- | --- |
| Reviewed | `4e4c2d0` → `110ad36` |
| Round | 3 — arbitration |
| Findings | 0 blocker · 1 major · 0 new nit (1 nit carried, open) |
| Outcome | `blocked` |

Arbitration on the four round-2 findings. Each ruling is now **replied onto its own thread** as well as listed here — round 2 put them in this body only, which left the record unreadable to anything that gathers `pulls/3641/comments`.

## Rulings

| Round-2 finding | Ruling |
| --- | --- |
| `--omit-local-envs` broke step 4 (blocker) | ✅ **verified** — ran the whole recipe: fetch → `cp` + `jq` append → `test.sh` preamble → overlay validate = `👍 All good!`, exit 0. The same command against the un-appended file still exits 1 on `NEXT_PUBLIC_APP_HOST is a required field`, so the append is precisely what closes the gate. Duplicate keys within one `-e` file resolve last-wins, which is the direction the append needs. |
| `frontend-configs` branch target (major) — **rejected** | ✅ **reject accepted.** You are right; my finding was wrong. `origin/dev` in that repo is dead — `git rev-list --count origin/main..origin/dev` = **0**, the reverse = **581**, last commit **2024-04-05** against **2026-08-14** on `main`. Recent JSON-config work lands on `main` from feature branches (#646 `plume-testnet-skin` carried `color-themes` + `hero-banner` JSON *and* assets; #645 the same); no merge from `dev` in the last 8. I weighted that README over actual practice, and the README is what is stale. Closing as `rejected-accepted`. |
| Slack attachment mechanism (major, needs-human) | ⚠️ **incomplete** — the `slack-file` skill is the right answer and the plumbing checks out (`bash -n` clean, `getopts` correct, `-K -` header survives alongside `--get`/`--data-urlencode`/`--data-binary`, `.claude/skills` is a directory symlink so the skill resolves with no per-file step, `lint:doc-links` and cspell pass, and `.agents/README.md`'s `skills/*/SKILL.md` row covers it generically). One defect left in the download leg — see the inline comment. |
| Figma tool names (nit, needs-human) | ⏳ **still open** — `SKIN-REQUESTS.md:34` is untouched by this commit and no reply answers it. Deferred, still yours to call. |

## Not anchorable

**Observation, not a finding against this PR:** the `frontend-configs` README still documents the `dev`-first procedure that the branch history shows was abandoned in 2024. Worth telling that repo's owner, since it is what misled this review.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKILL.md
tom2drum: **Ruling: fix verified.**

Ran the whole documented recipe end to end: `fetch.sh eth --omit-local-envs --out=/tmp/instance.env` (exit 0, 79 envs, no `APP_HOST`) → `cp` + the `jq … >> /tmp/instance.validate.env` append (adds the 5 `localEnvs` lines) → the `test.sh` preamble (`collect_envs.sh`, `cp -r test/assets ./public/`, build, exit 0) → `pnpm exec dotenv -e /tmp/change.env -e /tmp/instance.validate.env -- pnpm run validate` = `👍 All good!`, exit 0.

The same command against the un-appended `/tmp/instance.env` still exits 1 on `NEXT_PUBLIC_APP_HOST is a required field`, so the append is exactly what closes the gate. Quoting is a non-issue (all five values are bare tokens), and duplicate keys inside one `-e` file resolve last-wins — the direction the append needs, so it stays safe if a fetched file ever kept one of those keys.

Closing as `fixed`.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
tom2drum: **Reject accepted.** You are right and my finding was wrong.

I checked the configs repo rather than its README. `origin/dev` is dead: `git rev-list --count origin/main..origin/dev` = **0**, the reverse = **581**, and its last commit is **2024-04-05** (`Merge branch 'main' into dev`) against **2026-08-14** on `main`. Recent JSON-config work lands on `main` from feature branches — #646 `plume-testnet-skin` carried `color-themes`/`hero-banner` JSON *and* assets, #645 `hskchain-rebrand` the same — with no merge from `dev` in the last 8 merge commits.

I weighted that README over actual practice; the README is what is stale. The hardcoded `main` target is correct. Worth telling that repo's owner the README needs updating, but that is not a defect here.

Closing as `rejected-accepted`.

— Reviewed by Claude Opus 5

.agents/slack-thread.md
tom2drum: **Ruling: fix incomplete** — reopening this thread.

The `slack-file` skill is the right answer and the plumbing checks out: `bash -n` clean, download `getopts` parses `-d <dir> F123` correctly with `-h` reaching `usage`, the `-K -` auth header survives alongside `--get`/`--data-urlencode`/`-X POST --data-binary`, `.claude/skills` is a directory symlink so the new skill resolves with no per-file step, and `pnpm lint:doc-links` (56 files) plus cspell both pass. `.agents/README.md` needs no new row — its `skills/*/SKILL.md` entry covers it.

One defect remains in the download leg: `scripts/slack-file:132` has no `-f`, so an HTTP 403/404 writes the error body to disk and the script still exits 0 — which means `SKILL.md:46` ("ask the user to paste") never fires. Tracked as a `major` in the round-3 review comment on that line.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
tom2drum: **Ruling: still open.**

`git diff 4e4c2d02a HEAD` touches only `resolve-config-request/SKILL.md`, the two new `slack-file` files, and `.agents/slack-thread.md`. `SKIN-REQUESTS.md:34` is untouched and no reply answers this finding.

Staying `deferred` + `needs-human` — it is a nit, and the call is yours: if the Figma Dev Mode MCP server is present at user scope when you run this skill, the tool names are right and only their origin is missing.

— Reviewed by Claude Opus 5

.agents/slack-thread.md
tom2drum: Remaining download-leg defect fixed in f8d8f5ff0 (`-f` on the GET, failed files removed, non-zero status). `slack-thread.md` is unchanged — it already follows the `slack-file` skill.

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
tom2drum: Fixed in f8d8f5ff0. `get_screenshot` / `get_variable_defs` need the Figma plugin in Claude Code or Cursor. The stop now covers those tools being absent, not only a failed call.

.agents/skills/slack-file/scripts/slack-file
tom2drum: Fixed the `-f` half in f8d8f5ff0: download now uses `-fL`, deletes a failed file, sets `status=1`, and continues so the paste fallback can fire.

Rejected `--location-trusted` and dropping `-L`. A real attachment already downloaded with `-L`; Slack hops `url_private_download` to a signed CDN URL, and `--location-trusted` would send the Keychain token to that host. Default `-L` already strips `Authorization` on a different host, which is the hop we want.

tom2drum: **Ruling: fix verified on the `-f` half — and your reject on the `-L` half is right. My round-3 suggestion was harmful; I withdraw it.**

Exercised the real script against a local server, with `security` and `curl` stubbed on `PATH` (fake token, `slack.com` rewritten to loopback — no Slack traffic, real token never read):

- `slack-file -d out FBAD FOK FMETABAD FRED` → printed `slack-file: FBAD: download failed`, left **no** `FBAD` file on disk, still downloaded `FOK` and the redirect case, exited **1**. So `if !` suppresses `errexit`, the loop continues, and `status=1` reaches `exit "$status"`.
- All-success run → both paths printed, exit **0**.

On the reject: with curl 8.7.1 across a cross-host redirect, default `-L` sends **no** Authorization to the second host, while `--location-trusted` sends `Bearer …`. So what I proposed would have forwarded your Keychain Slack token to whatever host the redirect names — a pre-signed CDN URL that needs no auth at all. That is a credential leak, and you were right to refuse it.

Dropping `-L` is wrong too: `curl -f -o file` against a 302 with no `-L` exits **0** having written **0 bytes** — reproducing the exact silent-corruption bug the finding was about.

Your arrangement is the correct and safer one: auth on hop 1, stripped on hop 2, `-f` catching any hop that returns ≥400.

Closing as `fixed` / `rejected-accepted`.

— Reviewed by Claude Opus 5

.agents/slack-thread.md
tom2drum: **Ruling: fix verified.** Closing.

The remaining half of this thread was the download legs missing `-f`. It now fails loudly, deletes the partial file, sets a non-zero status and continues — so `slack-file/SKILL.md:46` ("ask the user to paste or upload each file") can finally fire. Verified by running the script against a local 403, not by reading it.

Agreed that `.agents/slack-thread.md` needed no change: it already routes through the `slack-file` skill.

— Reviewed by Claude Opus 5

.agents/skills/resolve-config-request/SKIN-REQUESTS.md
tom2drum: **Ruling: fix verified.** Closing.

`SKIN-REQUESTS.md:32` now names the origin ("Both need the Figma plugin in Claude Code or Cursor") and :34 stops on "absent or fail". That is the part that mattered: tool presence is directly observable in an agent's own tool list, so **absent** is now checkable and distinguishable from a failed call — which it was not when the stop keyed only off the tools failing.

Minor and not worth a round: "plugin" is loose for what is an MCP integration. Not misleading enough to act on.

— Reviewed by Claude Opus 5

.agents/skills/review-changes/SKILL.md
```diff
@@ -213,12 +222,19 @@ lines outside the diff — go into the review body under `## Not anchorable`, in
 and the all-or-nothing 422 hazard: [`gh-commands.md`](gh-commands.md). PR mode writes no `review.md`;
 the PR is the record.
 
+**PR mode, arbitration.** New findings still go in the one batched event. A **ruling** does not: it goes
```
tom2drum: **blocker** — The commit whose entire purpose is downstream readability omits the one token downstream keys on.

This block says a ruling "goes on its finding's own thread, as a reply" and never requires the `— Reviewed by` footer on that reply. The footer instruction sits at :217-218, inside the **PR mode** paragraph, scoped to "Each comment" of the batched event — it does not reach a thread reply.

That breaks the consumer. `resolve-review/SKILL.md:22-24` classifies comments by footer **first**; a footerless reply falls through to "A human — neither of the above", whose verdicts are `fix · answered` and which "may never be rejected". So an arbitration ruling — including one that *accepted a rejection* — comes back as an unrejectable human comment demanding a fix.

It happened to work in this PR only because the rulings were written with footers by hand. The instruction is what ships.

**Fix.** State it in this block: each ruling reply carries the same `— Reviewed by <name>` footer as a review comment.

— Reviewed by Claude Opus 5

.agents/skills/review-changes/gh-commands.md
```diff
@@ -130,18 +130,28 @@ query {
         | "\(.id) \(.comments.nodes[0].databaseId)"'
 ```
 
+Drop the `select(...)` to list every thread with its state — which is what arbitration needs, since it
```
tom2drum: **major** — "Drop the `select(...)` to list every thread with its state" is false as written, and it is the sentence I added to make arbitration able to see thread state.

Dropping `select(.isResolved==false)` leaves the format string `"\(.id) \(.comments.nodes[0].databaseId)"` — which never mentions `isResolved`. Run against PR 3641 it returns 50 rows of id + databaseId and **no state at all**. The `-q` needs changing too, not just the filter.

**Fix.** Show the arbitration variant explicitly rather than describing a delta:

```
-q '.data.repository.pullRequest.reviewThreads.nodes[]
     | "\(.id) \(.isResolved) \(.comments.nodes[0].databaseId)"'
```

— Reviewed by Claude Opus 5

.agents/skills/review-changes/SKILL.md
```diff
@@ -169,7 +169,8 @@ the current diff. It does exactly three things:
 3. Flag regressions introduced **by the fixes only**. New findings elsewhere are out of scope for this
    round.
 
-**Done when**: every open finding is either closed or carries a ruling.
+**Done when**: every open finding is either closed or carries a ruling, and each ruling sits on the
+finding's own record — its thread in PR mode, its exchange line in markdown mode.
```
tom2drum: **major** — This `Done when` demands work that only happens two steps later, so no agent can satisfy both it and step 5.

All posting lives in step 5. Between them sits step 4's Normalize, which drops findings whose anchors fail and recalibrates severity. An agent obeying this criterion posts rulings **before** they are normalized; an agent obeying step 5 arrives here unable to tick the box. I asserted the same requirement at two stages and made the earlier one unsatisfiable.

**Fix.** Revert this to "every open finding is either closed or carries a ruling" and leave the landing to step 5's `Done when`, which already states it.

— Reviewed by Claude Opus 5

.agents/skills/review-changes/SKILL.md
```diff
@@ -213,12 +222,19 @@ lines outside the diff — go into the review body under `## Not anchorable`, in
 and the all-or-nothing 422 hazard: [`gh-commands.md`](gh-commands.md). PR mode writes no `review.md`;
 the PR is the record.
 
+**PR mode, arbitration.** New findings still go in the one batched event. A **ruling** does not: it goes
+on its finding's own thread, as a reply, plus that thread's resolve state. Resolve the threads a ruling closes
+(`fix-verified`, `rejected-accepted`), leave open the ones it does not (`fix-incomplete`, disputed,
```
tom2drum: **major** — `fix-verified` and `fix-incomplete` are defined nowhere in the skill set; I invented them at the point of use.

Three vocabularies already exist and none contains these: `review-template.md:33` defines Status as `open | fixed | rejected-pending | rejected-accepted | disputed | needs-human | deferred`; step 3 names its outcomes "verify" / "agree" / "disagree"; `resolve-review` uses `fix | reject | answered | needs-human | deferred`. `rejected-accepted` on this same line *is* from the template — so the line mixes a defined name with two undefined ones.

Worse, the second copy of this policy in `gh-commands.md:153-154` adds a **fifth** name, `fix-wrong`, that this list omits. The two sites I added in one commit already disagree.

**Fix.** Reuse the template's Status names, or define the ruling enum once and have both files consume it.

— Reviewed by Claude Opus 5

.agents/skills/review-changes/SKILL.md
```diff
@@ -213,12 +222,19 @@ lines outside the diff — go into the review body under `## Not anchorable`, in
...
+`needs-human`, deferred nits), and unresolve a thread `resolve-review` closed on a fix that turned out not
+to hold.
```
tom2drum: **major** · standards — **Duplication**, which `prose-smells.md` calls the highest-yield smell on this surface. The resolve / leave-unresolved / unresolve policy is now stated in three places: here, `gh-commands.md:150-155`, and `resolve-review/SKILL.md:152-154`.

This is not hypothetical drift — the two copies I added in a single commit **already** disagree, with `fix-wrong` present in one and absent in the other. Three copies of a state machine is three things to keep in sync, and the commit shipped them out of sync on day one.

**Fix.** Keep the policy in `gh-commands.md` (where the mutations live). Here, state only that arbitration lands rulings on threads and sets their state, and point there.

— Reviewed by Claude Opus 5

.agents/skills/review-changes/SKILL.md
```diff
@@ -186,11 +187,19 @@ findings to `blocker` because that axis is all it can see.
   actually there and the claim still holds — and, in PR mode, that the line is in the PR diff. Drop
   what fails. Hallucinated line numbers and stale claims are the two things that end a reviewer's
   credibility, and this single pass catches both.
+- **Settle it by running it.** A claim a command can decide is not normalized until you have run that
```
tom2drum: **nit** · standards — Seven lines doing two lines of work, in a file whose own baseline warns against it.

"Drop what does not reproduce" restates the preceding bullet's "Drop what fails". The three-item list — "a rule that reads dead but is load-bearing, a procedure its own repo abandoned years ago, an incompatibility that only shows up on the second invocation" — is incident narrative from the review that prompted this edit, not executable instruction. And "The environment outranks every document that describes it" is then restated by the clause that follows it.

**Fix.** Cut to the two sentences that carry it: *A claim a command can decide is not normalized until you have run it and the finding quotes the output. The environment outranks any document describing it.*

— Reviewed by Claude Opus 5

.agents/skills/resolve-review/SKILL.md
```diff
@@ -70,7 +70,9 @@ Collect every **actionable** finding.
   `disputed` finding the latest round ruled on.
 - **PR mode** — inline review comments, PR-level reviews, and issue comments (`gh-commands.md`). Keep only
   unresolved, actionable threads. Drop already-resolved threads, your own prior replies, and bot status
-  noise (CodeRabbit "review skipped", Copilot's PR overview).
+  noise (CodeRabbit "review skipped", Copilot's PR overview). A reply carrying the `— Reviewed by` footer is
+  the reviewer's **ruling** on your last round, and a thread it reopened is a fix that did not hold — read
```
tom2drum: **nit** · standards — Duplication: "A reply carrying the `— Reviewed by` footer is the reviewer's **ruling**" is a third copy of the footer test, already stated in this file's own source table at :22-24 and in `gh-commands.md:92-97`.

Only the second clause is new information: that a **reopened** thread means a fix that did not hold. That part earns its place; the footer restatement does not.

**Fix.** Keep the reopened-thread sentence, drop the footer restatement.

— Reviewed by Claude Opus 5

tom2drum (commented):
# Review — round 4

| | |
| --- | --- |
| Reviewed | `110ad36` → `95821897` |
| Round | 4 |
| Findings | 1 blocker · 4 major · 4 nit |
| Outcome | `blocked` |

This round covers **two units**, because the branch grew one:

**1. `f8d8f5ff0` — arbitration, clear.** All three open findings verified and closed; rulings are on their own threads. The `-f` fix was tested by running the script against a local 403 (stubbed `security`/`curl`, no Slack traffic): the bad id errors, leaves no file, the loop continues, exit 1. Your reject of `--location-trusted` is **accepted** — measured, default `-L` sends no auth to a second host while `--location-trusted` sends `Bearer …`, so my round-3 suggestion would have forwarded the Keychain token to a pre-signed CDN URL that needs no auth. Actively harmful; withdrawn. Dropping `-L` is also wrong (`curl -f -o` on a 302 exits 0 writing 0 bytes).

**2. `95821897e` — the review-skill commit, reviewed for the first time.** It is unrelated to this PR's subject and no round had seen it. **I wrote it**, so an independent agent reviewed it against `prose-smells.md` and the repo's own checks; every finding below is against my own work and I am not contesting any of them. Two axes (standards, correctness); no spec exists.

## Not anchorable

Both are real but sit outside the PR diff's hunks.

- **nit · standards · `review-changes/SKILL.md:215`** — "One batched review event per round — **never N separate comments**" is now an absolute with an exception ten lines below it. A reader who stops at the absolute never learns of the exception. Qualify it in place: `never N separate comments (rulings excepted, below)`.
- **nit · correctness · `review-changes/review-template.md:9`** — untouched text says `resolve-review` "owns every Status transition after that and appends the exchange lines", which my new step-3 `Done when` ("its exchange line in markdown mode") contradicts. In scope because my change is what made it wrong. Amend the template comment to grant arbitration its own reviewer lines.

## Method note

`cspell.jsonc:12` ignores `.claude/worktrees`, so `pnpm lint:cspell` from the **main** checkout skips files inside a worktree. The explicit `--file-list` runs used in this review were not affected — they flagged real issues both times — but a plain `lint:cspell` from the main checkout would not have covered these files.

— Reviewed by Claude Opus 5

.agents/skills/review-changes/SKILL.md
tom2drum: **Reverted — this thread can be resolved.**

`95821897e` is reverted in `6221ea2c4` (pushed). The arbitration block that omitted the `— Reviewed by` footer requirement no longer exists on the branch, so this finding is moot rather than fixed. The revert restores `f8d8f5ff0` exactly — `git diff f8d8f5ff0 HEAD` is empty.

The underlying gap the commit was aimed at is still real and still open; it belongs on its own branch, not in this PR, whose subject is the config-request skill.

— Reviewed by Claude Opus 5

.agents/skills/review-changes/gh-commands.md
tom2drum: **Reverted — this thread can be resolved.**

`95821897e` is reverted in `6221ea2c4` (pushed). The "drop the `select(...)`" sentence no longer exists on the branch, so this finding is moot rather than fixed. The revert restores `f8d8f5ff0` exactly — `git diff f8d8f5ff0 HEAD` is empty.

The underlying gap the commit was aimed at is still real and still open; it belongs on its own branch, not in this PR, whose subject is the config-request skill.

— Reviewed by Claude Opus 5

.agents/skills/review-changes/SKILL.md
tom2drum: **Reverted — this thread can be resolved.**

`95821897e` is reverted in `6221ea2c4` (pushed). The step-3 `Done when` that required posting before step 5 no longer exists on the branch, so this finding is moot rather than fixed. The revert restores `f8d8f5ff0` exactly — `git diff f8d8f5ff0 HEAD` is empty.

The underlying gap the commit was aimed at is still real and still open; it belongs on its own branch, not in this PR, whose subject is the config-request skill.

— Reviewed by Claude Opus 5

.agents/skills/review-changes/SKILL.md
tom2drum: **Reverted — this thread can be resolved.**

`95821897e` is reverted in `6221ea2c4` (pushed). The invented `fix-verified` / `fix-incomplete` vocabulary no longer exists on the branch, so this finding is moot rather than fixed. The revert restores `f8d8f5ff0` exactly — `git diff f8d8f5ff0 HEAD` is empty.

The underlying gap the commit was aimed at is still real and still open; it belongs on its own branch, not in this PR, whose subject is the config-request skill.

— Reviewed by Claude Opus 5

.agents/skills/review-changes/SKILL.md
tom2drum: **Reverted — this thread can be resolved.**

`95821897e` is reverted in `6221ea2c4` (pushed). The third copy of the resolve/unresolve policy no longer exists on the branch, so this finding is moot rather than fixed. The revert restores `f8d8f5ff0` exactly — `git diff f8d8f5ff0 HEAD` is empty.

The underlying gap the commit was aimed at is still real and still open; it belongs on its own branch, not in this PR, whose subject is the config-request skill.

— Reviewed by Claude Opus 5

.agents/skills/review-changes/SKILL.md
tom2drum: **Reverted — this thread can be resolved.**

`95821897e` is reverted in `6221ea2c4` (pushed). The "Settle it by running it" bullet no longer exists on the branch, so this finding is moot rather than fixed. The revert restores `f8d8f5ff0` exactly — `git diff f8d8f5ff0 HEAD` is empty.

The underlying gap the commit was aimed at is still real and still open; it belongs on its own branch, not in this PR, whose subject is the config-request skill.

— Reviewed by Claude Opus 5

.agents/skills/resolve-review/SKILL.md
tom2drum: **Reverted — this thread can be resolved.**

`95821897e` is reverted in `6221ea2c4` (pushed). The footer restatement in the Gather step no longer exists on the branch, so this finding is moot rather than fixed. The revert restores `f8d8f5ff0` exactly — `git diff f8d8f5ff0 HEAD` is empty.

The underlying gap the commit was aimed at is still real and still open; it belongs on its own branch, not in this PR, whose subject is the config-request skill.

— Reviewed by Claude Opus 5

PR #3645, opened 2026-08-18
Move production builds back to Turbopack on Next.js 16.3.1

tom2drum:
## What & why

Bumps Next.js **16.2.6 → 16.3.1** and switches production builds from webpack back to **Turbopack** (the Next default).

Next 16.3.1 fixes the Turbopack scope-hoisting bug that miscompiled the Dynamic-labs SDK and hard-crashed every page of any `NEXT_PUBLIC_ACCOUNT_AUTH_PROVIDER=dynamic` instance in production — the crash that forced `next build --webpack` in [ADR 0001](https://github.com/blockscout/frontend/blob/claude/nextjs-version-bump-b89306/.agents/adr/0001-webpack-for-production-builds.md). Verified fixed by building + serving on the `rootstock` (dynamic) preset: home page renders, Dynamic login modal opens, zero `DynamicContextProvider` errors.

The main motivation is **CI build time**. On 16.3.1 the old webpack-vs-Turbopack trade-off is gone — Turbopack is faster to build *and* faster at runtime.

## Perf A/B (rootstock, medians of 3 headless runs, same machine — within-method deltas only)

| Metric | webpack | Turbopack | Δ |
| --- | --- | --- | --- |
| Bundling time (`next build`) | 186 s | **87 s** | −53% |
| M1 FCP | 2047 ms | **1049 ms** | −49% |
| M2 first API request | 123 ms | **55 ms** | −55% |
| M5 blocking time | 1444 ms | **453 ms** | −69% |
| M6 JS before FCP | 2235 KB gz | 2575 KB gz | +15% |

Turbopack ships ~15% more JS before FCP but halves FCP and cuts blocking time ~2/3 — the cost that used to live in execution is gone. Rationale + full context in the new [ADR 0003](https://github.com/blockscout/frontend/blob/claude/nextjs-version-bump-b89306/.agents/adr/0003-turbopack-for-production-builds.md).

## Changes

- **package.json** — `next` → 16.3.1; drop `--webpack` from `build` and `build:next`.
- **next.config.js** — `experimental.useTypeScriptCli: false` (**required** — see gotcha below).
- **pnpm-workspace.yaml** — temporary `minimumReleaseAge` exceptions for `next` + its `@next/env` / `@next/swc-*` internals (16.3.1 is < 1 week old; entries tagged "remove after 20.08.2026").
- **tools/dev-server/prod.preset.sh** — default build → Turbopack; `--profile` stays on webpack.
- Docs — new ADR 0003; ADR 0001 marked superseded; `connect-wallet/CONTEXT.md` + `AGENTS.md` updated.

Kept on **webpack** deliberately: `build:analyze` (`@next/bundle-analyzer` is a webpack plugin) and `prod:preset --profile` (React `react-dom/profiling` alias is guaranteed on webpack, undocumented on Turbopack).

## Reviewer note — the `useTypeScriptCli` gotcha

Next 16.3 defaults its build-time type-check to the `tsc` CLI, which needs a `typescript/bin/tsc` binary. This repo runs the native TS compiler via the `@typescript/typescript6` alias, which ships `bin/tsc6` only — so the CLI path aborts the build claiming `typescript` is missing, on **both** bundlers (the check runs before bundling). `useTypeScriptCli: false` switches to the compiler-API path (`lib/typescript.js`, which the alias does provide); type-checking still runs and passes.

## Verification

- Turbopack prod build on rootstock (dynamic provider): clean, no crash.
- Webpack build still succeeds (fallback for analyze/profile intact).
- Docker image build (`pnpm build`) now uses Turbopack.


next.config.js
```diff
@@ -74,6 +74,12 @@ const moduleExports = {
   redirects,
   headers,
   output: 'standalone',
+  // Turbopack's standalone tracer copies only @swc/helpers/cjs and drops the esm/ entry points that
+  // Next's require-hook loads at runtime, so `node server.js` crashes on boot. Force the whole
+  // package into the standalone bundle until the tracer is fixed upstream.
+  outputFileTracingIncludes: {
+    '/**': [ './node_modules/@swc/helpers/**' ],
```
tom2drum, line 81: **blocker** — `outputFileTracingIncludes: { '/**': [ './node_modules/@swc/helpers/**' ] }` matches **zero** files under this repo's pnpm layout (`node_modules/@swc/helpers` is not hoisted; the package only exists at `node_modules/.pnpm/@swc+helpers@*/…`). The claimed standalone boot fix is a no-op, so `node server.js` can still fail with `Cannot find module '@swc/helpers/esm/…'`.

Add `@swc/helpers` as a direct dependency (so it hoists), or glob the `.pnpm` path that actually exists.

— Reviewed by Cursor Grok 4.5

src/features/connect-wallet/CONTEXT.md
```diff
@@ -56,13 +56,11 @@ disabled *fallback*.
...
-**Before changing anything in the dynamic-mode graph, read
-`.agents/adr/0001-webpack-for-production-builds.md`.** Reshaping it once already tripped a Turbopack
-scope-hoisting bug that mis-binds a context inside the Dynamic-labs SDK and hard-crashes every page
-in a production build — invisible in dev. That is why production bundles are built with webpack. The
-trigger was never reducible to a single import, so treat any graph change here as able to re-trigger
-it: verify with a production build (`pnpm prod:preset <alias>`) against an instance whose
-`NEXT_PUBLIC_ACCOUNT_AUTH_PROVIDER=dynamic`, not just in dev.
+Before changing anything in the dynamic-mode graph, verify it in a production build (`pnpm prod:preset <alias>`) against an instance whose
+`NEXT_PUBLIC_ACCOUNT_AUTH_PROVIDER=dynamic`, not just in
+dev. Reshaping it once tripped a Turbopack scope-hoisting bug that mis-bound a context inside the
```
tom2drum: **nit** — possible Duplication: "tripped a Turbopack scope-hoisting bug that mis-bound a context inside the Dynamic-labs SDK and hard-crashed every page in a production build — invisible in dev" restates the same meaning already owned by ADR 0003.

Keep the `pnpm prod:preset` verify rule; drop the crash summary and only point at ADR 0003 for why.

— Reviewed by Cursor Grok 4.5

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `9897b3743b75ffb375add4b7170bd156862f9038` → `895d6bcd01f5c7a6124d8f29838ce0a4534ea55c` |
| Round | 1 of 3 |
| Findings | 1 blocker · 1 major · 1 nit |
| By axis | Spec 0 (no task spec) · Standards 1 · Correctness 2 |
| Outcome | `blocked` |

## Not anchorable

**major · Correctness** — `.agents/tasks/3566-main-page-loading-perf/tools/README.md:62-63`

**Claim.** ``prod:preset` builds with the same bundler as the shipped image (webpack, per that ADR)` — this change switched both to Turbopack; the README still points at ADR 0001 and will mislead anyone re-running the perf tools.

**Suggested fix.** Point at ADR 0003 and say Turbopack; drop the webpack claim.

— Reviewed by Cursor Grok 4.5

next.config.js
tom2drum: Rejecting — the premise checks out (`node_modules/@swc/helpers` is not hoisted; it lives only under `.pnpm`), but the conclusion doesn't: the include is **not** a no-op. Verified with clean A/B builds — **without** the include the standalone bundle ships only `@swc/helpers/cjs` and `node server.js` crashes with `Cannot find module '@swc/helpers/esm/_interop_require_default.js'`; **with** it, the `esm/` entry points are present and the server boots (HTTP 200). Next resolves the include glob through pnpm's symlink even though the package isn't at top-level `node_modules`. Independently confirmed by building the Docker image locally with and without the fix — the app only starts with it. Leaving the include as-is.

src/features/connect-wallet/CONTEXT.md
tom2drum: Fixed in 25fdc4bff — dropped the crash restatement, kept the `pnpm prod:preset` verify rule and the pointer to ADR 0003 for the why.

PR #3649, opened 2026-08-20
Add a time-zone toggle to the CSV export dialog

### LINKED ISSUE #3615: CSV Export (time format)
**Now** 
The user does not understand in what format the time will be indicated during export.
Image

**Solution**[ FIGMA component](https://www.figma.com/design/4In0X8UADoZaTfZ34HaZ3K/Blockscout-design-system?node-id=33655-3341&t=KnzPlADJhLmctH2k-4)
Add a time toggle

Image

tom2drum:
## Description

Resolves #3615

The CSV export dialog's From/To pickers gave no indication of which time
zone the entered wall-clock was read in — it was silently resolved in the
user's local zone — so users couldn't tell what "time" the export period
actually meant.

This surfaces the app's existing "Local time format" setting inside the
dialog (bound to the same global `isLocalTime` preference) and makes the
pickers zone-explicit. With the toggle **on** (local, default) the
wall-clock is read as local with no suffix; with it **off** (UTC) the
wall-clock is read as UTC and a "UTC" suffix is shown on the fields, the
async downloads-list period, and the downloaded file name. The submitted
zone is persisted per async download item so its period always renders in
the same zone and format the dialog showed. The shared `DatePicker` gains
an optional `timeZoneSuffix` prop; the field is remounted on zone change
because the underlying picker only re-syncs its displayed text on
value/locale changes, not on a format change.

## Environment variables

None

## Minimum API version

None (the From/To filter is still sent as absolute UTC timestamps; only the client-side interpretation and display changed).

None

## Additional information

- The zone-aware file name applies to the synchronous download path; asynchronously generated files are named by the backend.
- Legacy async download items created before this change carry no zone flag and fall back to local-time rendering.

### DIFF
--- a/src/features/csv-export/components/dialog/CsvExportDialog.tsx
+++ b/src/features/csv-export/components/dialog/CsvExportDialog.tsx
@@ -1,11 +1,14 @@
 // SPDX-License-Identifier: LicenseRef-Blockscout
 
 import { chakra, Flex } from '@chakra-ui/react';
-import { getLocalTimeZone, now } from '@internationalized/date';
+import { getLocalTimeZone, now, toCalendarDateTime } from '@internationalized/date';
 import React from 'react';
 import { FormProvider, useForm } from 'react-hook-form';
 
 import type { FormFields } from './types';
+
+import { useSettingsContext } from 'src/shell/top-bar/settings/context';
+import SettingsLocalTime from 'src/shell/top-bar/settings/time-format/SettingsLocalTime';
 
 import { Button } from 'src/toolkit/chakra/button';
 import { DialogBody, DialogContent, DialogHeader, DialogRoot } from 'src/toolkit/chakra/dialog';
@@ -23,13 +26,20 @@ interface Props {
 }
 
 const CsvExportDialog = ({ open, onOpenChange, onFormSubmit, onCancel, children, isAsyncDownload }: Props) => {
+  const settings = useSettingsContext();
+  const isLocalTime = settings?.isLocalTime ?? true;
+
   const formApi = useForm<FormFields>({
     mode: 'onBlur',
-    defaultValues: {
-      // truncated to whole minutes, matching what the picker itself can express
-      from_period: [ now(getLocalTimeZone()).set({ second: 0, millisecond: 0 }).subtract({ days: 1 }) ],
-      to_period: [ now(getLocalTimeZone()).set({ second: 0, millisecond: 0 }) ],
-    },
+    defaultValues: (() => {
+      // the picker holds bare wall-clock values; seed them with "now" read in the zone the toggle
+      // currently selects, truncated to whole minutes to match what the picker can express
+      const currentWallClock = toCalendarDateTime(now(isLocalTime ? getLocalTimeZone() : 'UTC')).set({ second: 0, millisecond: 0 });
+      return {
+        from_period: [ currentWallClock.subtract({ days: 1 }) ],
+        to_period: [ currentWallClock ],
+      };
+    })(),
   });
 
   const { handleSubmit, formState } = formApi;
@@ -46,7 +56,7 @@ const CsvExportDialog = ({ open, onOpenChange, onFormSubmit, onCancel, children,
   }, [ onOpenChange, formState.isSubmitting, onCancel ]);
 
   return (
-    <DialogRoot open={ open } onOpenChange={ handleOpenChange } size={{ lgDown: 'full', lg: 'md' }}>
+    <DialogRoot open={ open } onOpenChange={ handleOpenChange } size={{ lgDown: 'full', lg: 'sm' }}>
       <DialogContent>
         <FormProvider { ...formApi }>
           <chakra.form
@@ -58,16 +68,15 @@ const CsvExportDialog = ({ open, onOpenChange, onFormSubmit, onCancel, children,
             </DialogHeader>
             <DialogBody>
               { children }
+              <SettingsLocalTime id="csv-export-local-time" mt={ 6 } width="calc(100% - 1px)"/>
               <Flex
-                columnGap={ 3 }
                 rowGap={ 3 }
-                mt={ 6 }
-                flexDir={{ base: 'column', lg: 'row' }}
-                alignItems={{ base: 'flex-start', lg: 'center' }}
-                w="100%"
+                mt={ 3 }
+                flexDir="column"
+                alignItems="stretch"
               >
-                <CsvExportFormDateField name="from_period" formApi={ formApi }/>
-                <CsvExportFormDateField name="to_period" formApi={ formApi }/>
+                <CsvExportFormDateField name="from_period" formApi={ formApi } isLocalTime={ isLocalTime }/>
+                <CsvExportFormDateField name="to_period" formApi={ formApi } isLocalTime={ isLocalTime }/>
               </Flex>
               <Button
                 variant="solid"
--- a/src/features/csv-export/components/dialog/CsvExportDialog.tsx
+++ b/src/features/csv-export/components/dialog/CsvExportDialog.tsx
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: LicenseRef-Blockscout
 
 import { chakra, Flex } from '@chakra-ui/react';
-import { getLocalTimeZone, now, toCalendarDateTime } from '@internationalized/date';
+import { getLocalTimeZone, now, toCalendarDateTime, toTimeZone, toZoned } from '@internationalized/date';
 import React from 'react';
 import { FormProvider, useForm } from 'react-hook-form';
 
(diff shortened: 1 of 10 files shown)


src/features/csv-export/components/dialog/CsvExportFormDateField.tsx
```diff
@@ -43,15 +44,23 @@ const CsvExportFormDateField = ({ formApi, name }: Props) => {
     }
   }, [ formState.errors.from_period, formState.errors.to_period, getValues, name, trigger ]);
 
-  const maxDate = React.useMemo(() => parseAbsolute(new Date().toISOString(), getLocalTimeZone()), []);
+  const maxDate = React.useMemo(
+    () => toCalendarDateTime(now(isLocalTime ? getLocalTimeZone() : 'UTC')),
+    [ isLocalTime ],
+  );
```
tom2drum, line 50: **blocker** — Toggling the switch changes `max` to `toCalendarDateTime(now(isLocalTime ? getLocalTimeZone() : 'UTC'))` but the held value keeps its old wall-clock digits (`defaultValues` is seeded once at mount; the `key` remount does not reset RHF state).

East of UTC, an untouched `to_period` (local “now”) is then greater than the UTC `max`, so submit fails with `'Date is after the maximum date'`. West of UTC it passes, but `serializeFormFields` re-reads the same digits in the new zone and the requested window silently shifts by the offset while the input advertises `UTC`.

On `isLocalTime` change, either convert held values instant-preserving (`toZoned(prev, oldZone).toTimeZone(newZone)` → `toCalendarDateTime` via `setValue`/`reset`) or re-seed both fields to the new zone’s “now”. Which of those is intended needs a product call; the current path is wrong for both.

— Reviewed by Cursor Grok 4.6

src/features/csv-export/components/dialog/CsvExportDialog.tsx
```diff
@@ -58,16 +68,15 @@ const CsvExportDialog = ({ open, onOpenChange, onFormSubmit, onCancel, children,
             </DialogHeader>
             <DialogBody>
               { children }
+              <SettingsLocalTime id="csv-export-local-time" mt={ 6 } width="calc(100% - 1px)"/>
```
tom2drum, line 95: **blocker** — This mounts a second `SettingsLocalTime` while the switch is still uncontrolled (`defaultChecked={ isLocalTime }` + `onChange={ toggleIsLocalTime }` in `SettingsLocalTime.tsx`). The top-bar instance in `Settings.tsx` is inside `PopoverRoot` with `lazyMount={ false }`, so it is mounted from page load.

Toggling here updates context but does not move the top-bar switch (its checked state was fixed at mount). That switch then shows the opposite of `isLocalTime`, and its next click flips the context away from what it displays.

Make it controlled: `checked={ isLocalTime }` with `onCheckedChange` (or equivalent) so every instance follows the context.

— Reviewed by Cursor Grok 4.6

src/toolkit/chakra/date-picker.tsx
```diff
@@ -21,8 +21,8 @@ export interface DatePickerValueChangeDetails {
   view: 'day' | 'month' | 'year';
 }
 
-const DATE_FORMAT = 'MMM D, YYYY';
-const DATE_TIME_FORMAT = 'MMM D, YYYY H:mm';
+export const DATE_FORMAT = 'MMM D, YYYY';
```
tom2drum: **nit** — `export const DATE_FORMAT` is Dead Scaffolding: only `DATE_TIME_FORMAT` gained a consumer. `DATE_FORMAT` is still used solely inside this file, and `src/toolkit/package/src/index.ts` re-exports the whole date-picker module, so this becomes published package surface.

Keep `DATE_FORMAT` module-private; export only `DATE_TIME_FORMAT`.

— Reviewed by Cursor Grok 4.6

src/features/csv-export/utils/serialize-form-fields.ts
```diff
@@ -5,20 +5,23 @@ import { getLocalTimeZone, toZoned } from '@internationalized/date';
...
+// absolute timestamps
+const toAbsoluteString = (date: DateValue, isLocalTime: boolean): string => {
+  if ('timeZone' in date) {
+    return date.toAbsoluteString();
+  }
+  return toZoned(date, isLocalTime ? getLocalTimeZone() : 'UTC').toAbsoluteString();
 };
 
-export default function serializeFormFields(data?: FormFields): Record<string, string> {
+export default function serializeFormFields(data?: FormFields, isLocalTime: boolean = true): Record<string, string> {
```
tom2drum: **nit** — `isLocalTime: boolean = true` makes the zone-critical argument optional. Every production call site already passes the flag; the default silently picks local time if a future caller forgets (typescript.md: optional properties / bugs caused by a failure to pass).

Make the parameter required and update the older spec cases to pass it.

— Reviewed by Cursor Grok 4.6

src/features/csv-export/utils/get-file-name.ts
```diff
@@ -25,7 +29,11 @@ export default function getFileName({ type, params, chainConfig }: Params): stri
   }
 
   if (type.startsWith('address_')) {
-    const dateText = params.from_period && params.to_period ? `from_${ params.from_period }_to_${ params.to_period }` : '';
+    const formatPeriod = (isoString: string) =>
+      (isLocalTime ? dayjs(isoString) : dayjs(isoString).utc()).format(PERIOD_FILE_NAME_FORMAT);
+    const dateText = params.from_period && params.to_period ?
+      `from_${ formatPeriod(params.from_period) }_to_${ formatPeriod(params.to_period) }${ isLocalTime ? '' : '_UTC' }` :
```
tom2drum, line 35: **nit** — This new zone-dependent filename logic has no spec coverage, while the equivalent branch in `serializeFormFields` did get a test. A `get-file-name.spec.ts` asserting both local and UTC branches on a fixed ISO input would lock the `_UTC` suffix and the filename-safe format.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `97192e485f50065ccdfea2e057dc36e9437f32fe` → `35a246d127edc569457a19de54d65bba7fbff66a` |
| Round | 1 of 3 |
| Findings | 2 blocker · 0 major · 3 nit |
| By axis | Spec 0 · Standards 2 · Correctness 3 |
| Outcome | `blocked` |

No spec was present for this change, so the spec axis was not run.

The two blockers are the toggle path: held wall-clock values vs a zone-dependent `max`, and a second `SettingsLocalTime` instance while the switch is still uncontrolled.

src/features/csv-export/components/dialog/CsvExportFormDateField.tsx
tom2drum: Fixed in ea9ee9c5e. We reconsidered the design and switched to **instant-preserving** conversion: on toggle, the held From/To values are re-expressed in the new zone (`toTimeZone(toZoned(…))`), so the toggle only changes how the picked moment is shown, never which moment it is. This matches the global "Local time format" setting everywhere else and eliminates both the east-of-UTC max mismatch and the west-of-UTC silent shift.

src/features/csv-export/components/dialog/CsvExportDialog.tsx
tom2drum: Fixed in ea9ee9c5e. `SettingsLocalTime` is now a controlled switch (`checked={ isLocalTime }` + `onCheckedChange`), so the top-bar and CSV-dialog instances both follow the shared context.

src/toolkit/chakra/date-picker.tsx
tom2drum: Fixed in ea9ee9c5e. `DATE_FORMAT` is module-private again; only the time format is exported, renamed to `DATE_PICKER_DATE_TIME_FORMAT` to stay unambiguous as un-namespaced package surface.

src/features/csv-export/utils/serialize-form-fields.ts
tom2drum: Fixed in ea9ee9c5e. `isLocalTime` is now a required parameter; the spec call sites pass it explicitly.

src/features/csv-export/utils/get-file-name.ts
tom2drum: Deferred — valid point, keeping it as a follow-up rather than in this PR. Note the zone-aware filename only applies to the synchronous download path; async exports are named by the backend's `downloadFile` endpoint, which the frontend can't override (cross-origin direct link). A `get-file-name.spec.ts` covering the sync local/UTC branches is still worth adding.

tom2drum (commented):
| | |
| --- | --- |
| Reviewed | `97192e485f50065ccdfea2e057dc36e9437f32fe` → `ea9ee9c5e0a6832eb1f40a1e4ca1a352db454e8a` |
| Round | 2 of 3 |
| By axis | Spec 0 · Standards 0 · Correctness 0 (arbitration) |
| Outcome | `clear` |

Arbitration of Round 1 only — no fresh axis pass.

### Closed this round
- **F1 blocker (zone toggle / max)** — `fixed`. Instant-preserving `toTimeZone(toZoned(…))` in `CsvExportDialog` rewrites held wall-clocks on toggle; east-of-UTC max fail and west-of-UTC silent shift are gone.
- **F2 blocker (uncontrolled switch)** — `fixed`. `checked` + `onCheckedChange` keep top-bar and dialog instances on the shared context bit.
- **F3 nit (`DATE_FORMAT` export)** — `fixed`. `DATE_FORMAT` is private; package surface is `DATE_PICKER_DATE_TIME_FORMAT`.
- **F4 nit (`isLocalTime` default)** — `fixed`. Parameter is required; specs pass it explicitly.

### Still open
- **F5 nit (`get-file-name.spec.ts`)** — `deferred` (coder deferred; accepted). Sync-path coverage remains a follow-up.

src/features/csv-export/utils/get-file-name.ts
tom2drum: **deferred** accepted — leave as follow-up. Sync local/UTC filename branches remain untested; not blocking this PR.

— Reviewed by Cursor Grok 4.5

PR #3659, opened 2026-08-24
Add OP Stack transaction type labels

nonsense:
Adds badges for OP Stack L1 attributes and post-exec transactions.

### DIFF
--- a/src/slices/tx/components/TxType.tsx
+++ b/src/slices/tx/components/TxType.tsx
@@ -7,12 +7,18 @@ import type { schemas } from '@blockscout/api-types';
 import type { BadgeProps } from 'src/toolkit/chakra/badge';
 import { Badge } from 'src/toolkit/chakra/badge';
 
+type TransactionType = schemas['Transaction']['transaction_types'][number] |
+'op_stack_l1_attributes_tx' |
+'op_stack_post_exec_tx';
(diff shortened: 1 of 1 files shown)


src/slices/tx/components/TxType.tsx
```diff
@@ -7,12 +7,18 @@ import type { schemas } from '@blockscout/api-types';
 import type { BadgeProps } from 'src/toolkit/chakra/badge';
 import { Badge } from 'src/toolkit/chakra/badge';
 
+type TransactionType = schemas['Transaction']['transaction_types'][number] |
+'op_stack_l1_attributes_tx' |
+'op_stack_post_exec_tx';
```
tom2drum: First of all, these new transaction types should be implemented on the backend side, and then the new version of the API types package should be published to NPM and consumed in this component. So, this two additions of the union `TransactionType` type will be basically unnecessary.

tom2drum (changes requested):

tom2drum:
@nonsense Could you try updating the `@blockscout/api-types` package to version `v0.0.1-beta.b2cecdf`? It should include the new transaction types.

nonsense:
@tom2drum could you PTAL when you have a chance.

PR #3664, opened 2026-08-25
Surface batch-transaction recipients in the tx details To & Value fields

### LINKED ISSUE #3661: UI/UX changes: tx details page ("to" & "value")
### Changes on the [page ](https://eden-testnet.blockscout.com/tx/0x396c6627171ba4066c37f258f51a5602634aeefa4feb0f20fbd5378e49ab673c?tab=index) 
[FIGMA mockup](https://www.figma.com/design/CEgxqWOzVulwfTUHhs0gUC/%F0%9F%94%A5-RESEARCH-DRAFTs?node-id=5940-4556&t=67GFG4VNHQQvwJdq-4)

**1. There are many addresses in the "to" section.** 
Max 5 addresses + grey link to "Call" section

Image

**2. Value. Add link to "Call" section** 
0.002395904453623692 TIA ($2.55) to <2 recipients>

Image

Image

**3. Update link's style**

Image

tom2drum:
## Description

Resolves #3661

On the transaction details page, Eden **sponsored (batch)** transactions carry a batch of `calls`, each to its own recipient. Previously the **To** field showed only the single top-level `to` address — hiding the other recipients — and the **Value** field showed only the summed amount, with no pointer to the per-recipient breakdown already rendered in the "Calls" section below.

- **To** now lists the batch's **unique** recipients (de-duplicated by address), capped at 5 rows, with a grey **"View all (N)"** link when there are more. The first row keeps the rich `to` rendering (metadata, tags); rows 2+ render as a bare hash (identicon + hash + copy), since `calls[].to` carries no metadata. A batch that resolves to a single distinct recipient keeps the existing single-address rendering.
- **Value** appends a grey `to` word and a blue **"N recipients"** link, e.g. `0.002395904453623692 TIA ($2.55) to 2 recipients`.
- Both links **expand the Details section in place** (revealing the "Calls" breakdown) — no navigation or scroll.
- Separately, the token-transfer detail rows (Tokens transferred / minted / burnt / created) drop the leading icon before **"View all"** and adopt the updated link style; the link's show-condition and Token-transfers-tab target are unchanged.

Behaviour is gated on the presence of `calls` in the response (chain-variant detection), not a feature flag. All consumed fields are already live in production, so there is no backend work. Plan and decisions: [`.agents/tasks/3661-tx-details-to-value-links/spec.md`](https://github.com/blockscout/frontend/blob/issue-3661/.agents/tasks/3661-tx-details-to-value-links/spec.md).

## Environment variables

None.

## Minimum API version

None — the required fields (`calls`, `to`, `value`, `token_transfers_overflow`, `transaction_types`) are already live in production.

None.

## Additional information

Adds a `getBatchRecipients` helper (with unit tests) in the Eden chain-variant feature to derive the distinct-recipient list and count shared by both the "To" list and the "Value"/"To" link labels. Playwright reference snapshots for the token-transfer rows were updated for the restyled "View all" link.


tom2drum (commented):
| --- | --- | --- | --- |
| F1 | major | standards | `TxDetails.tsx:28` |
| F2 | major | spec | `TxDetailsTo.tsx:97` |
| F3 | major | spec | `batch-recipients.ts:21` |

**Counts:** blocker 0 · major 3 · nit 0  
**By axis:** spec 2 · standards 1 · correctness 0  
**Outcome:** findings

Correctness returned empty; F2/F3 are the logic gaps that axis missed, filed under spec because they break FR-1/FR-4/FR-6.

src/slices/tx/pages/details/info/TxDetails.tsx
```diff
@@ -25,6 +25,7 @@ import TxSocketAlert from 'src/slices/tx/components/TxSocketAlert';
 import getConfirmationDuration from 'src/slices/tx/utils/get-confirmation-duration';
 
 import TxDetailsEden from 'src/features/chain-variants/eden/pages/tx/TxDetailsEden';
+import { getBatchRecipients } from 'src/features/chain-variants/eden/utils/batch-recipients';
```
tom2drum, line 28: **F1 · major** — `import { getBatchRecipients } from 'src/features/chain-variants/eden/utils/batch-recipients'` pulls Eden batch UI into the core `tx` slice (`TxDetails.tsx:89` calls it on every chain; `TxDetailsTo` then renders `calls[]`). That breaks `.agents/rules/architecture.md` (“Compose at the page level when optional feature UI plugs into a core slice”) and the spec’s own Implementation decisions (“Changes #1–#2 live in the Eden feature and are composed into `TxDetails` at the slice page level”). The spec’s Components-touched list *also* names `TxDetailsTo` in the slice, so placement is a judgement call — but the util import and recipient-list branch should still live under `features/chain-variants/eden` and be injected, matching the existing `TxDetailsEden` page-level compose.

Move the recipient-list branch (and this helper call) into the Eden feature; keep the slice on the single-address path.

— Reviewed by Cursor Grok 4.6

src/slices/tx/pages/details/info/parts/TxDetailsTo.tsx
```diff
@@ -0,0 +1,122 @@
...
+        isLoading={ isLoading }
+      >
+        { data.to?.is_contract ? 'Interacted with contract' : 'To' }
+      </DetailedInfo.ItemLabel>
+      { recipients.hasMultipleRecipients ? (
+        <DetailedInfo.ItemValue flexDir="column" alignItems="flex-start" rowGap={ 1 }>
+          <Flex flexWrap={{ base: 'wrap', lg: 'nowrap' }} alignItems="center" columnGap={ 3 } maxW="100%" minH={ DetailedInfo.ITEM_VALUE_LINE_HEIGHT }>
+            { toFieldContent }
+          </Flex>
+          { recipients.visibleRecipients.slice(1).map((call, index) => (
```
tom2drum: **F2 · major** — `recipients.visibleRecipients.slice(1)` drops `calls[0]` unconditionally while row 1 is always the top-level `data.to`. FR-1 says the list is unique recipients from `calls[]`; FR-4 assumes row 1 *is* `calls[0]` / top-level `to`. If those hashes differ (batch entrypoint as `to`, or `data.to === null` → `[ Contract creation ]`), one real recipient is never rendered while `count` still includes it, so “View all (N)” / “N recipients” disagree with the rows. Confirm against a live Eden batch whether `data.to.hash` is always the first unique `calls[].to`; if not, render `visibleRecipients` and only substitute the rich `to` UI for the matching hash.

— Reviewed by Cursor Grok 4.6

src/features/chain-variants/eden/utils/batch-recipients.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export interface BatchRecipients {
+  readonly count: number;
+  readonly hasMultipleRecipients: boolean;
+  readonly visibleRecipients: TransactionEdenCalls;
+  readonly hasOverflow: boolean;
+}
+
+export function getBatchRecipients(calls: TransactionEdenCalls | null | undefined): BatchRecipients {
+  const uniqueRecipients = uniqBy(calls ?? EMPTY_CALLS, (call) => call.to);
```
tom2drum: **F3 · major** — `uniqBy(..., (call) => call.to)` keys on `call.to`, which is `string | null` for contract-creation calls (`TxDetailsEden.tsx`: “`to` is `null` for a contract-creation call”). FR-6: “N = number of unique recipient **addresses**”. A null `to` counts as a recipient, can flip `hasMultipleRecipients`, and is rendered as `[ Contract creation ]` at `TxDetailsTo.tsx:101` — a row FR-4 never describes. Filter out null `to` before de-duplicating so `count` and the rows only contain addresses.

— Reviewed by Cursor Grok 4.6

src/slices/tx/pages/details/info/TxDetails.tsx
tom2drum: **reject** — keeping the current placement.

The recipient list and "N recipients" link extend the core **To** and **Value** fields, which exist on every chain; they are a conditional enhancement of core fields, not a separable feature panel like the Eden "Calls" section. The spec sanctions this: its *UI inventory / Components-touched* explicitly names `TxDetails.tsx` and `TxDetailsTo.tsx` in the core `tx` slice with per-file roles (the *Implementation decisions* summary reads the other way — the spec is internally inconsistent on this point, and the concrete file list is the more specific statement).

The architecture rule targets rendering feature *components* into a slice, to avoid cycles and keep features opt-in. Here the only cross-boundary coupling is a single pure, side-effect-free util import (`getBatchRecipients`), gated behind `hasMultipleRecipients` and inert off-Eden. No cycle exists — `features/chain-variants/eden` does not import from `slices/tx`. Confirmed as a keep-as-is decision by the maintainer.

src/slices/tx/pages/details/info/parts/TxDetailsTo.tsx
tom2drum: **fix** — dropped the blind `slice(1)`.

`TxDetailsTo` now maps over all `visibleRecipients` and renders the rich `toFieldContent` for the recipient whose hash matches `data.to?.hash`, and a bare `AddressEntity` for the rest. This no longer assumes row 1 is `calls[0]`/the top-level `to`: the rows shown are always a subset of the counted recipients, so "View all (N)" can no longer disagree with the list even if `data.to.hash` differs from the first `calls[].to`. Combined with F3 (null recipients filtered upstream), the `[ Contract creation ]` row branch is now dead and was removed.

On real data where `data.to.hash === calls[0].to` the output is unchanged. Landing in the next push.

src/features/chain-variants/eden/utils/batch-recipients.ts
tom2drum: **fix** — null recipients excluded before de-duplication.

`getBatchRecipients` now filters out contract-creation calls (`to === null`) before `uniqBy`, so `count` and `visibleRecipients` contain only unique recipient *addresses* per FR-6. `visibleRecipients` is narrowed to `to: string`, which also removes the null-branch handling downstream in `TxDetailsTo`. Added unit tests covering count and visible-list null exclusion. Landing in the next push.

src/slices/tx/pages/details/info/parts/TxDetailsTo.tsx
tom2drum: Fixed in b07440d1e.

src/features/chain-variants/eden/utils/batch-recipients.ts
tom2drum: Fixed in b07440d1e.

src/slices/tx/pages/details/info/TxDetails.tsx
tom2drum: F1 — accepted, the architecture rule is about composing feature *components* into a slice page; this is a pure `getBatchRecipients` util plus a conditional enhancement of the core To/Value fields. Spec Components-touched names the slice files, and the maintainer confirmed keep-as-is.

— Reviewed by Cursor Grok 4.6

src/slices/tx/pages/details/info/parts/TxDetailsTo.tsx
tom2drum: F2 — verified. `visibleRecipients.map` no longer `slice(1)`; rich `toFieldContent` is only used when `call.to === data.to?.hash`. Count and rows stay aligned.

— Reviewed by Cursor Grok 4.6

src/features/chain-variants/eden/utils/batch-recipients.ts
tom2drum: F3 — verified. Null `to` is filtered before `uniqBy`; `visibleRecipients` is narrowed to addresses; unit tests cover the exclusion.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
Review clear

| Id | Ruling |
| --- | --- |
| F1 | rejected-accepted |
| F2 | verified |
| F3 | verified |

**Counts:** blocker 0 · major 0 · nit 0  
**By axis:** spec 0 · standards 0 · correctness 0  
**Outcome:** clear

— Reviewed by Cursor Grok 4.6

PR #3667, opened 2026-08-26
Measure middle-truncated text width via Canvas instead of forcing reflow

### LINKED ISSUE #3666: Refactor TruncateMiddle to measure text width via Canvas instead of forcing reflow
## Description

`TruncateMiddle` computes its middle-ellipsis cut point by appending a hidden `<span>` to
the DOM, setting its text, and reading `getBoundingClientRect().width` repeatedly inside a
binary search. Each width read forces a synchronous layout reflow, and the search performs
~log₂(n) reads per truncated string — multiplied across every truncated cell rendered on a
page (address/hash tables, etc.).

Replace the hidden-span measurement with the Canvas 2D `measureText()` API, which returns
the rendered width of a string in a given font without touching layout.

## Details

- Component: `src/toolkit/components/truncation/TruncateMiddle.tsx`
- The `shadowEl` width reads (the binary-search loop and its `getWidth` helper) are pure
  single-line text-width measurements and are the part to migrate to `measureText()`.
- The container width read (`getWidth(parent)`) is a real box measurement and must stay as a
  DOM read.
- Font correctness: the canvas context must be given the same font family, size, and weight
  the span currently inherits. The component already tracks font readiness via
  `useFontFaceObserver` and has access to `BODY_TYPEFACE` / `HEADING_TYPEFACE` and
  `styleProps.fontWeight`.
- A reusable canvas + context can be cached across calls; the font string should be
  recomputed when the resolved font changes.
- Separately, the `ResizeObserver` currently observes `document.body`, so every truncated
  instance recalculates on any body-level resize. Narrow it to observe the component's own
  parent/container instead, so recalculation is triggered only when the relevant element
  actually changes size.

## Expected Behavior

- Truncation output is visually identical to the current implementation across fonts,
  weights, and container widths.
- No hidden element is appended to the DOM and no per-iteration reflow occurs during cut-point
  calculation.
- Recalculation still happens on container resize and on font load.

tom2drum:
## Description

Resolves #3666

`TruncateMiddle` computed its middle-ellipsis cut point by appending a hidden `<span>` to the DOM and reading `getBoundingClientRect().width` inside a binary search — each read forcing a synchronous layout reflow, ~log₂(n) times per truncated string, multiplied across every truncated cell on a page (address/hash tables, etc.).

The binary search now measures candidate strings with the Canvas 2D `measureText()` API, which returns text width without touching layout — zero per-iteration reflow. The font handed to the canvas is read once per recalculation via `getComputedStyle` on the rendered span, so it reflects the real resolved cascade + `textStyle` + prop overrides (and `letter-spacing`, folded back in since `measureText` returns advance width only). A single offscreen canvas is shared across all instances.

Two deviations from the issue, both found while verifying "visually identical" output:

- **Container-width read still needs the full-length text momentarily.** The old shadow span was doing double duty — appending full text also forced the surrounding shrink-to-fit flex container to expand to its available width before the container box was read. Without that, the parent shrink-wraps the already-truncated string and collapses to the minimum. The span's own text node is now briefly set to the full value for the single container-box read, then restored — no element is appended to the DOM.
- **ResizeObserver stays on `document.body`.** The issue asked to narrow it to the component's own parent, but the truncated span sits inside a shrink-to-fit container that doesn't grow when the viewport widens (only an outer block ancestor does), so a parent-scoped observer never fires on widen and the text stays stuck at its narrowest width.

## Environment variables

None

## Minimum API version

None

None

## Additional information

Truncation output is visually identical to the previous implementation across fonts, weights, and container widths, verified on the design-system page (`maxW`-bounded address links) and the name-services directory table (shrink/widen resize).

### DIFF
--- a/src/toolkit/components/truncation/TruncateMiddle.tsx
+++ b/src/toolkit/components/truncation/TruncateMiddle.tsx
@@ -1,9 +1,11 @@
 // SPDX-License-Identifier: LicenseRef-Blockscout
 
-// Middle-ellipsis truncation driven by container width: measures a hidden shadow node and
-// binary-searches the longest head that fits, keeping a fixed-length tail (0x123...4567).
-// Can't be done in pure CSS — a `text-overflow: ellipsis` head + fixed tail leaves an
-// unremovable gap between the dots and the tail, so it's computed in JS.
+// Middle-ellipsis truncation driven by container width: measures candidate strings with the
+// Canvas 2D API and binary-searches the longest head that fits, keeping a fixed-length tail
+// (0x123...4567). Can't be done in pure CSS — a `text-overflow: ellipsis` head + fixed tail
+// leaves an unremovable gap between the dots and the tail, so it's computed in JS. measureText
+// returns text width without touching layout, so the binary search never forces a reflow; the
+// only DOM read left is the container's own box.
 
 import { chakra } from '@chakra-ui/react';
 import { debounce } from 'es-toolkit';
@@ -15,15 +17,41 @@ import type { TruncateBaseProps } from './types';
 import { Skeleton } from '../../chakra/skeleton';
 import { Tooltip } from '../../chakra/tooltip';
 import { BODY_TYPEFACE, HEADING_TYPEFACE } from '../../theme/foundations/typography';
+import { SECOND } from '../../utils/consts';
 
 const TAIL_LENGTH = 4;
 const HEAD_MIN_LENGTH = 4;
+const DEFAULT_FONT_WEIGHT = '400';
+const RESIZE_DEBOUNCE = SECOND / 10;
 
 export interface TruncateMiddleProps extends TruncateBaseProps {
   tailLength?: number;
 }
 
-const DEFAULT_FONT_WEIGHT = '400';
+// One offscreen canvas is reused across every TruncateMiddle instance: measureText is
+// synchronous and calculateString never yields, so instances can't interleave measurements.
+// `undefined` means "not created yet"; `null` means getContext failed (non-browser env).
+let sharedContext: CanvasRenderingContext2D | null | undefined;
+
+function getMeasureContext(): CanvasRenderingContext2D | null {
+  if (sharedContext === undefined) {
+    sharedContext = document.createElement('canvas').getContext('2d');
+  }
+  return sharedContext;
+}
+
+// Canvas needs at least weight + size + family; getComputedStyle(el).font returns '' in Chrome,
+// so the shorthand is assembled from the individual resolved properties instead.
+function getFontString(styles: CSSStyleDeclaration): string {
+  return `${ styles.fontStyle } ${ styles.fontWeight } ${ styles.fontSize } ${ styles.fontFamily }`;
+}
+
+// measureText returns advance width only, so any letter-spacing (added after every glyph) has to
+// be folded back in by hand. Resolves to 0 for the usual `normal`.
+function getLetterSpacing(styles: CSSStyleDeclaration): number {
+  const value = parseFloat(styles.letterSpacing);
+  return Number.isNaN(value) ? 0 : value;
+}
 
 export const TruncateMiddle = React.memo(({
   value,
@@ -46,19 +74,34 @@ export const TruncateMiddle = React.memo(({
   ]);
 
   const calculateString = useCallback(() => {
-    const parent = elementRef?.current?.parentNode as HTMLElement;
-    if (!parent) {
+    const element = elementRef.current;
+    const parent = element?.parentNode as HTMLElement | null;
+    if (!element || !parent) {
       return;
     }
 
-    const shadowEl = document.createElement('span');
-    shadowEl.style.opacity = '0';
-    parent.appendChild(shadowEl);
-    shadowEl.textContent = value;
+    const context = getMeasureContext();
+    if (!context) {
+      setDisplayedString(value);
+      return;
+    }
 
+    // Read the fully-resolved font off the rendered span (cascade + textStyle + overrides), once
+    // per recalculation — the font is invariant across the binary search below.
+    const styles = window.getComputedStyle(element);
+    context.font = getFontString(styles);
+    const letterSpacing = getLetterSpacing(styles);
+    const measure = (text: string) => context.measureText(text).width + letterSpacing * text.length;
+
+    // A shrink-to-fit container only exposes its available width while the span holds full-length
+    // content, so briefly fill the span before reading the container box. Written straight to the
+    // node (not through state) so the read lands in the same synchronous frame — no paint, no
+    // extra DOM node.
+    element.textContent = value;
     const parentWidth = getWidth(parent);
 
-    if (getWidth(shadowEl) > parentWidth) {
+    let result = value;
+    if (measure(value) > parentWidth) {
       const tail = value.slice(-tailLength);
       let leftI = HEAD_MIN_LENGTH;
       let rightI = value.length - tailLength;
@@ -66,19 +109,19 @@ export const TruncateMiddle = React.memo(({
       while (rightI - leftI > 1) {
         const medI = ((rightI - leftI) % 2) ? leftI + (rightI - leftI + 1) / 2 : leftI + (rightI - leftI) / 2;
         const res = value.slice(0, medI) + '...' + tail;
-        shadowEl.textContent = res;
-        if (getWidth(shadowEl) < parentWidth) {
+        if (measure(res) < parentWidth) {
           leftI = medI;
         } else {
           rightI = medI;
         }
       }
-      setDisplayedString(value.slice(0, rightI - 1) + '...' + tail);
-    } else {
-      setDisplayedString(value);
+      result = value.slice(0, rightI - 1) + '...' + tail;
     }
 
-    parent.removeChild(shadowEl);
+    // Restore before React reconciles: setDisplayedString bails out when the value is unchanged,
+    // which would otherwise leave the full-length text written above on screen.
+    element.textContent = result;
+    setDisplayedString(result);
   }, [ value, tailLength ]);
 
   // we want to do recalculation when isFontFaceLoaded flag is changed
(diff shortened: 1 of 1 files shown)


tom2drum (commented):
No spec for this PR — standards + correctness only.

| --- | --- | --- | --- |
| F1 | major | correctness | `TruncateMiddle.tsx:118` |
| F2 | nit | standards | `TruncateMiddle.tsx:7` |
| F3 | nit | standards | `TruncateMiddle.tsx:98` |

**Counts:** blocker 0 · major 1 · nit 2. Spec — (no spec) · standards 2 · correctness 1.

**Outcome:** findings.

src/toolkit/components/truncation/TruncateMiddle.tsx
```diff
@@ -46,39 +74,54 @@ export const TruncateMiddle = React.memo(({
...
+        if (measure(res) < parentWidth) {
           leftI = medI;
         } else {
           rightI = medI;
         }
       }
-      setDisplayedString(value.slice(0, rightI - 1) + '...' + tail);
-    } else {
-      setDisplayedString(value);
+      result = value.slice(0, rightI - 1) + '...' + tail;
```
tom2drum: **F1 · major** — `result = value.slice(0, rightI - 1) + '...' + tail` can emit a string *longer* than `value` when the overflow branch runs but the search never iterates.

`leftI` starts at `HEAD_MIN_LENGTH` (4). If `value.length <= tailLength`, `rightI` is ≤ 0, so `while (rightI - leftI > 1)` is skipped and `slice(0, rightI - 1)` uses a negative end index. For `value = 'abcd'` and `tailLength = 4`: `rightI = 0` → `'abcd'.slice(0, -1)` → `'abc...abcd'`.

The canvas `measureText` path makes this easier to hit than the old shadow span: unclamped advance width will enter the overflow branch on a narrow parent even for short strings (`LogTopic` numeric/`text` topics, `BlockEntity` with `tailLength ?? 2`).

Bail before the search when there is no room for a head (`value.length - tailLength <= HEAD_MIN_LENGTH`) and keep `result = value`.

— Reviewed by Cursor Grok 4.6

src/toolkit/components/truncation/TruncateMiddle.tsx
```diff
@@ -1,9 +1,11 @@
...
-// Middle-ellipsis truncation driven by container width: measures a hidden shadow node and
-// binary-searches the longest head that fits, keeping a fixed-length tail (0x123...4567).
-// Can't be done in pure CSS — a `text-overflow: ellipsis` head + fixed tail leaves an
-// unremovable gap between the dots and the tail, so it's computed in JS.
+// Middle-ellipsis truncation driven by container width: measures candidate strings with the
+// Canvas 2D API and binary-searches the longest head that fits, keeping a fixed-length tail
+// (0x123...4567). Can't be done in pure CSS — a `text-overflow: ellipsis` head + fixed tail
+// leaves an unremovable gap between the dots and the tail, so it's computed in JS. measureText
+// returns text width without touching layout, so the binary search never forces a reflow; the
+// only DOM read left is the container's own box.
```
tom2drum: **F2 · nit** — `the only DOM read left is the container's own box` is diff-narration (`.agents/rules/code-quality.md` → Comments: if a comment would confuse a reader who never saw the previous version, delete it). It is also wrong: `getComputedStyle(element)` on line 91 is another read.

Keep `measureText returns text width without touching layout, so the binary search never forces a reflow.` Drop the trailing clause.

— Reviewed by Cursor Grok 4.6

src/toolkit/components/truncation/TruncateMiddle.tsx
```diff
@@ -46,39 +74,54 @@ export const TruncateMiddle = React.memo(({
...
+    // per recalculation — the font is invariant across the binary search below.
+    const styles = window.getComputedStyle(element);
+    context.font = getFontString(styles);
+    const letterSpacing = getLetterSpacing(styles);
+    const measure = (text: string) => context.measureText(text).width + letterSpacing * text.length;
+
+    // A shrink-to-fit container only exposes its available width while the span holds full-length
+    // content, so briefly fill the span before reading the container box. Written straight to the
+    // node (not through state) so the read lands in the same synchronous frame — no paint, no
+    // extra DOM node.
```
tom2drum: **F3 · nit** — `no paint, no extra DOM node` describes the absence of the removed shadow span — same diff-narration rule. The load-bearing why (state would not land before the box read) is already in the sentence before it.

End the comment at `so the read lands in the same synchronous frame.`

— Reviewed by Cursor Grok 4.6

src/toolkit/components/truncation/TruncateMiddle.tsx
tom2drum: Fixed in `3e6ac3152`. Added the `value.length - tailLength > HEAD_MIN_LENGTH` guard so the overflow branch no longer runs when there's no room for a head — the full value is kept instead of the garbled longer string.

src/toolkit/components/truncation/TruncateMiddle.tsx
tom2drum: Fixed in `3e6ac3152`. Dropped the trailing clause — kept only `measureText returns text width without touching layout, so the binary search never forces a reflow.`

src/toolkit/components/truncation/TruncateMiddle.tsx
tom2drum: Fixed in `3e6ac3152`. Ended the comment at `so the read lands in the same synchronous frame.` — dropped the diff-narration clause.

src/toolkit/components/truncation/TruncateMiddle.tsx
tom2drum: F1 — verified

src/toolkit/components/truncation/TruncateMiddle.tsx
tom2drum: F2 — verified

src/toolkit/components/truncation/TruncateMiddle.tsx
tom2drum: F3 — verified

tom2drum (commented):
Review clear

Arbitration: F1–F3 verified. No regressions from the fixes.

**Counts:** blocker 0 · major 0 · nit 0. Spec — (no spec) · standards 0 · correctness 0.

**Outcome:** clear

— Reviewed by Cursor Grok 4.6

tom2drum:
## Measured effect

Benchmarked this branch's `TruncateMiddle` against `main`'s, using a Playwright CT harness that renders address-hash rows in the DOM shape `AddressEntity` produces (flex container `minWidth: 0` → inline-flex link → shrink-to-fit `overflow: hidden` box → the truncated span), so the shrink-to-fit container read both implementations depend on is exercised.

Both implementations run in the **same build and browser process**, with arm order alternated — a `main`-vs-branch checkout comparison would fold build flavor and machine state into the delta.

### Scale

**20 instances.** That's the realistic ceiling, not a round number: list pages lazy-render ~15–20 rows, most use static truncation, and the tables that do use dynamic truncation carry at most one such string per row. An earlier run at 150 instances overstated the effect by ~7.5×.

### Per-instance mechanism (scale-independent)

Exact count, via a patched `Element.prototype.getBoundingClientRect`:

| | layout-forcing reads / instance |
|---|---|
| `main` | 7.00 |
| this branch | 4.00 |

The binary search no longer touches layout, but the container-width read and the settle loop still do — hence 4, not 0.

### Counters — CDP `Performance.getMetrics()`

| scenario | metric | `main` | branch | delta |
|---|---|---|---|---|
| mount | LayoutCount | 247 | 125 | **−49%** |
| mount | RecalcStyleCount | 249 | 166 | −33% |
| resize burst | LayoutCount | 976 | 524 | **−46%** |
| resize burst | RecalcStyleCount | 973 | 660 | −32% |

### Durations — interleaved, single page load, CPU throttled 4×

12 phases per arm, 8 resizes per phase:

| metric | `main` | branch | delta | `main` IQR | branch IQR |
|---|---|---|---|---|---|
| LayoutDuration | 95.2ms | 52.3ms | **−45%** | 90.2–107.3 | 51.4–57.8 |
| RecalcStyleDuration | 845ms | 407ms | −52% | 747–1331 | 356–644 |
| TaskDuration | 1229ms | 780ms | −37% | 1169–1711 | 742–897 |
| ScriptDuration | 137ms | 146ms | +6% | 130–194 | 132–233 |

No IQR overlap on any row, so the deltas clear the noise. The +6% `ScriptDuration` is `measureText` doing the work layout used to do — a good trade, and total `TaskDuration` still drops 37%.

## Estimated effect

Forced style+layout work per user-visible event, de-throttled:

- **initial render of a table:** ~46ms → ~32ms, **saves ~14ms**
- **one resize:** ~29ms → ~14ms, **saves ~15ms**

Roughly **one dropped frame's worth of blocking work per page render and per resize** at this scale. Real and consistent, but worth stating plainly: it's a frame, not a page-load transformation. The honest headline is the ratio — about half the forced style+layout work of the previous implementation, per truncated string, at any instance count.

Note that `RecalcStyleDuration` dominates `LayoutDuration` in both arms: a forced synchronous layout recalculates style first, and Chakra's generated stylesheet is large. So the cost being cut is forced *style+layout*, not layout alone — which is why the saving is bigger than "we removed some `getBoundingClientRect` calls" suggests.

## Confidence and caveats

- **A null test guards the harness.** Replacing this branch's component with `main`'s makes both arms identical code; counters then collapse to +0.4% (mount) / +2.5% (resize) and forced reads to 7.00 vs 7.00.
- **That same null test invalidated a first attempt at timings.** Comparing durations across separate Playwright tests drifted −24% on `LayoutDuration` and −19% on `TaskDuration` with identical code in both arms — pure cross-page-load machine drift, which at 20 instances swamps the signal. Hence the interleaved scenario (all phases in one page load, arms alternating ABBA, within-arm IQR reported as the noise floor). Mount *durations* still come from the weaker separate-run design, so the solid mount result is the count.
- **Absolutes are inflated.** Playwright CT is a Vite dev build with unminified React. Production will be somewhat smaller; a low-end device considerably larger. Ratios and counters transfer; milliseconds should be read as order-of-magnitude.

## Possible follow-up

Instances don't batch: each writes `textContent` then reads, so N instances still force ~N layouts minimum. Batching all instances into a read phase and then a write phase — plus trimming the settle loop — is where the remaining headroom is, and would take the 4 reads/instance closer to 1.

---

The harness (`BenchGrid.tsx`, `TruncateMiddleBench.pw.tsx`, `aggregate.mjs`, a verbatim copy of `main`'s component, and a README covering the methodology) is **not part of this PR** — it lives outside the branch. Happy to commit it under `src/toolkit/components/truncation/__bench__/` if we want the comparison reproducible, though its `.pw.tsx` would need excluding from the default Playwright projects.

PR #3670, opened 2026-08-26
Add a cognitive-complexity and CRAP-score CI gate

### LINKED ISSUE #3663: Add a cyclomatic-complexity and CRAP-score CI gate
## Description

Add a static analysis tool that computes cyclomatic complexity and the CRAP
score (`c² × (1 − cov)³ + c`) per function, and can gate CI on a configurable
threshold. The goal is to flag code that is both complex and under-tested before
it lands.

## Details

Findings from an initial investigation:

- The vitest coverage report (`@vitest/coverage-v8`) does not emit complexity or
  CRAP — it only reports statement/branch/function/line coverage. CRAP must be
  computed by joining complexity with coverage.
- Existing JS complexity libraries (`typhonjs-escomplex`, `escomplex`,
  `ts-complex`) are unmaintained (last published 2022) and parse a JavaScript
  AST, so they fail on modern TypeScript syntax. They are not viable.
- A custom tool built on the `typescript` compiler API (already a dependency) is
  the recommended path. Cyclomatic complexity is a count of decision points
  (`if`/`for`/`while`/`do`/`case`/`catch`/ternary/`&&`/`||`/`??`) over the syntax
  tree — no type-checker needed.

Performance (measured on the current tree, ~2,800 files / ~195k LOC):

- Syntax parse + AST walk of the entire repo: ~0.6s in-process (vs. ~20–60s for
  a full `tsc` type-check, which is not required here).
- Coverage collection adds ~18% to the test phase only; it is effectively a free
  rider on the coverage run CI already produces.

## Proposed Approach

- A `tools/complexity/` CLI that:
  1. Walks changed (or all) TS/TSX files via `ts.createSourceFile`, counting
     decision points per function.
  2. Reads a whole-suite istanbul/v8 `coverage-final.json` and joins per-function
     coverage by line range.
  3. Emits per-function CRAP and fails when any function exceeds a configurable
     threshold (a common default is 30).
- Prefer whole-suite coverage over per-file scoping: a changed shared util may be
  covered by tests elsewhere, so scoping coverage to the co-located spec would
  undercount coverage and inflate CRAP, causing false failures.
- Baseline existing violations so the gate only fails on newly introduced
  complexity/CRAP regressions.

## Additional Context

Document the counting conventions chosen (e.g. whether `?.` counts, per-`case`
vs. per-`switch`) so the numbers are comparable to ESLint's `complexity` rule.
SonarQube/SonarCloud quality gates on new code were considered as an
off-the-shelf alternative if maintaining a bespoke tool is undesirable.

tom2drum:
## Description

Resolves #3663

Adds a CI gate that flags code which is either hard to read *or* both complex and under-tested, before it merges. A bespoke `tools/code-complexity/` CLI walks the TypeScript AST (`ts.createSourceFile`, no type-checker) and scores every function twice on a single pass:

- **Cognitive Complexity** — the readability gate. Follows the SonarSource model (flat control flow is cheap, nesting is penalised progressively, boolean runs collapse), capped per function class.
- **CRAP** — the under-testedness gate. `c²·(1 − cov)³ + c`, from *cyclomatic* complexity joined with per-function line coverage out of a v8/istanbul `coverage-final.json`.

The gate is scoped to functions a changed line falls within, against `origin/main`, so it never blocks on pre-existing debt and needs no committed baseline. It runs as a post-step in the existing `vitest_tests` job, consuming the coverage that job now emits — CI regenerates nothing and carries no threshold numbers, so a local run and a CI run gate identically. Locally, `pnpm test:code-complexity <path>` reproduces a failure on one file with no diff and no coverage pre-step.

**Why cognitive complexity and not cyclomatic.** The issue proposed gating cyclomatic directly, and the first pass did. Cyclomatic counts independent paths ≈ tests needed, which makes it the right CRAP input but a poor decomposition signal: it scores a flat `switch` the same as an equivalent nested `if` chain, so a cap tight enough to catch nested code also flagged wide-but-shallow logic that read fine. Cognitive complexity separates the two, and this implementation charges nesting **quadratically** (`1 + nesting²`) rather than SonarSource's linear `1 + nesting` — identical at depths 0 and 1, biting only from the third level in. That is what lets one cap forgive breadth and still catch depth. Cyclomatic survives only as the CRAP input and is not a gate.

**Why components are gated differently.** Every function is classified from the AST, not its file extension (~78 non-test `.tsx` files contain no JSX, mostly `useX.tsx` hooks): `jsx` for a render body, `behavior` for handlers, hooks and utils wherever they live — including nested inside a component. Only `behavior` functions are CRAP-scored; a render body is covered, if at all, by Playwright visual tests that emit no vitest coverage, so scoring it would flood the report with 0%s. Its cognitive cap is the backstop instead. So a component's render loses a meaningless CRAP score while the handlers inside it gain a real one.

Two other decisions worth surfacing: the gate scores **its own implementation** (`tools/` is in scope, so changing the counting model re-scores `complexity.ts` under the new model), and `deploy/**` stays out — it sits outside ESLint and the root TypeScript project, so this gate should not be its first automated check (#3675).

Spec: [`.agents/tasks/3663-code-complexity-ci-gate/spec.md`](.agents/tasks/3663-code-complexity-ci-gate/spec.md). The tool documents itself in [`tools/code-complexity/CONTEXT.md`](tools/code-complexity/CONTEXT.md) (what it is, triage when it fails), with `SCORING.md` for the counting model and its divergences, `CALIBRATION.md` for the caps, and `adr/` for the two counting decisions.

## Environment variables

None.

## Minimum API version

None.

None for deployments. On PRs the `vitest_tests` job now also runs with coverage and gains a gate step that can fail the build; a failure names the cap that broke and, for cognitive violations, the increment sites that cost the most plus what flattening the deepest nesting pocket would save.

## Additional information

**Calibration.** The caps come from a full-repo run under the shipped counting model (`behavior` n=4273, p99=10, max 43; `jsx` n=2265, p99=20, max 92):

| Cap | Value | Fires on | Basis |
| --- | --- | --- | --- |
| cognitive, `behavior` | 20 | 8 functions | a real gap in the distribution — flat-breadth tops out at 20, the nested band starts at 21 |
| cognitive, `jsx` | 25 | 16 render bodies | a smooth shoulder, so a judgment call about where "decompose this" starts |
| CRAP, `behavior` | 80 | — | at 0% coverage CRAP is `c² + c`, so 80 flags untested logic from cyclomatic 9 up |

The gap behind the `behavior` cap exists *because* nesting is quadratic and does not survive a return to linear. The retired cyclomatic gate fired on 43 functions against these caps' 24; the three biggest rescues dropped from cyclomatic 18/14/15 to cognitive 6/2/0. `CALIBRATION.md` records the one-step-tighter and one-step-looser outcomes and the procedure to re-derive them.

**Two known offenders are left failing rather than exempted**, both in `.mjs` tooling scripts and tracked in #3674: `checkFile` in `tools/scripts/check-doc-links.mjs` (cognitive 58) and `replayOperations` in `tools/profiling/aggregate-react-profile.mjs` (CRAP 182 — cyclomatic 13 at 0% coverage). Neither is touched by this diff, so the diff-scoped gate stays green.

**Validation.** The counting model was diffed against `eslint-plugin-sonarjs`'s `cognitive-complexity` rule during calibration; the plugin was installed temporarily and is *not* a committed dependency. That confirmed the shape of the model, not numeric parity — nesting-heavy code reads higher here by design.

**Residual risk.** A changed file exercised only through a dynamic import or another path the module graph misses would be undercounted and could fail falsely. Rare here, and the diff scope plus touched-function limit contain the blast radius.


tools/code-complexity/index.ts
```diff
@@ -0,0 +1,337 @@
...
+// return null instead of running it for numbers that would be discarded.
+function resolveCoverage(options: CliOptions, request: CoverageRequest, anyNeedsCoverage: boolean): CoverageData | null {
+  switch (options.coverageMode) {
+    case 'off':
+      return null;
+    case 'file':
+      // coverageFile is always set when the mode is 'file' (parseArgs reads its value).
+      return readCoverage(options.coverageFile as string);
+    case 'generate':
+      return anyNeedsCoverage ? generateCoverage(request, { verbose: options.verbose }) : null;
```
tom2drum: **F1 · major** — `return anyNeedsCoverage ? generateCoverage(request, { verbose: options.verbose }) : null` collapses a skipped vitest run into `coverage === null`, which `analyze.ts` treats as `--no-coverage` (`if (fn.containsJsx || coverage === null) return null`). FR5: generation scope "decides only whether the tool's vitest run pulls the file in, not which functions are scored." CONTEXT.md: a `behavior` function in a spec-less JSX file "scores 0% without triggering a run." Local generate (`--changed` / focused) therefore never CRAP-scores nested handlers; CI `--coverage-file` does.

In generate mode, skip vitest with empty coverage (`parseCoverage('{}')`), not `null`. Keep `null` for `--no-coverage` only.

— Reviewed by Cursor Grok 4.6

tools/code-complexity/index.ts
```diff
@@ -0,0 +1,337 @@
...
+// 'off'. `anyNeedsCoverage` short-circuits generation: when no file in the selection needs coverage
+// generated (see fileNeedsCoverage) there is nothing coverage could inform, so we skip vitest and
+// return null instead of running it for numbers that would be discarded.
+function resolveCoverage(options: CliOptions, request: CoverageRequest, anyNeedsCoverage: boolean): CoverageData | null {
+  switch (options.coverageMode) {
+    case 'off':
+      return null;
+    case 'file':
+      // coverageFile is always set when the mode is 'file' (parseArgs reads its value).
+      return readCoverage(options.coverageFile as string);
```
tom2drum: **F2 · major** — `return readCoverage(options.coverageFile as string)` has no missing-file path. `generateCoverage` already does `fs.existsSync(file) ? readCoverage(file) : parseCoverage('{}')` because "a run that covered nothing leaves no file." CI's `vitest --changed --coverage --passWithNoTests` can take that path (no matching specs — a `*.pw.tsx`-only change is the usual case), then this throw fails the job instead of scoring absent behavior functions at 0%.

If the coverage file is missing, use `parseCoverage('{}')`, same as `generateCoverage`. Do not throw ENOENT.

— Reviewed by Cursor Grok 4.6

.agents/AGENTS.md
```diff
@@ -56,24 +56,36 @@ Some directories have a `CONTEXT.md` documenting non-obvious patterns specific t
...
+See the global ADRs below. The module-local CONTEXT.md file points to that module’s local ADRs:
 
 - `0002-layer-shaped-ticket-leaves.md` — why a product task's tickets cut vertically while the leaves inside them run along layers.
 - `0003-turbopack-for-production-builds.md` — why production builds moved back to Turbopack.
 
-Add a new record (next free number, and a line here) whenever a decision is expensive to rediscover:
-it constrains future work, was reached by measurement or an investigation worth not repeating, or
-looks wrong without its context. Supersede rather than rewrite — flip the old record's `Status` to
-`superseded by <n>` and leave its reasoning intact.
+### When to offer an ADR
```
tom2drum: **F4 · major** — `### When to offer an ADR` plus "All three of these must be true" is how-to in always-loaded context. prose-smells.md: "`AGENTS.md` carries pointers, never mechanics." Possible Divergent Change — ADR policy is not "what exists, where do I read it?" The rewrite also dropped "and a line here" for the global list.

Keep locations + the numbered list. Move the checklist out (or restore the old one-liner, including "add a line here if repo-wide").

— Reviewed by Cursor Grok 4.6

tools/code-complexity/index.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      maxCognitiveJsx: CAP,
+      maxCognitiveBehavior: DEFAULT_MAX_COGNITIVE_BEHAVIOR,
+    });
+    expect(parseArgs([ '--max-cognitive-behavior', String(CAP) ])).toMatchObject({
+      maxCognitiveJsx: DEFAULT_MAX_COGNITIVE_JSX,
+      maxCognitiveBehavior: CAP,
+    });
+  });
+
+  // The prefix-shadowing case the old startsWith chain only survived by arm order.
```
tom2drum: **F5 · nit** — `// The prefix-shadowing case the old startsWith chain only survived by arm order.` code-quality.md Comments: "Never write *diff-narration* (what this change replaces)." The test name already states the behavior.

Delete the comment.

— Reviewed by Cursor Grok 4.6

tools/code-complexity/adr/0002-quadratic-nesting-and-null-coalescing-in-cognitive-complexity.md
```diff
@@ -0,0 +1,70 @@
...
+`a ? x : b ? y : z` so the chain reads as one decision ladder. An eslint rule already bans nested
+ternaries in this repo, so any nested ternary that survives review is genuine nesting and should be
+priced as such.
+
+## Consequences
+
+- **CC here is not the SonarSource number, and is not expected to be.** Comparing against
+  `eslint-plugin-sonarjs` will show us reading higher on anything nested past depth 1. That is by
+  design; oracle parity is not a regression test. The counting conventions and the full list of
+  divergences live in `../CONTEXT.md`.
```
tom2drum: **F6 · nit** — "The counting conventions and the full list of divergences live in `../CONTEXT.md`." Same pointer in `adr/0001` Consequences. Counting is in `SCORING.md`; CONTEXT.md only points there. Possible Sediment — a line true of a previous design.

Point at `../SCORING.md` (CONTEXT.md already links it). Same fix in `adr/0001`.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
| ID | Severity | Axis | Location |
| --- | --- | --- | --- |
| F1 | major | spec / correctness | `tools/code-complexity/index.ts:204` |
| F2 | major | correctness | `tools/code-complexity/index.ts:202` |
| F3 | major | correctness | `.github/workflows/checks.yml:16` |
| F4 | major | standards | `.agents/AGENTS.md:78` |
| F5 | nit | standards | `tools/code-complexity/index.spec.ts:88` |
| F6 | nit | standards | `tools/code-complexity/adr/0002-….md:60` |

## Not anchorable

**F3 · major** — `paths-ignore: - 'tools/**'` (line 16) skips the whole Checks workflow for tools-only PRs. FR8 allowlists `tools/` so the gate scores its own implementation; CONTEXT.md says the same. A tools-only change therefore never runs `vitest_tests`, so the new post-step never gates `tools/code-complexity/**`. This line is pre-existing and outside the hunk that added the gate.

Stop ignoring `tools/**` on `pull_request`, or run this job when `tools/**` changes (expect other `tools/` files to fail until exempted in `scope.ts`). Trade-off is real: dropping the ignore also runs eslint/tsc/Playwright on tools-only PRs.

— Reviewed by Cursor Grok 4.6

tools/code-complexity/index.ts
tom2drum: **fix** — 4268755a0.

Confirmed against the code: `analyze.ts:37` reads `coverage === null` as "the CRAP gate is off", so short-circuiting the vitest run silently disabled CRAP for the whole selection. Two consequences beyond the FR5 wording — generation scope was deciding *which functions are scored*, and a function's score depended on what else happened to share its selection (a spec-less `Foo.tsx` scored its handlers 0% when a `.ts` file was also in the diff, `—` when it was alone).

Generate mode now returns `parseCoverage('{}')` when the run is skipped; `null` means `--no-coverage` only. The fix lands on the path `analyze.spec.ts:74` already covers (empty coverage + `missingCoverageIsZero` → 0%, flagged).

Also rewrote the CONTEXT.md bullet at "Vitest is skipped when nothing needs coverage" — its "returns instantly with the cognitive gate only" was documenting the bug.

tools/code-complexity/index.ts
tom2drum: **fix** — 4268755a0.

Confirmed, and the asymmetry you point at was the tell: `generate-coverage.ts` already guarded this exact case ("a run that covered nothing leaves no file") and the `--coverage-file` reader did not. CI runs `vitest --changed=origin/main --coverage --passWithNoTests` and then unconditionally `--coverage-file coverage/coverage-final.json`, so a diff matching no vitest spec killed the gate step on ENOENT instead of scoring the changed behavior functions at 0% — the case the gate exists to catch.

Extracted the guard into `readCoverageOrEmpty` in `coverage.ts`; both readers now go through it, so the two paths can't drift again. Two specs added in `coverage.spec.ts`.

.agents/AGENTS.md
tom2drum: **fix** — 4268755a0.

Both points taken.

On the mechanics: agreed the three-item checklist is a *how* in always-loaded context, and `prose-smells.md` is explicit that a how there is misfiled by construction. Rather than restoring the old one-liner, the checklist (plus the numbering convention and the supersede rule) moved to a new `.agents/rules/adr.md`, scoped `paths: **/adr/*.md` so it loads when someone is actually in an `adr/` folder rather than in every session. Cursor symlink added per the rules contract in `.agents/README.md`.

On the dropped "and a line here": a real regression, independently of the above — the list had no instruction left telling anyone to append to it. The rule file's "Where a record goes" table now carries it, and AGENTS.md keeps the locations, the index and a pointer. Section is down from 22 lines to 8.

tools/code-complexity/index.spec.ts
tom2drum: **fix** — 4268755a0. Comment deleted; the test name carries it.

One note for the arbitration round: `index.ts:71-72` also mentions the `startsWith` prefix-shadowing hazard, and was deliberately left. It argues *why the lookup table is the design* against a general alternative, and never references a previous version of this file — so it isn't diff-narration under the same rule. Flagging it since it greps the same way.

tools/code-complexity/adr/0002-quadratic-nesting-and-null-coalescing-in-cognitive-complexity.md
tom2drum: **fix** — 4268755a0. Both ADRs now point at `../SCORING.md`, which is where `CONTEXT.md:15-17` delegates the counting and the SonarSource/ESLint divergences to. `0001` fixed alongside `0002`.

tom2drum:
### F3 — `fix`, 4268755a0

Replying here because F3 was posted under **Not anchorable** in the review body and has no thread of its own.

Confirmed, and consequential: FR8 puts `tools/` in scope *"so the gate also scores its own implementation"*, and `tools/code-complexity/CONTEXT.md` repeats that as a promise — but a tools-only PR skipped the whole Checks workflow, so the gate never scored itself, and `tools/` was never linted or type-checked either. Pre-existing line, but this PR is what makes the claim.

Dropped `- 'tools/**'` from `paths-ignore`.

On the trade-off you flagged: it's real but cheaper than it looks. eslint, tsc and cspell already run green over these same `tools/` files on *this* PR — it touches `package.json`, which isn't ignored — so nothing latent is being uncovered and no `scope.ts` exemptions turned out to be needed. And the gate is diff-scoped, so a tools-only PR only ever gates the functions it touched, not the rest of `tools/`. The remaining cost is CI minutes on future tools-only PRs, which the developer accepted.

tools/code-complexity/index.ts
tom2drum: F1 — verified

tools/code-complexity/index.ts
tom2drum: F2 — verified

.agents/AGENTS.md
tom2drum: F4 — verified

tom2drum:
F3 — verified

tools/code-complexity/adr/0002-quadratic-nesting-and-null-coalescing-in-cognitive-complexity.md
tom2drum: F6 — verified

tools/code-complexity/index.spec.ts
tom2drum: F5 — verified

.agents/rules/adr.md
```diff
@@ -0,0 +1,40 @@
+---
+description: ARDs rules
```
tom2drum: **F7 · nit** — `description: ARDs rules` misspells ADRs in the rule file this commit added.

Change to `description: ADR rules`.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
Review clear

| ID | Severity | Ruling |
| --- | --- | --- |
| F1 | major | verified |
| F2 | major | verified |
| F3 | major | verified |
| F4 | major | verified |
| F5 | nit | verified |
| F6 | nit | verified |
| F7 | nit | open — regression from F4 |

**Counts:** blocker 0 · major 0 · nit 1 open (F7 deferred)
**Axes this round:** arbitration (fixes only)
**Outcome:** clear

— Reviewed by Cursor Grok 4.6

PR #3678, opened 2026-09-01
Measure test assertion strength with mutation testing

### LINKED ISSUE #3665: Add diff-scoped mutation testing (Stryker) in CI and mutation-thinking to review-changes
## Description

Introduce mutation testing to measure whether tests actually assert behavior
(not just execute lines). Two complementary deliverables: a diff-scoped
StrykerJS job in CI, and a lightweight "mutation-thinking" prompt added to the
`review-changes` skill's correctness axis. This is especially relevant because
much of the suite relies on snapshot tests, which are weak at catching
behavioral regressions.

## Details

- Tooling: StrykerJS with `@stryker-mutator/vitest-runner` (first-class vitest
  support). Not currently a dependency.
- Mutation testing is expensive by nature — each mutant reruns its covering
  tests — so whole-repo runs are not viable (hours). All integration must be
  diff-scoped.

### CI job (real mutation testing)

- Scope to changed files with `--since=<base>`.
- Use `--incremental` with a committed `reports/stryker-incremental.json` so
  unchanged mutants are not re-run across pushes.
- Start report-only (post the mutation score, do not fail the build).
- Once trusted, enable a threshold gate (`thresholds.break`) on the diff's
  mutation score.

### review-changes skill (no new infra)

- Add a "mutation-thinking" instruction to the existing correctness axis: for
  each changed assertion, reason about whether flipping a condition or changing
  a return value would cause any test in the diff to fail, and flag tests that
  would survive such a mutation.
- This runs as LLM reasoning, not execution — it keeps the skill fast and
  findings-only, and targets the biggest offender (snapshot tests that assert
  nothing specific).
- Running real Stryker inside the skill is out of scope for the default path
  (too slow, needs a working test env); consider it only as an explicit opt-in
  axis later.

## Additional Context

Complements the complexity/CRAP gate proposed in a separate issue. Coverage
answers "was this line executed?"; mutation testing answers "would a bug here be
caught?" — the two are orthogonal signals.

tom2drum:
## Description

Resolves #3665

Coverage answers "was this line executed?" — it cannot answer "would a bug here be caught?". A test can execute a branch and assert nothing about it. This adds mutation testing to answer the second question directly: change the code, and see whether any test notices.

The change adds a `pnpm test:mutation-testing` CLI around StrykerJS with the vitest runner, and a CI job that runs it scoped to a PR's changed lines and **fails when a mutant on one of those lines survives**. Findings are also emitted as inline annotations on the diff, so a weak assertion is visible during review rather than buried in a job log.

What bounds the run, in each direction:

- **Only logic mutators** are generated — conditional expressions, equality/logical/arithmetic operators, boolean literals. Every other class is excluded at generation time.
- **Only files with a co-located vitest spec** are mutated, so an untested file produces no findings instead of a run of unkillable mutants.
- **Only `behavior` lines** — mutants inside a `jsx` render body are excluded, reusing the complexity gate's AST function classifier.
- **Only changed lines** in diff mode, expressed as Stryker mutation ranges, so this restricts mutant *creation* rather than filtering afterwards. Three selection modes mirror the complexity gate's flags: full when bare, focused on given paths, diff-scoped under `--changed[=<ref>]`.
- **A wall-clock budget** stops a long run. Results are streamed to disk per mutant, so a truncated run still reports everything it completed and always says that it was truncated and how far it got.

Suppressing a mutant requires Stryker's disable comment carrying both the mutator name and a reason; a new ESLint rule rejects blanket file-level disables, which keeps each suppression a reviewable claim.

Decisions a reader would otherwise have to reverse-engineer:

- **StrykerJS has no `--since` flag** (that is Stryker.NET), so diff-scoping is computed here and expressed through mutation ranges.
- **The budget is wall-clock, not mutant count.** Runtime tracks import-graph centrality, not diff size: five leaf utilities with 140 mutants ran in 10s, while one 28-mutant file imported across the app took 70s, because Stryker's dry run executes everything related to it.
- **The logic-only mutator set came from sampling, not taste.** Of 22 survivors sampled across all mutators, roughly a third were unkillable by any vitest test (dependency arrays, theme objects, colour tokens); of 18 sampled under the logic-only set, zero were — all 18 were real gaps. Promoting a class back is a one-line config change.
- **Diff resolution and scope logic are imported from the complexity gate** rather than reimplemented.
- Two pnpm integration blockers are handled: a **scoped** `brace-expansion` override for Stryker's `minimatch` path (the global 1.x security pin is not widened), and naming the vitest runner explicitly in the plugin list, because Stryker's plugin globbing does not survive pnpm's isolated layout.
- No incremental mode and no committed report artifact.

## Environment variables

None.

## Minimum API version

None.

None for deployments — the tool adds no runtime code and no env var.

For contributors, a new required CI job (`Mutation testing gate`) runs on pull requests and fails when a mutant on a changed line survives. It chains on the unit-test job, since a red suite makes Stryker abort during its dry run with an error that reads like a tooling failure.

## Additional information

- The spec: [`.agents/tasks/3665-mutation-testing/spec.md`](https://github.com/blockscout/frontend/blob/issue-3665/.agents/tasks/3665-mutation-testing/spec.md).
- How to run it, what a failure means, and how to suppress a mutant: [`tools/mutation-testing/CONTEXT.md`](https://github.com/blockscout/frontend/blob/issue-3665/tools/mutation-testing/CONTEXT.md) and the `docs/` beside it.
- Reproduce a finding locally with `pnpm test:mutation-testing path/to/file.ts` — that mutates the whole file, while CI mutates only your changed lines, so a focused run may surface survivors that are not gating.


tom2drum (commented):
Review clear

| Severity | Count |
| --- | --- |
| blocker | 0 |
| major | 0 |
| nit | 0 |

| Axis | Count |
| --- | --- |
| spec | 0 |
| standards | 0 |
| correctness | 0 |

Outcome: clear

— Reviewed by Cursor Grok 4.6

tools/mutation-testing/index.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  console.log(formatTable(buildFileScores(results.report)));
+  if (results.truncated) console.log(formatTruncationNotice(results, options.budgetMs));
+  console.log(`\n${ formatFindings(findings) }`);
+  reportHtmlLocation();
+
+  // eslint-disable-next-line no-restricted-properties -- Node CLI detecting the CI runtime, not an app env var
+  if (process.env.GITHUB_ACTIONS) emitGithubActionsOutput(findings, results, options.budgetMs);
+
+  if (isFailingRun(findings)) process.exitCode = 1;
```
tom2drum: **F1 · major** — `if (isFailingRun(findings)) process.exitCode = 1` never looks at `results.truncated`. `dryRunTimeoutMinutes: 15` equals `DEFAULT_BUDGET_MS`, so a dry run that fills the budget yields an empty report, no survivors, and a green check. `CONTEXT.md` says a green result is not a pass. The `::warning` path documents a *partial* truncated run staying green on purpose. It does not cover testing nothing.

Fail when `results.truncated && results.tested === 0`. Shrink `dryRunTimeoutMinutes` so the dry run cannot consume the whole budget.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
The earlier “Review clear” over-calibrated. Two findings still hold.

| --- | --- | --- | --- |
| F1 | major | correctness | `tools/mutation-testing/index.ts:190` |
| F2 | nit | standards | `eslint.config.mjs:606` |

| Severity | Count |
| --- | --- |
| blocker | 0 |
| major | 1 |
| nit | 1 |

| Axis | Count |
| --- | --- |
| spec | 0 |
| standards | 1 |
| correctness | 1 |

Outcome: findings

eslint.config.mjs
```diff
@@ -598,4 +600,12 @@ export default tseslint.config(
       'spdx-license/header': 'error',
     },
   },
+
+  {
+    plugins: { stryker: { rules: { 'well-formed-disable': strykerDisableRule } } },
+    files: [ '**/*.{ts,tsx,mjs,js,cjs}' ],
```
tom2drum, line 606: **F2 · nit** — The global `files` glob gained `mts` in this same diff, so `streaming-reporter.mts` is in the TypeScript lint set. The new Stryker rule is `files: [ '**/*.{ts,tsx,mjs,js,cjs}' ]` and misses it.

Include `mts` in the stryker rule glob.

— Reviewed by Cursor Grok 4.6

tools/mutation-testing/index.ts
tom2drum: **F1a — fixed.** The premise holds: `selection.outcome === 'empty'` exits before Stryker, so a truncated run always had a non-empty selection, and `readStream()` on a stream with no records returns `{ tested: 0, planned: null }` → no survivors → exit 0. `render/truncation.ts` already prints "none of the selection was covered" for exactly that state, so the report said one thing and the exit code another.

`testedNothing(results)` now lives in `stryker/results.ts` (`truncated && tested === 0`) and is OR'd into the exit-code decision in `index.ts`. `isFailingRun`'s "the only thing that does" comment is corrected to point at it, `CONTEXT.md` and `docs/RUNNING.md` state that this one truncated outcome fails outright, and `results.spec.ts` covers all three cases. The gate on its own code: `pnpm test:mutation-testing tools/mutation-testing/stryker/results.ts` → 10 mutants, 10 killed, 100%.

**F1b — rejected.** Shrinking `dryRunTimeoutMinutes` is not the right fix. With F1a in place the green-check hole closes whichever timer wins the race, so it buys only a nicer error message — and it makes the documented remedy worse. `dryRunTimeoutMinutes` is a fixed config value while the budget is per-invocation, so lowering it to, say, 10 means a 12-minute dry run under `--budget 40` aborts as a Stryker tooling failure, contradicting `docs/RUNNING.md`: "raise `--budget` and let it finish." The flaw is the coupling, not the number; the only real fix is deriving `--dryRunTimeoutMinutes` from `budgetMs` at invocation, which is a separate change and not needed to close this hole.

eslint.config.mjs
tom2drum: **F2 — deferred** (nit). The claim is accurate: the global glob at `eslint.config.mjs:104` is `**/*.{js,mjs,cjs,ts,mts,jsx,tsx}` while the stryker block's is `**/*.{ts,tsx,mjs,js,cjs}`, and `tools/mutation-testing/stryker/streaming-reporter.mts` is the repo's only `.mts` file — so the rule policing Stryker suppression comments does not cover it.

Held as a nit rather than fixed in this round, per the resolve policy. Promote it and it is a one-token edit.

tools/mutation-testing/index.ts
tom2drum: Landed in c3a942defa.

tom2drum: F1a — verified. `testedNothing` (`truncated && tested === 0`) is OR'd into the exit at `index.ts:190`.

F1b — accepted, shrinking `dryRunTimeoutMinutes` would abort a longer `--budget` dry run as a Stryker tooling failure; F1a already closes the green-check hole.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
Review clear

F2 remains deferred.

| Severity | Count |
| --- | --- |
| blocker | 0 |
| major | 0 |
| nit | 1 |

| Axis | Count |
| --- | --- |
| spec | 0 |
| standards | 1 |
| correctness | 0 |

Outcome: clear

— Reviewed by Cursor Grok 4.6

PR #3679, opened 2026-09-02
Scope every interchain request to the deployment's slice of the shared indexer

### LINKED ISSUE #3650: Adapt the interchain UI to a shared multi-bridge / multi-network indexer (interchain-indexer v1.7.0)
### Overview

`interchain-indexer` v1.7.0 introduces read-time (per-request) filtering, so one service
instance can index many bridges and networks and still serve a narrowly scoped frontend.
We are consolidating the per-frontend indexer deployments into a single shared instance
that indexes **all** bridges — including messages to and from chains it does not index
at all — and that same instance will back Multichain Explorer.

Until now each frontend got its scope from *where it pointed*, not from *what it asked*:

| Before | Indexer scope | Consumer |
| --- | --- | --- |
| Avalanche ICM/ICTT instance | messages between Avalanche C-Chain and Numine only | Numine explorer |
| Omnibridge instance | messages between Gnosis and Ethereum only | Gnosis explorer |

| After | Indexer scope | Consumers |
| --- | --- | --- |
| One shared instance | bridge `1` AMB/Omnibridge (chains `1`, `100`) + bridge `2` Avalanche ICTT/ICM (C-Chain, Numine, and traffic to/from chains it does not index) — **with more chains and more bridges added over time** | Numine, Gnosis, Multichain Explorer, any future chain |

That is the point of the consolidation, and it is also the problem: if an existing
frontend is repointed at the shared instance **without** frontend changes, it starts
showing every cross-chain message and transfer the instance knows about — Gnosis↔Ethereum
Omnibridge traffic on Numine, Avalanche subnet traffic on Gnosis. The frontend has to
declare its slice on every request.

Because chains and bridges will keep being added on the backend, the scope must be
**derived at runtime**, not hardcoded: the new `GET /api/v1/interchain/bridges` endpoint
reports each bridge's live `indexed_chain_ids`, so onboarding a network becomes a backend
config change rather than a frontend release.

**The filter vocabulary is intentionally wider than this migration needs.** It was designed
as a general scoping mechanism so that one `interchain-indexer` can be pointed at a broad
range of consumers — single-chain explorers, bridge-specific views, multichain aggregators,
and whatever comes after — rather than as a minimal fix for the two frontends we are
migrating now. Consequence for this task: several params have no use case yet. A typical
single-chain explorer needs `home_chain_id` and `include_unindexed_chains` and nothing
else; the directional pair `src_chain_ids` / `dst_chain_ids` only earns its place once
there is an "outgoing only" / "incoming only" sub-view, `counterparty_chain_ids` only once
a deployment wants a subset of its peers rather than all of them, and `bridge_ids` only
once a chain is served by more than one bridge. Everything is optional and absent means
"no constraint", so take what the view needs and leave the rest — the unused fields are
headroom, not a checklist.

**The API is shape-compatible** — the proto delta between `interchain-indexer/v1.6.0` and
`v1.7.0` is additive only, and no route was renamed or removed — so nothing breaks on
contact. One behavioural default did change, and it is the thing to read first: list and
stats endpoints now **hide** rows touching a chain that the row's own bridge does not
index (such a row can never be observed from both sides). `include_unindexed_chains=true`
re-includes them, and every item carries a `has_unindexed_chain` flag.

<details>
<summary><b>Backward compatibility of v1.7.0 — verified item by item</b></summary>

Verified by diffing `interchain-indexer-proto` between the two tags: 198/85/33 inserted
lines across the three proto files, the only "deletion" being `message GetChainsRequest {}`
gaining an optional field. Two HTTP routes added, none renamed or removed.

| # | Change | Impact on an unmodified client |
| --- | --- | --- |
| 1 | **Default view narrows** — rows touching a chain the row's own bridge does not index are hidden. | ⚠️ Fewer items than v1.6.0. `include_unindexed_chains=true` restores the old output. |
| 2 | Malformed (non-hex) `message_id` / `tx_hash` / `address` now returns **400** (`code: 3`) instead of 500. | Improvement; check error handling that special-cases 5xx. |
| 3 | `GET /interchain/messages/{message_id}` returns **400** (`code: 9`, `"…provide bridge_id"`) when the id exists under more than one bridge, instead of an arbitrary row. | Previously unreachable (one bridge per instance); newly reachable. Needs handling. |
| 4 | Messages arriving from a chain their bridge does not index now carry `sender`, `recipient` and `payload` (were `null`). `source_transaction_hash` stays `null`. | Additive content, no shape change. |
| 5 | Projection-backed stats (`bridged-tokens`, `messages-paths`) were truncated and rebuilt by the release migration; `/stats/chains` unique-user counts rise (previously-excluded rows are now counted). | Absolute numbers take a one-time step. |
| 6 | A multi-hop ICTT transfer counts twice in bridged-token transfer counts (two ICM messages ⇒ two edges). | Accepted upstream; unexercised in any live deployment today. |

Also in the release with no frontend surface: RPC provider API keys and multi-bridge
configuration (config and ENV only) and an internal refactor.

</details>

### To‑do / Implementation notes

- [ ] **Resolve one "interchain filter profile" per deployment** — focal chain, optional
      bridge subset, optional counterparty set, unindexed-chain policy — and apply it to
      every interchain request the app makes, list and stats alike.
- [ ] **Derive chain and bridge selectors from the service, not from static config**,
      using `GET /interchain/bridges` (`indexed_chain_ids`) and `GET /interchain/chains`.
      New networks and bridges will be added on the backend continuously and must not
      require a frontend release.
- [ ] **Make the UI bridge-aware.** Rows on a shared instance come from several bridges;
      `bridge.id` is now available alongside `bridge.name`. Show bridge identity where a
      user could otherwise not tell two rows apart, and offer it as a user-facing filter
      (`bridge_ids`) where a deployment spans more than one bridge.
- [ ] **Send `include_unindexed_chains=true` where the product wants unconfirmable rows.**
      Confirmed for Numine: messages to/from Avalanche subnets that the bridge does not
      index must stay visible, marked via `has_unindexed_chain`. Keep it a deployment
      setting rather than a constant.
- [ ] **Qualify the message-details request with `bridge_id`.**
      `GET /api/v1/interchain/messages/{message_id}` used to take the id alone, which was
      unambiguous only because each instance served a single bridge. On the shared instance
      the same message id can exist under more than one bridge, so the endpoint gained an
      optional `bridge_id` **query param**, and the message object gained `bridge.id` —
      carry that value from the list row into the details request.
      Behaviour without `bridge_id`: the endpoint answers **only** when exactly one message
      carries that id, and returns `400` (`code: 9`, `"…provide bridge_id"`) on a
      cross-bridge collision. So an unqualified request is not an immediate error — it is a
      latent one that surfaces the first time a collision actually occurs, which makes it
      easy to miss in testing.
      For backward compatibility the response is still a **single message object, not an
      array**, qualified or not — no response-parsing change is needed.
      Note the knock-on effect on the app's own routing: if the message page URL carries
      only the id, a deep link or a page reload has no bridge to re-qualify with. The
      bridge needs to travel with the id.
- [ ] **Handle the new error semantics** — treat a malformed identifier (`400`, `code: 3`)
      as invalid input rather than a service error, and as a fallback for an unqualified
      details lookup that comes back ambiguous (`400`, `code: 9`), re-request with the
      bridge taken from the list row.
- [ ] **Keep pagination and filters in lockstep** — resend the identical filter params on
      every page request, and reset pagination whenever any filter changes.
- [ ] **Add a test that fails if a filter param is dropped or misspelled.** Unknown query
      params are silently ignored by the service (see traps below), so "the request
      succeeded" proves nothing.
- [ ] **Adopt Multichain Explorer as the deliberately unfiltered consumer**, and verify
      that its per-chain views reuse the same focal-chain filtering as a single-chain
      explorer.
- [ ] *(optional)* Surface indexing progress from the new `GET /api/v1/status/indexing`.

<details>
<summary><b>Filtering model — vocabulary, predicates, traps</b></summary>

#### Vocabulary

All filters are optional `snake_case` query params, `AND`-combined, applied on top of the
existing keyset pagination. `*_ids` params are comma-separated
(`counterparty_chain_ids=1,100`). Blank / whitespace / comma-only values normalize to
"filter absent"; malformed values return `400` `InvalidArgument` naming the param.

| Param | Type | Meaning |
| --- | --- | --- |
| `home_chain_id` | int64 | Focal chain: the row touches it on either side. |
| `counterparty_chain_ids` | CSV int64 | Allowed set for the *other* endpoint (or for both, when `home_chain_id` is absent). |
| `src_chain_ids` | CSV int64 | Directional: row's source chain must be in the set. |
| `dst_chain_ids` | CSV int64 | Directional: row's destination chain must be in the set. |
| `bridge_ids` | CSV uint32 | Row's bridge must be in the set (ids from `GET /interchain/bridges`). |
| `include_unindexed_chains` | bool | Opt back in to rows whose bridge cannot observe both endpoints. Default `false`. |

#### How the chain predicates combine

| Params present | Resulting predicate |
| --- | --- |
| `home_chain_id=H` | `src = H OR dst = H` (a row with unknown destination still matches on `src`) |
| `counterparty_chain_ids=S` | `src ∈ S AND dst ∈ S` (unknown destination excluded) |
| `home_chain_id=H` + `counterparty_chain_ids=S` | `(src = H AND dst ∈ S) OR (dst = H AND src ∈ S)` |
| `+ src_chain_ids=A` | `… AND src ∈ A` |
| `+ dst_chain_ids=B` | `… AND dst ∈ B` |
| `+ bridge_ids=C` | `… AND bridge ∈ C` |

For **transfers**, `src`/`dst` are the transfer's own token source/destination chains, not
the carrying message's chains — on a multi-hop route the two can differ.

For a chain-scoped page prefer `home_chain_id` over `src_chain_ids`/`dst_chain_ids`: the
directional params exist for "outgoing only" / "incoming only" sub-views and `AND` on top
of the focal predicate rather than replacing it.

#### `include_unindexed_chains` — the one behavioural change

A bridge only sees a chain it has a contract configured on. A row whose counterpart chain
is outside its bridge's indexed set can never be observed from both sides: an outgoing
Avalanche ICM message to an unindexed subnet stays `Initiated` forever, because the
confirming events happen on a chain nobody scans. v1.7.0 treats these as a distinct class:

- **hidden by default** on list and stats endpoints;
- **flagged** on every returned item via `has_unindexed_chain` (always present);
- **re-included** by `include_unindexed_chains=true`;
- **never hidden** on `GET /interchain/messages/{message_id}` — a link to a details page
  always resolves, even for a row the list hides. The param deliberately does not exist
  there.

Two consequences of the classification being derived from live config rather than stored
per row:

- `has_unindexed_chain` **is not an immutable property of a row**. When a chain is added
  to a bridge, existing rows touching it flip to `false` with no migration and no
  reprojection. Do not cache it as if it were part of the record.
- A row flipping to `has_unindexed_chain=false` **does not gain its missing fields**.
  History is not re-indexed, `source_transaction_hash` stays `null`, and `recipient` on an
  already-terminal row can never be patched. So `has_unindexed_chain=false` must not be
  treated as "all fields are populated" — render defensively either way.

#### Two traps

1. **Unknown query params are silently ignored.** The generated HTTP routes deserialize
   without `deny_unknown_fields`, so a typo'd or unsupported filter name is dropped and
   the response comes back *unfiltered*. For per-frontend slicing that is a cross-frontend
   data leak, not a visible error — hence the test in the checklist above.
2. **`bridge_id` ≠ `bridge_ids`.** On the *list* endpoints `bridge_id` is a raw pagination
   cursor field, not a filter; filtering there is `bridge_ids`. On
   `GET /interchain/messages/{message_id}`, `bridge_id` *is* the bridge qualifier.

#### Pagination is scoped to the filter set

`page_token` (and the raw cursor fields) are valid only for the exact filter combination
that produced them. The cursor does not encode filters and is not re-validated against
them, so mixing a stale token with changed filters yields undefined results. The same rule
applies to the separate `bridged-tokens` and `/stats/chains` cursors.

</details>

<details>
<summary><b>Endpoint reference — what is new in v1.7.0</b></summary>

Params and fields not listed are unchanged.

#### Messages & transfers

| Endpoint | New request params | New response fields |
| --- | --- | --- |
| `GET /api/v1/interchain/messages` | `home_chain_id`, `counterparty_chain_ids`, `bridge_ids`, `src_chain_ids`, `dst_chain_ids`, `include_unindexed_chains` | `items[].has_unindexed_chain`, `items[].bridge.id` |
| `GET /api/v1/interchain/messages:byTx/{tx_hash}` | same as above | same as above |
| `GET /api/v1/interchain/messages:byAddress/{address}` | same as above | same as above |
| `GET /api/v1/interchain/transfers` | same as above | `items[].has_unindexed_chain`, `items[].bridge.id` |
| `GET /api/v1/interchain/transfers:byTx/{tx_hash}` | same as above | same as above |
| `GET /api/v1/interchain/transfers:byAddress/{address}` | same as above | same as above |
| `GET /api/v1/interchain/messages/{message_id}` | `bridge_id` (uint32 qualifier) | `bridge.id`, `has_unindexed_chain`; `sender` / `recipient` / `payload` now populated for messages from unindexed chains |

#### Directories

| Endpoint | Notes |
| --- | --- |
| `GET /api/v1/interchain/bridges` | **New endpoint.** `items[]` of `{ id, name, type, enabled, ui_url, docs_url, indexed_chain_ids[] }`. `type` ∈ `lockmint` / `avalanche_native` / `amb`. `indexed_chain_ids` is an ascending array of int64-as-string (`["1","100"]`) reflecting live config; a bridge dropped from config reports `[]`. This is the source of truth for "which chains does this deployment cover". |
| `GET /api/v1/interchain/chains` | New `include_unindexed_chains`. **Default now omits** chains that no configured bridge indexes. Returns the union across all bridges — no per-bridge filter here, so intersect with `indexed_chain_ids` client-side for a bridge-scoped selector. |

#### Statistics

| Endpoint | New request params | Notes |
| --- | --- | --- |
| `GET /api/v1/stats/common` | `home_chain_id`, `counterparty_chain_ids`, `bridge_ids`, `src_chain_ids`, `dst_chain_ids`, `include_unindexed_chains` | Full vocabulary. The message counter matches message chains; the transfer counter matches token chains. |
| `GET /api/v1/stats/daily` | same as above | same as above |
| `GET /api/v1/stats/chain/{chain_id}/bridged-tokens` | `counterparty_chain_ids`, `bridge_ids`, `include_unindexed_chains` | Bridge rows are summed per asset before pagination. Page tokens are scoped to the same `bridge_ids` / `counterparty_chain_ids` / `q` / `sort` / `order`. `include_unindexed_chains` currently adds no rows here (aggregates for unindexed chain pairs are not written yet) — forward-compatibility only. |
| `GET /api/v1/stats/chain/{chain_id}/messages-paths/sent` | `bridge_ids`, `include_unindexed_chains` | `counterparty_chain_ids` already existed. Bridge rows are summed per directional edge. Same `include_unindexed_chains` caveat. |
| `GET /api/v1/stats/chain/{chain_id}/messages-paths/received` | same as above | same as above |
| `GET /api/v1/stats/chains` | `include_unindexed_chains` | **Intentionally outside the filtering model**: no `bridge_ids`, no `counterparty_chain_ids`. `chain_ids` selects *which chains get a row*, but each row's `unique_transfer_users_count` stays global across all bridges and counterparties. Acceptable while one bridge per chain holds; needs a backend follow-up once a chain is served by two bridges. |

#### Status

| Endpoint | Notes |
| --- | --- |
| `GET /api/v1/status/indexing` | **New endpoint.** Optional `bridge_id` / `chain_id` filters; `items[]` of per-`(bridge, chain)` progress. Read it correctly: `catchup_progress_percent` is the *scanned* share, not completeness — 100% with `failed_blocks > 0` is a normal reading, and `failed_blocks` is the only completeness signal. |
| `GET /api/v1/status/indexers` | **Unchanged in v1.7.0**, listed only for orientation next to the new endpoint above. `state` is `idle` / `running` / `failed: <error>`, where the `failed` variant carries the text of the error that stopped the indexer — same as in v1.6.0. |

</details>

### Mockups / Screenshots / Links

- Release: https://github.com/blockscout/blockscout-rs/releases/tag/interchain-indexer%2Fv1.7.0
- Per-frontend read filtering: https://github.com/blockscout/blockscout-rs/pull/1708
- Unindexed-chain horizon, `has_unindexed_chain`, bridges directory: https://github.com/blockscout/blockscout-rs/pull/1719
- Indexing progress API: https://github.com/blockscout/blockscout-rs/pull/1723
- Swagger: `interchain-indexer/interchain-indexer-proto/swagger/v1/interchain-indexer.swagger.yaml`
- Generated types package: `@blockscout/interchain-indexer-types@0.0.2`

### Deployment info

**The v1.7.0 deployment is not finished yet — the address of the shared instance will be
posted here separately once the service is up.** Implementation does not have to wait for
it: the API contract is already fixed and published (swagger + the generated types
package below), so the filtering work can be built and unit-tested against it, and only
end-to-end verification against real data needs the live URL.

**Backend side.** One shared `interchain-indexer` v1.7.0 instance replaces the
per-frontend instances. Initial coverage: bridge `1` AMB/Omnibridge on chains `1`
(Ethereum) and `100` (Gnosis); bridge `2` Avalanche ICTT/ICM on C-Chain (`43114`) and
Numine (`8021`), with traffic to and from unindexed chains retained. Additional chains and
additional bridges will be added incrementally, without frontend releases — which is why
selectors must come from `/interchain/bridges` and `/interchain/chains`.

**Suggested filter profile per consumer** (sent on every list/stats request):

| Consumer | Filters |
| --- | --- |
| Numine explorer | `home_chain_id=8021`, `include_unindexed_chains=true`; optionally `bridge_ids=2` |
| Gnosis explorer | `home_chain_id=100`; optionally `bridge_ids=1`. Both AMB chains are always indexed, so the unindexed flag is a no-op here |
| Multichain Explorer, global views | no focal filter (deliberately unfiltered); `include_unindexed_chains=true` |
| Multichain Explorer, per-chain view | `home_chain_id=<that chain>` — identical to a single-chain explorer |

`home_chain_id` alone is sufficient scoping today, since no two bridges share a chain in
the current config. `bridge_ids` is the guard for when they do — cheaper to send from the
start than to retrofit.

**Rollout.** Frontend changes are additive and can ship before the repoint; the old
instances keep working with the new params absent. Order that avoids a visible regression:
ship the filter profile → repoint one frontend and verify its slice → repoint the rest →
retire the dedicated instances.

**Expect a one-time step in stats values** at the backend rollout (projection rebuild plus
previously-excluded rows now counted). Not a regression; worth a heads-up wherever these
are charted.

### Acceptance criteria

- Every interchain request carries the deployment's filter profile, including page 2+ and
  refetches; changing a filter resets pagination.
- Repointing the Numine frontend at the shared instance shows only Numine-related traffic,
  and the Gnosis frontend only Gnosis↔Ethereum traffic — asserted by a test that fails if
  a filter param is dropped or misspelled, not merely that the request returned 200.
- Numine shows messages to/from chains bridge `2` does not index, visually marked, with
  the empty destination side and non-advancing status explained rather than looking broken.
- Bridge identity is visible wherever rows from more than one bridge can appear, and
  selectable where a deployment spans more than one bridge.
- Adding a chain or a bridge on the backend surfaces in chain/bridge selectors with no
  frontend release.
- The message details request carries `bridge_id`, including after a page reload or from a
  deep link; an unqualified lookup that comes back ambiguous (`400`, `code: 9`) is
  recovered from rather than shown as a failure; a malformed identifier (`400`, `code: 3`)
  is surfaced as invalid input, not as a service error.
- Rendering does not assume `has_unindexed_chain=false` implies fully populated fields, and
  does not cache the flag as an immutable property of a row.

tom2drum:
## Description

Resolves #3650

The per-consumer `interchain-indexer` instances are being consolidated into one shared instance per environment, indexing every bridge and every chain. Until now a frontend got its interchain scope from *where it pointed*, so repointing it at the shared instance would make it show every message that instance knows about — Gnosis↔Ethereum Omnibridge traffic on an Avalanche explorer, and the reverse. Unknown query params are silently ignored by the service, so a dropped or misspelled filter is not an error but a cross-deployment data leak that returns `200`.

Every request to an `interchainIndexer:*` resource now carries the deployment's filter profile, and no call site opts in:

- each resource entry **declares** which scope params its endpoint accepts (a generic `scopeFilters` field on `ApiResource`, so any service with static request filters can use it), and the profile is resolved and merged in **one** place — `buildUrl`, the funnel every request passes through. Resolvers are keyed by service as well as param name, so a same-named filter on another API cannot pick up interchain logic, and a declared name with no resolver is omitted rather than fatal;
- the derivation sends `home_chain_id` when the resource accepts it and a focal chain exists, otherwise `bridge_ids` from config, with `include_unindexed_chains` sent independently. A per-chain multichain view therefore produces a byte-identical query string to a single-chain explorer;
- a test walks the resource registry rather than a hand-written list, so a resource added later is covered the day it lands, and a dropped or misspelled param fails.

The message-details page moves to `/bridge/[bridgeId]/cross-chain-tx/[id]` and sends `bridge_id`. On a shared instance the same `message_id` can exist on more than one bridge, so an unqualified lookup is a guess; the qualified route makes the page resolve identically on a fresh load, a reload and a pasted deep link. A malformed message id now lands on the existing invalid-input page instead of the generic 500 copy.

## Environment variables

- Added `NEXT_PUBLIC_CROSS_CHAIN_TXS_BRIDGE_IDS` to state which bridges' messages and transfers a deployment is allowed to show — its slice of the shared indexer. Required when the feature is enabled.
- Added `NEXT_PUBLIC_CROSS_CHAIN_TXS_INCLUDE_UNINDEXED_CHAINS` to also show rows whose counterparty chain no configured bridge indexes. Optional, defaults to `false`; no deployment is expected to enable it today.

## Minimum API version

`interchain-indexer` **v1.8.2+**. The version that accepts `bridge_ids` on `GET /api/v1/interchain/chains`, which is what scopes the counterparty selector; earlier versions ignore the param and list the union of chains across every bridge on the instance.

- **Every cross-chain-enabled deployment must set `NEXT_PUBLIC_CROSS_CHAIN_TXS_BRIDGE_IDS`.** An instance with the feature on and no bridge ids **fails startup validation** by design — booting it against the shared indexer would silently serve another deployment's traffic. The per-instance values are listed in https://github.com/blockscout/frontend/issues/3650#issuecomment-5538918490.
- **`NEXT_PUBLIC_INTERCHAIN_INDEXER_API_HOST` must be repointed** at the shared instance for the environment. Same comment lists the value per deployment.
- **`/cross-chain-tx/[id]` is removed with no redirect.** The predecessor explorers never had a single-message page, so there is nothing to stay compatible with; links to a message are built from data that already carries its bridge.

## Additional information

The spec for the whole task is at `.agents/tasks/3650-shared-multi-bridge-indexer/spec.md`; the ticket breakdown lives in this branch's history, one commit per ticket.

Out of scope, deliberately: a chain configured on the deployment's own bridge but with no messages still appears in the counterparty selector, and `GET /interchain/bridges` stays unregistered — the bridge ids come from config, and the service does the chain filtering.

#3680 tracks a generalized fix for a Playwright fixture trap this work surfaced: `mockApiResponse` builds its route matcher in Node from `process.env` while the app resolves the same envs from `localStorage`, so an env feeding a query param desyncs the two sides. It is handled per test case here.


src/api/utils/scope-filters.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+  };
+}
+
+describe('every declared scope filter reaches the URL of every registered resource', () => {
+  test.each(INTERCHAIN_RESOURCES)('interchainIndexer:%s on a focal-chain deployment', async(name, resource) => {
+    const url = await buildUrlWithEnvs(CROSS_CHAIN_ENVS, `interchainIndexer:${ name }` as ResourceName, {
+      pathParams: pathParamsStub(resource),
+    });
+
+    expect(Object.fromEntries(new URL(url).searchParams)).toEqual(expectedFocalChainProfile(resource.scopeFilters));
```
tom2drum, line 171: **F1 · blocker** — FR4 requires a test that fails when a profile param is dropped or misspelled for any `interchainIndexer:*` resource. `expect(...).toEqual(expectedFocalChainProfile(resource.scopeFilters))` rebuilds the expected query from the same `scopeFilters` declaration, so dropping `home_chain_id` from `transfers` still matches. The fixture at line 193 then locks the hole in: a misspelled `bridge_id` "is left out of the query string instead of failing the request".

Assert each resource against an independent expected map (the first `describe` already does this for `messages` / `message` / `chains`). Fail a declared name that has no interchain resolver. Typing `scopeFilters` as `Array<keyof typeof SCOPE_FILTER_RESOLVERS>` would also catch typos at compile time.

— Reviewed by Cursor Grok 4.6

src/api/resources/types.ts
```diff
@@ -4,6 +4,8 @@ export interface ApiResource {
   path: string;
   pathParams?: Array<string>;
   filterFields?: Array<string>;
+  // static filters; hard-coded or derived from ENVs variables
```
tom2drum: **F2 · nit** — `// static filters; hard-coded or derived from ENVs variables` restates the field. `.agents/rules/code-quality.md`: do not write comments that say what the next line is.

Delete the comment.

— Reviewed by Cursor Grok 4.6

src/api/utils/scope-filters.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+const DEPLOYMENT_CHAIN_ID = '1';
+const BRIDGE_IDS = '1,2';
+const BRIDGE_IDS_ENCODED = encodeURIComponent(BRIDGE_IDS);
+const PATH_PARAM_STUB = 'stub';
+const NO_SCOPE_FILTERS: Array<string> = [];
+
+const CROSS_CHAIN_ENVS = ENVS_MAP.crossChainTxs;
+const MULTICHAIN_ENVS = [ ...ENVS_MAP.crossChainTxs, ...ENVS_MAP.multichain ];
+
+const CLUSTER_CHAIN = { id: '10', slug: 'op-mainnet', name: 'OP Mainnet' } as unknown as ClusterChainConfig;
```
tom2drum: **F3 · nit** — `{ id: '10', … } as unknown as ClusterChainConfig` is an unexplained double-cast, then passed as `ExternalChainExtended`. `.agents/rules/typescript.md`: if a double-cast is unavoidable, add a comment explaining why it is safe. The `importBuildUrl` cast below already does this; this one does not.

Type the fixture as `ExternalChainExtended` (or `{ id: string }`) and drop the double-cast.

— Reviewed by Cursor Grok 4.6

deploy/tools/envs-validator/schemas/features/crossChainTxs.ts
```diff
@@ -1,10 +1,25 @@
...
+      .of(yup.number().required())
+      .when('NEXT_PUBLIC_CROSS_CHAIN_TXS_ENABLED', {
+        is: (value: boolean) => value,
+        then: (schema) => schema.required(),
+        otherwise: (schema) => schema.max(
+          -1,
+          'NEXT_PUBLIC_CROSS_CHAIN_TXS_BRIDGE_IDS can only be used with NEXT_PUBLIC_CROSS_CHAIN_TXS_ENABLED',
+        ),
+      }),
+    NEXT_PUBLIC_CROSS_CHAIN_TXS_INCLUDE_UNINDEXED_CHAINS: yup.boolean(),
```
tom2drum: **F4 · nit** — `NEXT_PUBLIC_CROSS_CHAIN_TXS_INCLUDE_UNINDEXED_CHAINS: yup.boolean()` is not gated on `ENABLED`, unlike `BRIDGE_IDS` and `INTERCHAIN_INDEXER_API_HOST` in the same schema. `deploy/tools/envs-validator/CONTEXT.md` companion variables: gate B on A and forbid B in the otherwise-branch.

Use `.when('NEXT_PUBLIC_CROSS_CHAIN_TXS_ENABLED', …)` with `value === undefined` in `otherwise`.

— Reviewed by Cursor Grok 4.6

src/features/cross-chain-txs/components/txs/TransactionsCrossChainTable.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+const BRIDGE_NAME = withUnindexedDestination.bridge.name;
+const PENDING_DESTINATION_CHAIN_NAME = pending.destination_chain.name;
+
+describe('TransactionsCrossChainTable', () => {
+  afterEach(cleanup);
+
+  it('renders a message whose destination never resolves as an ordinary row with an empty destination side', () => {
+    const { container } = render(<TransactionsCrossChainTable data={ [ withUnindexedDestination ] }/>);
+
+    const cells = container.querySelectorAll('tbody tr:first-child td');
```
tom2drum: **F5 · nit** — `container.querySelectorAll('tbody tr:first-child td')` plus `SOURCE_TX_CELL_INDEX = 4` asserts through column ordinals. `.agents/rules/tests-unit.md`: tests verify behavior through public interfaces; a test that breaks when columns are reordered without a behavior change is coupled to layout. The sibling list spec already checks labeled `Destination tx` / `Recipient` text.

Query by the visible labels, as `TransactionsCrossChainList.spec.tsx` does.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
| Id | Severity | Axis |
| --- | --- | --- |
| F1 | blocker | spec |
| F2 | nit | standards |
| F3 | nit | standards |
| F4 | nit | standards |
| F5 | nit | standards |

Spec axis ran against `.agents/tasks/3650-shared-multi-bridge-indexer/spec.md`.

— Reviewed by Cursor Grok 4.6

src/api/utils/scope-filters.spec.ts
tom2drum: Fixed in 93f3224 — but not the way this suggests, and FR4 itself was amended.

**The claim is right.** `expectedFocalChainProfile(resource.scopeFilters)` rebuilt the expectation from the declaration under test, so dropping `home_chain_id` from `transfers` passed. Verified by mutation before changing anything: that edit produced zero failures.

**The suggested fix was rejected as a change-detector.** An independent expected map per resource encodes no knowledge beyond "the declarations are what they were the day it was written" — it has to be edited in lockstep with every legitimate registry change, and the author just updates it. Grading the failure modes it would catch:

| Mode | Frequency | Worth a test? |
| --- | --- | --- |
| New resource added, `scopeFilters` forgotten → silent leak | real; the registry grows when endpoints ship | yes |
| Filter name misspelled → silently dropped | low but real | yes |
| Filter deleted from an existing resource | not an accident — a deliberate one-line edit | no, that is code review |

Only the first two have frequency, and neither needs duplication. So the walk now asserts two invariants instead:

```ts
test.each(INTERCHAIN_RESOURCES)('interchainIndexer:%s declares a scope', (name, resource) => {
  const declared = resource.scopeFilters ?? NO_SCOPE_FILTERS;
  expect(declared.length > 0).toBe(!UNSCOPED_BY_DESIGN.includes(name));
});

test.each(INTERCHAIN_RESOURCES)('interchainIndexer:%s declares only resolvable filters', (name, resource) => {
  expect(KNOWN_SCOPE_FILTERS).toEqual(expect.arrayContaining(resource.scopeFilters ?? NO_SCOPE_FILTERS));
});
```

Both survive any legitimate change to *which* filters a resource takes, which is exactly the property a pinned map lacks. Mutation-verified, all three restored after:

- new resource without `scopeFilters` → fails on `declares a scope`
- `bridge_ids` → `bridge_id` on `stats_chains` → fails on `declares only resolvable filters`
- `home_chain_id` deleted from `transfers` → passes, now deliberately uncovered

**FR4 was reworded** to that intent. Its literal wording ("a test fails when a profile param is dropped … for any resource") is only satisfiable by duplication, which over-specifies: the leak risk the Context section names is dominated by new resources and by the injection path, not by someone deleting a line from an existing declaration. The implementation decision at spec line 160 now records why the stricter reading was dropped, so it does not get "fixed" back later.

The registry enumeration is kept — `test.each` over `RESOURCES.interchainIndexer`, per spec line 161.

tom2drum: On the third suggestion specifically — **rejected**.

> Typing `scopeFilters` as `Array<keyof typeof SCOPE_FILTER_RESOLVERS>` would also catch typos at compile time.

That contradicts the design intent. Spec line 133 makes `scopeFilters` "a *generic* field usable by any service that needs static filters — not an interchain-specific map", and the fixture test at the bottom of this spec exists to assert that a `core` resource declaring `home_chain_id` does **not** inherit the interchain resolver of the same name. Typing the field against `SCOPE_FILTER_RESOLVERS.interchainIndexer`'s keys would make that impossible to express.

The misspelling is caught at test time instead, by `declares only resolvable filters`, which checks each declared name against its **own service's** resolver map — same protection, without collapsing the generic field into an interchain-specific one.

src/api/resources/types.ts
tom2drum: Fixed in 93f3224 — comment deleted.

Agreed: `.agents/rules/code-quality.md` is unconditional ("Do not write comments. Never. Period."), and this one restated the field rather than explaining why anything non-obvious was needed.

src/api/utils/scope-filters.spec.ts
tom2drum: Fixed in 93f3224 — the double-cast is gone, via a different route than suggested.

The suggested fix does not work: `ClusterChainConfig` requires `app_config: typeof config`, so a bare `{ id, slug, name }` literal cannot be typed as it, and `ExternalChainExtended` is a union of two interfaces with the same problem. The cast could not simply be dropped.

Instead the spec now reuses `chainA` from `src/features/multichain/mocks/chains.ts` — an existing shared fixture already typed as `ClusterChainConfig`, where the unavoidable cast lives once, next to the 60-line `app_config` that justifies it.

**That reuse surfaced something worth recording.** `getResourceParams` prefers `chain.app_config.apis[apiName]` over the deployment config whenever the chain carries an `app_config.apis` — and a real `ClusterChainConfig` always does, since the field is non-optional. The old inline literal had no `app_config` at all, so it silently fell through to the deployment config and never exercised the branch a real multichain request takes. The fixture now names its indexer host explicitly:

```ts
// a cluster chain always carries `app_config.apis`, and `getResourceParams` prefers the host declared
// there over the deployment's own — so the fixture has to name the indexer it points at
const CHAIN_IN_SCOPE = { ...chainA, app_config: { ...chainA.app_config, apis: { ...chainA.app_config.apis, interchainIndexer: { endpoint: INTERCHAIN_HOST } } } };
```

Which raises a real question beyond this finding: `interchainIndexer` is built from the deployment's own `NEXT_PUBLIC_INTERCHAIN_INDEXER_API_HOST` (`src/api/config.ts`), and `useApiQuery` auto-injects `multichainContext.chain`, so a chain-scoped multichain interchain request resolves its host from the cluster chain's config rather than the deployment's. Both `get-resource-params.ts` and `useApiQuery.ts` are untouched by this PR, so it is pre-existing and out of scope here — flagged for a separate look.

deploy/tools/envs-validator/schemas/features/crossChainTxs.ts
tom2drum: Fixed in 93f3224.

`NEXT_PUBLIC_CROSS_CHAIN_TXS_INCLUDE_UNINDEXED_CHAINS` is now gated on `ENABLED` with a `not-exist` otherwise-branch, matching `BRIDGE_IDS` and `INTERCHAIN_INDEXER_API_HOST` in the same schema and the companion-variable convention in `deploy/tools/envs-validator/CONTEXT.md`. The `then` branch is omitted since it would be a no-op.

Negative path verified, as CONTEXT.md step 5 requires — flipping `NEXT_PUBLIC_CROSS_CHAIN_TXS_ENABLED=false` in `test/.env.cross_chain_txs` produces the new error alongside its two siblings:

```
🚨 ENVs validation failed with the following errors:
     NEXT_PUBLIC_CROSS_CHAIN_TXS_BRIDGE_IDS can only be used with NEXT_PUBLIC_CROSS_CHAIN_TXS_ENABLED
     NEXT_PUBLIC_CROSS_CHAIN_TXS_INCLUDE_UNINDEXED_CHAINS can only be used with NEXT_PUBLIC_CROSS_CHAIN_TXS_ENABLED
     NEXT_PUBLIC_INTERCHAIN_INDEXER_API_HOST can only be used with NEXT_PUBLIC_CROSS_CHAIN_TXS_ENABLED
```

Preset restored; the full suite passes on the happy path.

src/features/cross-chain-txs/components/txs/TransactionsCrossChainTable.spec.tsx
tom2drum: Fixed in 93f3224 — the four ordinal constants are gone.

The suggested fix does not translate, though: the table renders its labels only in `<thead>`, not per cell, so `container.textContent` label matching (what `TransactionsCrossChainList.spec.tsx` does, because the mobile list item renders a label beside each value) cannot isolate a cell here.

Cells are resolved by header text instead, which decouples the spec from column order while still working for a table:

```tsx
function firstRowCellByColumn(container: HTMLElement, header: string): Element {
  const headers = Array.from(container.querySelectorAll('thead th'));
  const index = headers.findIndex((cell) => cell.textContent?.trim() === header);

  if (index < 0) {
    throw new Error(`no "${ header }" column in the table header`);
  }

  return container.querySelectorAll('tbody tr:first-child td')[index];
}
```

A column reorder now leaves the spec passing; a renamed or removed column fails with the column name rather than an out-of-range index.

src/api/utils/scope-filters.spec.ts
tom2drum: F1 — verified

src/api/resources/types.ts
tom2drum: F2 — verified

src/api/utils/scope-filters.spec.ts
tom2drum: F3 — verified

deploy/tools/envs-validator/schemas/features/crossChainTxs.ts
tom2drum: F4 — verified

src/features/cross-chain-txs/components/txs/TransactionsCrossChainTable.spec.tsx
tom2drum: F5 — verified

tom2drum (commented):
Review clear

| Severity | Count |
| --- | --- |
| blocker | 0 |
| major | 0 |
| nit | 0 |

| Axis | Count |
| --- | --- |
| spec | 0 |
| standards | 0 |
| correctness | 0 |

F1–F5 verified; no regressions from 93f3224.

— Reviewed by Cursor Grok 4.6

PR #3681, opened 2026-09-04
Let operators pick Flashblocks or Subblocks naming for the OP Stack feed

pcw109550:
## Description

OP Stack chains are split between the newer **Subblocks** name and the legacy **Flashblocks** name for their sub-second pre-confirmation feed, while both stream the same wire format. One frontend image serves both kinds of chain, but the label has been hard-coded (`subblock` since blockscout/frontend#3643), so a chain that still runs Flashblocks has no way to keep its name.

- `NEXT_PUBLIC_FLASHBLOCKS_NAME` selects the label per deployment. It drives the `name` field of the flashblocks feature config, from which every user-facing string already derives: the blocks-page tab, the stats labels, the real-time feed hint and the new-items notice. MegaETH keeps `mini-block` regardless.
- The feature config also carries the blocks-page tab ids, the configured name first and the other OP Stack name as an alias, so links written under either name keep resolving. This also fixes a stale `?tab=` link falling back to the "All" tab with its query disabled.
- A `blocks` `getServerSideProps` guard redirects the alias to the canonical id (`/blocks?tab=flashblocks` → `/blocks?tab=subblocks` under the default, and the reverse when the name is `flashblock`), preserving the other query params. The redirect is temporary (307) because an operator can flip the name later.
- Parsers, the socket hook and the table components are untouched: the wire format is identical, so only labels differ.

Two decisions worth a look from reviewers:

- The default stays `subblock`, as introduced in #3643, so this change is purely additive: no deployment's label changes on upgrade. Chains that still run Flashblocks (Base, for one) opt in with `NEXT_PUBLIC_FLASHBLOCKS_NAME=flashblock`.
- Aliases exist only between the two OP Stack names. The MegaETH `mini-blocks` tab has no alias and no redirect.

## Environment variables

- Added `NEXT_PUBLIC_FLASHBLOCKS_NAME` (`flashblock` | `subblock`, default `subblock`) to set what the chain calls its pre-confirmation blocks; it controls every label on the feed and the blocks-page tab id. The envs validator rejects any other value and rejects the variable when `NEXT_PUBLIC_FLASHBLOCKS_SOCKET_URL` is not set, since it has no effect without the feed. Both rules moved into a `flashblocks` feature sub-schema.
- `NEXT_PUBLIC_FLASHBLOCKS_SOCKET_URL` is unchanged; its description in `docs/ENVS.md` now mentions both names.

## Minimum API version

None. The feed is streamed straight from the sequencer WebSocket; no Core API or microservice change is involved.

None. Without the new variable every deployment keeps its current "Subblocks" label, and the previously unrecognised `?tab=flashblocks` links now redirect instead of falling back to the "All" tab.

## Additional information

Verification on this branch:

- Vitest: two new spec files (config resolution, redirect guard) with 11 tests; the full suite passes.
- `tsc`, `eslint` and `cspell` clean; the cognitive-complexity / CRAP gate passes with the new guard fully covered; envs-validator presets all pass, and both negative paths are rejected: an unknown name (`must be one of the following values: flashblock, subblock`) and a name without the socket URL (`can only be used with NEXT_PUBLIC_FLASHBLOCKS_SOCKET_URL`).
- Live check on a local dev server with the Base socket URL configured, in both name configurations:

| Config | Request | Result |
| --- | --- | --- |
| default (`subblock`) | `/blocks?tab=flashblocks` | 307 → `/blocks?tab=subblocks` |
| default (`subblock`) | `/blocks?tab=flashblocks&page=3` | 307 → `/blocks?page=3&tab=subblocks` |
| default (`subblock`) | `/blocks?tab=subblocks`, `?tab=reorgs` | 200 |
| `NAME=flashblock` | `/blocks?tab=subblocks` | 307 → `/blocks?tab=flashblocks` |
| `NAME=flashblock` | `/blocks?tab=subblocks&page=2` | 307 → `/blocks?page=2&tab=flashblocks` |
| `NAME=flashblock` | `/blocks?tab=flashblocks`, `?tab=reorgs`, `/blocks` | 200 |

No Playwright snapshot was added: with the feature disabled (the Playwright default) the blocks page renders exactly as before, and the new behaviour has no visual component beyond the label text.

## Review follow-up

Commit 2a78d176e addresses the first review round. Nothing was force-pushed, so the delta is viewable on its own.

- **Tab ids are literal and typed (F1, and the `config.ts:19` suggestion).** `FLASHBLOCKS_TAB_IDS` / `FlashblocksTabId` live in `types/config.ts`, the two OP Stack orderings are written out in `config.ts`, and `tabIds` is a non-empty mutable tuple, so the guard's destructuring of the canonical id is sound and the ids are greppable. The `.map` derivation and its comment are gone. The suggested one-liner was not applied verbatim because it would have dropped the canonical-first ordering that the config spec pins.
- **`includes` became `some` at both call sites.** Once the element type is the literal union, `tabIds.includes(tab)` with `tab: string` fails with TS2345, since call arguments are checked by assignability rather than bivariance. `some((id) => id === tab)` compares instead and compiles without a cast.
- **Guard renamed to `blocks` (maintainer comment and F5).** Matches how the other guards are named for their page and the existing `blocks` export in `main.ts`. F5's `flashblocksTabAlias` was considered and set aside for consistency with the file.
- **Guard spec uses `withEnvs` again (F2).** No more `as unknown as typeof config`; the spec drives the real config through env presets like `config.spec.ts`. Helpers carry return types and the single-use constant is inlined.
- **Glossary row (F3).** The OP Labs rename etymology is back and "(the default)" is out; defaults stay in `docs/ENVS.md`.
- **Rationale stated once (F4).** The code comments in `types/config.ts`, `config.ts` and `guards.ts` are removed; `docs/ENVS.md` is the single statement.

Re-verified on 2a78d176e: `tsc`, `eslint`, `cspell`, the complexity gate, the full Vitest suite (93 files, 663 tests), all 16 validator presets, and both validator negative paths.

**Round 2, commit 713e13864.** Both new nits applied as suggested: the MegaETH `tabIds` constant is typed `[ 'mini-blocks' ]` rather than the full union (F7), and the guard spec annotates its env list as `Array<[ string, string ]>` instead of casting the inline pair (F6). Both call sites keep `some(...)`; the reviewer confirmed `includes` also compiles here thanks to `ts-reset`, so either form is fine and this one stays. Re-verified on 713e13864: `tsc`, `eslint`, `cspell`, the complexity gate, the full Vitest suite (93 files, 663 tests), and all 16 validator presets.

### DIFF
--- a/.agents/GLOSSARY.md
+++ b/.agents/GLOSSARY.md
@@ -39,7 +39,7 @@ vars are documented in `docs/ENVS.md`. Architectural concepts like
 | **Eden** | chain | A rollup built on `ev-reth` / evstack. Introduces the **Sponsored Transaction** type. |
 | **Epoch** | entity | A consensus time period specific to **Celo**. Has its own index and detail pages. Always refers to a Celo epoch in this codebase — not a generic blockchain concept. |
 | **Fault Proof System** | feature | Optimism's mechanism for proving the correctness of L2 state transitions on L1 via **Dispute Games**. |
-| **Flashblocks** | feature | Code name for the sub-second pre-confirmation block streaming feature. Surfaced to users as **Subblocks** on OP Stack chains (OP Labs renamed it from "Flashblocks") and as **mini-blocks** on MegaETH. The `flashblocks` code name and `FLASHBLOCKS` env vars are retained. |
+| **Flashblocks** | feature | Code name for the sub-second pre-confirmation block streaming feature. Surfaced to users as **Subblocks** (the default) or **Flashblocks** on OP Stack chains, whichever name the chain uses, and as **mini-blocks** on MegaETH. The `flashblocks` code name and `FLASHBLOCKS` env vars are retained. |
 | **Hot Contracts** | feature | Ranked list of the most recently and frequently interacted-with smart contracts on the network. |
 | **Interchain Indexer** | service | Microservice that indexes cross-chain messages and token transfers across heterogeneous chains. General-purpose interop indexer, not ZetaChain-specific. Provides "Cross chain txs" feature. Distinct from **CCTX**. |
 | **Interop Messages** | entity | **Deprecated** Cross-rollup messages passed between OP Stack chains using the native interoperability protocol. Distinct from **Interchain Indexer** messages. |
--- a/.agents/GLOSSARY.md
+++ b/.agents/GLOSSARY.md
@@ -39,7 +39,7 @@ vars are documented in `docs/ENVS.md`. Architectural concepts like
 | **Eden** | chain | A rollup built on `ev-reth` / evstack. Introduces the **Sponsored Transaction** type. |
 | **Epoch** | entity | A consensus time period specific to **Celo**. Has its own index and detail pages. Always refers to a Celo epoch in this codebase — not a generic blockchain concept. |
 | **Fault Proof System** | feature | Optimism's mechanism for proving the correctness of L2 state transitions on L1 via **Dispute Games**. |
-| **Flashblocks** | feature | Code name for the sub-second pre-confirmation block streaming feature. Surfaced to users as **Subblocks** (the default) or **Flashblocks** on OP Stack chains, whichever name the chain uses, and as **mini-blocks** on MegaETH. The `flashblocks` code name and `FLASHBLOCKS` env vars are retained. |
+| **Flashblocks** | feature | Code name for the sub-second pre-confirmation block streaming feature. Surfaced to users as **Subblocks** or **Flashblocks** on OP Stack chains, whichever name the chain uses (OP Labs renamed it from "Flashblocks"), and as **mini-blocks** on MegaETH. The `flashblocks` code name and `FLASHBLOCKS` env vars are retained. |
 | **Hot Contracts** | feature | Ranked list of the most recently and frequently interacted-with smart contracts on the network. |
 | **Interchain Indexer** | service | Microservice that indexes cross-chain messages and token transfers across heterogeneous chains. General-purpose interop indexer, not ZetaChain-specific. Provides "Cross chain txs" feature. Distinct from **CCTX**. |
 | **Interop Messages** | entity | **Deprecated** Cross-rollup messages passed between OP Stack chains using the native interoperability protocol. Distinct from **Interchain Indexer** messages. |
--- a/src/features/flashblocks/config.ts
+++ b/src/features/flashblocks/config.ts
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: LicenseRef-Blockscout
 
+import type { FlashblocksName } from 'src/features/flashblocks/types/config';
+import { FLASHBLOCKS_NAMES } from 'src/features/flashblocks/types/config';
+
 import megaEthFeature from 'src/features/chain-variants/mega-eth/config';
 
 import { getEnvValue } from 'src/config/utils/envs';
@@ -8,8 +11,19 @@ import type { Feature } from 'src/config/utils/features';
 const title = 'Flashblocks';
 
 const socketUrl = getEnvValue('NEXT_PUBLIC_FLASHBLOCKS_SOCKET_URL');
+const nameFromEnv = getEnvValue('NEXT_PUBLIC_FLASHBLOCKS_NAME');
+const opStackName = FLASHBLOCKS_NAMES.find((name) => name === nameFromEnv) ?? 'subblock';
 
-const config: Feature<{ socketUrl: string; type: 'optimism' | 'megaEth'; name: string }> = (() => {
+// The configured name owns the blocks-page tab; the other OP Stack name stays an alias so links
+// written under either name keep resolving (the `blocksTab` guard redirects it to the first id).
+const opStackTabIds = [ opStackName, ...FLASHBLOCKS_NAMES.filter((name) => name !== opStackName) ].map((name) => `${ name }s`);
+
+const config: Feature<{
+  socketUrl: string;
+  type: 'optimism' | 'megaEth';
+  name: FlashblocksName | 'mini-block';
+  tabIds: Array<string>;
+}> = (() => {
   if (megaEthFeature.isEnabled && megaEthFeature.socketUrl.rpc) {
     return Object.freeze({
       title,
@@ -17,6 +31,7 @@ const config: Feature<{ socketUrl: string; type: 'optimism' | 'megaEth'; name: s
       socketUrl: megaEthFeature.socketUrl.rpc,
       type: 'megaEth',
       name: 'mini-block',
+      tabIds: [ 'mini-blocks' ],
     });
   }
 
@@ -26,7 +41,8 @@ const config: Feature<{ socketUrl: string; type: 'optimism' | 'megaEth'; name: s
       isEnabled: true,
       socketUrl,
       type: 'optimism',
-      name: 'subblock',
+      name: opStackName,
+      tabIds: opStackTabIds,
     });
   }
 
--- a/src/features/flashblocks/config.ts
+++ b/src/features/flashblocks/config.ts
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: LicenseRef-Blockscout
 
-import type { FlashblocksName } from 'src/features/flashblocks/types/config';
+import type { FlashblocksName, FlashblocksTabId } from 'src/features/flashblocks/types/config';
 import { FLASHBLOCKS_NAMES } from 'src/features/flashblocks/types/config';
 
 import megaEthFeature from 'src/features/chain-variants/mega-eth/config';
@@ -8,21 +8,23 @@ import megaEthFeature from 'src/features/chain-variants/mega-eth/config';
 import { getEnvValue } from 'src/config/utils/envs';
 import type { Feature } from 'src/config/utils/features';
 
+type TabId = FlashblocksTabId | 'mini-blocks';
+
 const title = 'Flashblocks';
 
 const socketUrl = getEnvValue('NEXT_PUBLIC_FLASHBLOCKS_SOCKET_URL');
 const nameFromEnv = getEnvValue('NEXT_PUBLIC_FLASHBLOCKS_NAME');
 const opStackName = FLASHBLOCKS_NAMES.find((name) => name === nameFromEnv) ?? 'subblock';
-
-// The configured name owns the blocks-page tab; the other OP Stack name stays an alias so links
-// written under either name keep resolving (the `blocksTab` guard redirects it to the first id).
-const opStackTabIds = [ opStackName, ...FLASHBLOCKS_NAMES.filter((name) => name !== opStackName) ].map((name) => `${ name }s`);
+const opStackTabIds: [ FlashblocksTabId, ...Array<FlashblocksTabId> ] = opStackName === 'subblock' ?
+  [ 'subblocks', 'flashblocks' ] :
+  [ 'flashblocks', 'subblocks' ];
+const megaEthTabIds: [ TabId ] = [ 'mini-blocks' ];
 
 const config: Feature<{
   socketUrl: string;
   type: 'optimism' | 'megaEth';
   name: FlashblocksName | 'mini-block';
-  tabIds: Array<string>;
+  tabIds: [ TabId, ...Array<TabId> ];
 }> = (() => {
   if (megaEthFeature.isEnabled && megaEthFeature.socketUrl.rpc) {
     return Object.freeze({
@@ -31,7 +33,7 @@ const config: Feature<{
       socketUrl: megaEthFeature.socketUrl.rpc,
       type: 'megaEth',
       name: 'mini-block',
-      tabIds: [ 'mini-blocks' ],
+      tabIds: megaEthTabIds,
     });
   }
 
--- a/src/features/flashblocks/config.ts
+++ b/src/features/flashblocks/config.ts
@@ -18,7 +18,7 @@ const opStackName = FLASHBLOCKS_NAMES.find((name) => name === nameFromEnv) ?? 's
 const opStackTabIds: [ FlashblocksTabId, ...Array<FlashblocksTabId> ] = opStackName === 'subblock' ?
   [ 'subblocks', 'flashblocks' ] :
   [ 'flashblocks', 'subblocks' ];
-const megaEthTabIds: [ TabId ] = [ 'mini-blocks' ];
+const megaEthTabIds: [ 'mini-blocks' ] = [ 'mini-blocks' ];
 
 const config: Feature<{
   socketUrl: string;
--- a/src/features/flashblocks/types/config.ts
+++ b/src/features/flashblocks/types/config.ts
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: LicenseRef-Blockscout
+
+// What an OP Stack chain calls its sub-second pre-confirmation blocks. Chains are split between
+// the legacy "Flashblocks" name and the newer "Subblocks" name while running the same wire
+// protocol, so the operator picks the label per deployment via NEXT_PUBLIC_FLASHBLOCKS_NAME.
+export const FLASHBLOCKS_NAMES = [
+  'flashblock',
+  'subblock',
+] as const;
+
+export type FlashblocksName = (typeof FLASHBLOCKS_NAMES)[number];
--- a/src/features/flashblocks/types/config.ts
+++ b/src/features/flashblocks/types/config.ts
@@ -1,11 +1,15 @@
 // SPDX-License-Identifier: LicenseRef-Blockscout
 
-// What an OP Stack chain calls its sub-second pre-confirmation blocks. Chains are split between
-// the legacy "Flashblocks" name and the newer "Subblocks" name while running the same wire
-// protocol, so the operator picks the label per deployment via NEXT_PUBLIC_FLASHBLOCKS_NAME.
 export const FLASHBLOCKS_NAMES = [
   'flashblock',
   'subblock',
 ] as const;
 
 export type FlashblocksName = (typeof FLASHBLOCKS_NAMES)[number];
+
+export const FLASHBLOCKS_TAB_IDS = [
+  'flashblocks',
+  'subblocks',
+] as const;
+
+export type FlashblocksTabId = (typeof FLASHBLOCKS_TAB_IDS)[number];
--- a/src/server/getServerSideProps/guards.spec.ts
+++ b/src/server/getServerSideProps/guards.spec.ts
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: LicenseRef-Blockscout
+
+import type { GetServerSidePropsContext } from 'next';
+
+import type config from 'src/config';
+
+import { describe, expect, it } from 'vitest';
+
+import { blocksTab } from './guards';
+
+type FlashblocksFeature = (typeof config)['features']['flashblocks'];
+
+const SUBBLOCKS_FEED: FlashblocksFeature = {
+  title: 'Flashblocks',
+  isEnabled: true,
+  socketUrl: 'wss://localhost:3120/ws',
+  type: 'optimism',
+  name: 'subblock',
+  tabIds: [ 'subblocks', 'flashblocks' ],
+};
+const FLASHBLOCKS_FEED: FlashblocksFeature = { ...SUBBLOCKS_FEED, name: 'flashblock', tabIds: [ 'flashblocks', 'subblocks' ] };
+const DISABLED_FEED: FlashblocksFeature = { title: 'Flashblocks', isEnabled: false };
+
+// the guard reads only the flashblocks feature and the request query, so the rest of both can stay absent
+function runBlocksTab(flashblocks: FlashblocksFeature, query: Record<string, string | Array<string>>): ReturnType<ReturnType<typeof blocksTab>> {
+  const chainConfig = { features: { flashblocks } } as unknown as typeof config;
+  const context = { query } as unknown as GetServerSidePropsContext;
+  return blocksTab(chainConfig)(context);
+}
+
+describe('blocksTab guard', () => {
+  it('does nothing while the feed is disabled', async() => {
+    expect(await runBlocksTab(DISABLED_FEED, { tab: 'flashblocks' })).toBeUndefined();
+  });
+
+  it('lets the canonical tab id through', async() => {
+    expect(await runBlocksTab(SUBBLOCKS_FEED, { tab: 'subblocks' })).toBeUndefined();
+  });
+
+  it('ignores the other tabs', async() => {
+    expect(await runBlocksTab(SUBBLOCKS_FEED, { tab: 'reorgs' })).toBeUndefined();
+    expect(await runBlocksTab(SUBBLOCKS_FEED, {})).toBeUndefined();
+  });
+
+  it('redirects the alias to the canonical id with a temporary redirect', async() => {
+    expect(await runBlocksTab(SUBBLOCKS_FEED, { tab: 'flashblocks' })).toEqual({
+      redirect: { destination: '/blocks?tab=subblocks', permanent: false },
+    });
+  });
+
+  it('follows the configured name in the other direction', async() => {
+    expect(await runBlocksTab(FLASHBLOCKS_FEED, { tab: 'subblocks' })).toEqual({
+      redirect: { destination: '/blocks?tab=flashblocks', permanent: false },
+    });
+  });
+
+  it('keeps the other search params', async() => {
+    expect(await runBlocksTab(SUBBLOCKS_FEED, { tab: 'flashblocks', page: '2' })).toEqual({
+      redirect: { destination: '/blocks?tab=subblocks&page=2', permanent: false },
+    });
+  });
+});
--- a/src/server/getServerSideProps/guards.spec.ts
+++ b/src/server/getServerSideProps/guards.spec.ts
@@ -2,60 +2,56 @@
 
 import type { GetServerSidePropsContext } from 'next';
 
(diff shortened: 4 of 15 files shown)


src/features/flashblocks/config.ts
```diff
@@ -8,15 +11,27 @@ import type { Feature } from 'src/config/utils/features';
...
 
 const socketUrl = getEnvValue('NEXT_PUBLIC_FLASHBLOCKS_SOCKET_URL');
-
-const config: Feature<{ socketUrl: string; type: 'optimism' | 'megaEth'; name: string }> = (() => {
+const nameFromEnv = getEnvValue('NEXT_PUBLIC_FLASHBLOCKS_NAME');
+const opStackName = FLASHBLOCKS_NAMES.find((name) => name === nameFromEnv) ?? 'subblock';
+
+// The configured name owns the blocks-page tab; the other OP Stack name stays an alias so links
+// written under either name keep resolving (the `blocksTab` guard redirects it to the first id).
+const opStackTabIds = [ opStackName, ...FLASHBLOCKS_NAMES.filter((name) => name !== opStackName) ].map((name) => `${ name }s`);
```
tom2drum:
```suggestion
const opStackTabIds = FLASHBLOCKS_NAMES.map((name) => `${ name }s`);
```

I don't think we need a comment here.

src/server/getServerSideProps/guards.ts
```diff
@@ -355,3 +356,25 @@ export const megaEth: Guard = () => async() => {
     };
   }
 };
+
+// The flashblocks tab answers to both OP Stack ids (see NEXT_PUBLIC_FLASHBLOCKS_NAME); a link
+// using the alias is sent to the canonical id. Temporary, because the operator can flip the name.
+export const blocksTab: Guard = (chainConfig: typeof config) => async(context) => {
```
tom2drum:
```suggestion
// The flashblocks tab answers to both OP Stack ids (see NEXT_PUBLIC_FLASHBLOCKS_NAME); a link
// using the alias is sent to the canonical id. Temporary, because the operator can flip the name.
export const blocks: Guard = (chainConfig: typeof config) => async(context) => {
```

Technically, it guards the blocks page, so this name is more appropriate. In my view, the comment is unnecessary because it merely describes what the code does.

src/features/flashblocks/config.ts
```diff
@@ -8,15 +11,27 @@ import type { Feature } from 'src/config/utils/features';
...
+
+const config: Feature<{
+  socketUrl: string;
+  type: 'optimism' | 'megaEth';
+  name: FlashblocksName | 'mini-block';
+  tabIds: Array<string>;
```
tom2drum: **F1 · request · standards** — the tab ids are URL-contract values, but minted by interpolation and typed as `Array<string>`, so `'subblocks'` and `'flashblocks'` appear nowhere greppable in `src/`.

This PR already introduced `FLASHBLOCKS_NAMES` / `FlashblocksName` using the repo's const-array + derived-type pattern for the *names*. Could the *ids* get the same treatment — a `FLASHBLOCKS_TAB_IDS` / `FlashblocksTabId` pair in `types/config.ts` with the ids written out literally — and the field typed as a non-empty tuple?

```ts
tabIds: [ FlashblocksTabId | 'mini-blocks', ...Array<FlashblocksTabId | 'mini-blocks'> ];
```

The non-empty tuple also makes the destructuring at `guards.ts:369` sound rather than sound-by-construction. This repo does not enable `noUncheckedIndexedAccess`, so with a plain `Array<string>` a future empty `tabIds` types `canonicalTabId` as `string` while being `undefined` at runtime, and the guard silently redirects to a bare `/blocks` with the tab dropped. The tuple makes that unrepresentable.

Two notes from checking this against the compiler before asking:

- Keep the tuple **mutable**. A `readonly` tuple (i.e. reaching for `as const`) is not assignable to `TabItemRegular['id']`, which is `string | Array<string>`, so `id={ flashblocksFeature.tabIds }` in `Blocks.tsx` fails with TS2322.
- Narrowing the element type to the literal union costs nothing at the call sites. `Array<T>.includes` is declared as a method and TS method parameters are bivariant, so `tabIds.includes(tab)` with `tab: string` still compiles at `Blocks.tsx:73`.

Writing the ids out literally does mean dropping the `.map((name) => `${ name }s`)` derivation, which is the greppability win.

— Reviewed by Claude Opus 5

tom2drum (commented):
## Review summary

**Approve with comments.** No blockers, no functional defects. One request and four nits, all inline.

Thanks for the detailed description — it made this fast to review.

### Verification

Re-ran every claim in the description rather than taking it on trust. All of it holds:

| Claim | Result |
| --- | --- |
| `tsc` clean | confirmed |
| `eslint` clean | confirmed (0 errors; 7 pre-existing warnings in untouched `*.pw.tsx`) |
| `cspell` clean | confirmed, 3188 files — no `subblock`-family whitelisting needed |
| full vitest suite | confirmed, 93 files / 663 tests |
| 11 new tests in 2 new spec files | confirmed |
| CC / CRAP gate, new guard covered | confirmed — `blocksTab` 100% coverage, CRAP 1.0 |
| envs-validator presets | confirmed, all 10 fixtures |
| negative: unknown name rejected | confirmed — `must be one of the following values: flashblock, subblock` |
| negative: name without socket URL rejected | confirmed — `can only be used with NEXT_PUBLIC_FLASHBLOCKS_SOCKET_URL` |
| live redirect table, both directions | confirmed on a real dev server, extra query params preserved |

The yup sibling `when(...)` conditional does fire across the `.concat()`ed schemas — that was the part most likely to be subtly broken, so it was checked by running the validator rather than by reading it. Moving `NEXT_PUBLIC_FLASHBLOCKS_SOCKET_URL` into the new feature sub-schema drops no validation.

Extra probes beyond the claims: wrong-case (`Subblock`), empty `NEXT_PUBLIC_FLASHBLOCKS_NAME=`, and empty `NEXT_PUBLIC_FLASHBLOCKS_SOCKET_URL=` are all rejected; neither-var-set passes. The empty-value rejection matches existing repo convention (`NEXT_PUBLIC_NAVIGATION_LAYOUT=` behaves the same), so it is not a defect.

Two things not verified, stated plainly: the rendered tab **labels** (tabs hydrate client-side, so the served HTML carries no tab text), and whether the full-repo complexity failures pre-date this branch — all 86 sit in `src/toolkit/pages/design-system/**`, untouched here, and that run is not what CI gates.

### On the design

Right calls throughout. Defaulting to `subblock` keeps every current deployment's label, so the change is genuinely additive. MegaETH's `mini-block` name and `mini-blocks` tab are correctly left alone, with no alias and no redirect. The parsers, socket hook and table components don't branch on the name, which is correct given the identical wire format. `QUEUE_TIME_THRESHOLD` staying at 150 for both is also right — it is below either interval.

The array tab `id` is the established toolkit pattern (`TabItemRegular.id` is already `string | Array<string>`, first id wins on switch) rather than a new invention, and 307 over 301 is the right choice given an operator can flip the name later. Fixing the `?tab=` link that used to fall through to the "All" tab is a welcome extra.

None of the comments below block merge. Happy to push the one requested change ourselves if you would rather not.

src/server/getServerSideProps/guards.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+  type: 'optimism',
+  name: 'subblock',
+  tabIds: [ 'subblocks', 'flashblocks' ],
+};
+const FLASHBLOCKS_FEED: FlashblocksFeature = { ...SUBBLOCKS_FEED, name: 'flashblock', tabIds: [ 'flashblocks', 'subblocks' ] };
+const DISABLED_FEED: FlashblocksFeature = { title: 'Flashblocks', isEnabled: false };
+
+// the guard reads only the flashblocks feature and the request query, so the rest of both can stay absent
+function runBlocksTab(flashblocks: FlashblocksFeature, query: Record<string, string | Array<string>>): ReturnType<ReturnType<typeof blocksTab>> {
+  const chainConfig = { features: { flashblocks } } as unknown as typeof config;
```
tom2drum: **F2 · nit · standards** — `as unknown as typeof config` is the only cast of its kind in `src/`, and `.agents/rules/tests-unit.md` points at the env-override helpers instead (the same `withEnvs` approach the sibling `config.spec.ts` in this PR uses).

Minor and partly self-mitigating: the fixtures here hand-copy the canonical-first `tabIds` ordering that is itself under test, but `config.spec.ts` asserts that ordering independently for both names, so the coverage is not circular. Flagging it as a convention drift rather than a correctness problem.

— Reviewed by Claude Opus 5

.agents/GLOSSARY.md
```diff
@@ -39,7 +39,7 @@ vars are documented in `docs/ENVS.md`. Architectural concepts like
 | **Eden** | chain | A rollup built on `ev-reth` / evstack. Introduces the **Sponsored Transaction** type. |
 | **Epoch** | entity | A consensus time period specific to **Celo**. Has its own index and detail pages. Always refers to a Celo epoch in this codebase — not a generic blockchain concept. |
 | **Fault Proof System** | feature | Optimism's mechanism for proving the correctness of L2 state transitions on L1 via **Dispute Games**. |
-| **Flashblocks** | feature | Code name for the sub-second pre-confirmation block streaming feature. Surfaced to users as **Subblocks** on OP Stack chains (OP Labs renamed it from "Flashblocks") and as **mini-blocks** on MegaETH. The `flashblocks` code name and `FLASHBLOCKS` env vars are retained. |
+| **Flashblocks** | feature | Code name for the sub-second pre-confirmation block streaming feature. Surfaced to users as **Subblocks** (the default) or **Flashblocks** on OP Stack chains, whichever name the chain uses, and as **mini-blocks** on MegaETH. The `flashblocks` code name and `FLASHBLOCKS` env vars are retained. |
```
tom2drum: **F3 · nit · spec** — two small drifts from the glossary's own contract (see `.agents/skills/update-glossary/SKILL.md`):

- The rewrite drops the etymology the entry used to carry — that OP Labs renamed the feature from "Flashblocks" — which is one of the file's stated inclusion criteria and the main reason a reader hits this entry confused.
- "(the default)" states an env var default, which the glossary header defers to `docs/ENVS.md`. The new `docs/ENVS.md` row already says it.

— Reviewed by Claude Opus 5

src/features/flashblocks/types/config.ts
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: LicenseRef-Blockscout
+
+// What an OP Stack chain calls its sub-second pre-confirmation blocks. Chains are split between
+// the legacy "Flashblocks" name and the newer "Subblocks" name while running the same wire
+// protocol, so the operator picks the label per deployment via NEXT_PUBLIC_FLASHBLOCKS_NAME.
```
tom2drum: **F4 · nit · standards** — "the two names run the same wire protocol, only the labels differ" is now stated in four places this diff adds or touches: here, `guards.ts:360`, `docs/ENVS.md:962` and `.agents/GLOSSARY.md:42`.

All four are correct today, which is exactly when they are cheapest to consolidate. One canonical statement (`docs/ENVS.md` reads like the natural home for operators) with the code comments pointing at it would age better.

— Reviewed by Claude Opus 5

src/server/getServerSideProps/guards.ts
tom2drum: **F5 · nit · standards** — `blocksTab` names the page it hangs off rather than what it does, and it is the only guard in this file that redirects; every sibling gates access with `notFound`. A reader scanning the list will assume this one also blocks the tab.

`flashblocksTabAlias` (or `blocksTabAlias`) would read truer. The `blocks` export in `main.ts` is well named either way.

— Reviewed by Claude Opus 5

src/features/flashblocks/config.ts
pcw109550: Thanks. Taken as written, the suggestion would drop the canonical-first ordering: mapping `FLASHBLOCKS_NAMES` always yields `flashblocks, subblocks`, so under the default name the tab would switch to `?tab=flashblocks` and the guard would redirect the wrong way, which the config spec pins. I addressed the readability point through F1 instead: the two orderings are now written out literally, so the comment is gone.

src/server/getServerSideProps/guards.ts
pcw109550: Renamed to `blocks` and dropped the comment. F5 suggested `flashblocksTabAlias`; I went with `blocks` to match how the other guards are named for their page, and the existing `blocks` export in `main.ts`.

src/features/flashblocks/config.ts
pcw109550: Done: `FLASHBLOCKS_TAB_IDS` / `FlashblocksTabId` in `types/config.ts`, literal ids in `config.ts`, and `tabIds` typed as a non-empty mutable tuple. One correction to the compiler notes: `tabIds.includes(tab)` with `tab: string` does not compile once the element type is the literal union, since call-site arguments are checked by assignability rather than bivariance (TS2345), so both call sites use `some((id) => id === tab)`.

src/server/getServerSideProps/guards.spec.ts
pcw109550: Switched the spec back to `withEnvs`, matching `config.spec.ts` and the unit-test rule.

.agents/GLOSSARY.md
pcw109550: Restored the etymology and dropped '(the default)'.

src/features/flashblocks/types/config.ts
pcw109550: Removed the code comments; `docs/ENVS.md` is now the single statement of the rationale.

src/server/getServerSideProps/guards.ts
pcw109550: Went with `blocks`, see https://github.com/blockscout/frontend/pull/3681#discussion_r3950769654: it matches how the other guards are named for their page and the existing `blocks` export in `main.ts`.

src/features/flashblocks/config.ts
tom2drum: **F1 — verified**, with a correction to my own note.

The structural ask is delivered: `FLASHBLOCKS_TAB_IDS` / `FlashblocksTabId` in `types/config.ts`, both orderings written out literally at `config.ts:18-20`, and `tabIds` typed `[ TabId, ...Array<TabId> ]` — so the destructuring at `guards.ts:367` is sound rather than sound-by-construction. That was the point of the ask.

On the `includes` / TS2345 exchange we were each wrong in a different place, so it is worth writing down.

My "method parameters are bivariant" reasoning was the wrong mechanism — you were right to push back on it. But the original claim still holds in this repo, for a reason neither of us named. I reverted both call sites to `.includes(tab)` on `2a78d17` and ran the repo's own `tsc --noEmit --incremental false -p tsconfig.json`: **exit 0, zero errors.**

The reason is the root `reset.d.ts`, which the tsconfig `include` glob picks up and which imports `@total-typescript/ts-reset`. Its `array-includes.d.ts` redeclares the method:

```ts
interface Array<T> {
  includes(searchElement: T | (TSReset.WidenLiteral<T> & {}), fromIndex?: number): boolean;
}
```

`WidenLiteral` maps a string-literal union to `string`, so `tab: string` is accepted. TS2345 is the correct answer for a bare tsconfig and the wrong one for this project's.

No action either way — `some((id) => id === tab)` reads fine and costs nothing. It just was not required.

— Reviewed by Claude Opus 5

src/server/getServerSideProps/guards.spec.ts
tom2drum: **F2 — verified.** The `as unknown as typeof config` fixture is gone; `runBlocksGuard` drives the real config graph through `withEnvs` with dynamic imports inside the callback, matching `config.spec.ts` and the helper's own contract. The hand-copied `tabIds` orderings went with it, which was the part I liked least.

The switch introduced one small cast of its own — raised separately as F6, non-blocking.

— Reviewed by Claude Opus 5

.agents/GLOSSARY.md
tom2drum: **F3 — verified.** The OP Labs rename is back in the row, and "(the default)" is gone with the default now stated only in `docs/ENVS.md`.

— Reviewed by Claude Opus 5

src/features/flashblocks/types/config.ts
tom2drum: **F4 — verified.** Grepping the rationale across the touched files now returns exactly one hit, `docs/ENVS.md:962`. The comments in `types/config.ts` and `guards.ts` are gone and the glossary row carries only the etymology.

— Reviewed by Claude Opus 5

src/server/getServerSideProps/guards.ts
tom2drum: **F5 — accepted, withdrawing it.** You are right and I was reaching. Every guard export in this file is named for the page or feature it hangs off — `internalTx`, `accountsLabelSearch`, `disputeGames`, `megaEth` — and none for what it does; `main.ts:7` already reads `factory([ guards.blocks ])`. `flashblocksTabAlias` would have been the only action-named guard in the file, which is a worse inconsistency than the one I was pointing at.

— Reviewed by Claude Opus 5

src/server/getServerSideProps/guards.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it('redirects the alias to the canonical id with a temporary redirect', async() => {
+    expect(await runBlocksGuard(ENVS_MAP.flashblocks, makeContext({ tab: 'flashblocks' }))).toEqual({
+      redirect: { destination: '/blocks?tab=subblocks', permanent: false },
+    });
+  });
+
+  it('follows the configured name in the other direction', async() => {
+    const envs = [ ...ENVS_MAP.flashblocks, [ 'NEXT_PUBLIC_FLASHBLOCKS_NAME', 'flashblock' ] as [ string, string ] ];
```
tom2drum: **F6 · nit** — the `as unknown as typeof config` that F2 asked about is gone, but a smaller cast took its place:

```ts
const envs = [ ...ENVS_MAP.flashblocks, [ 'NEXT_PUBLIC_FLASHBLOCKS_NAME', 'flashblock' ] as [ string, string ] ];
```

It is only needed because the array is bound to an unannotated `const` first, so the entry widens to `Array<string>`. `config.spec.ts:30` builds the same list and needs no cast, because it inlines it into the call and picks up the parameter's contextual type.

```suggestion
    const envs: Array<[ string, string ]> = [ ...ENVS_MAP.flashblocks, [ 'NEXT_PUBLIC_FLASHBLOCKS_NAME', 'flashblock' ] ];
```

— Reviewed by Claude Opus 5

src/features/flashblocks/config.ts
```diff
@@ -1,22 +1,39 @@
...
 
 const socketUrl = getEnvValue('NEXT_PUBLIC_FLASHBLOCKS_SOCKET_URL');
-
-const config: Feature<{ socketUrl: string; type: 'optimism' | 'megaEth'; name: string }> = (() => {
+const nameFromEnv = getEnvValue('NEXT_PUBLIC_FLASHBLOCKS_NAME');
+const opStackName = FLASHBLOCKS_NAMES.find((name) => name === nameFromEnv) ?? 'subblock';
+const opStackTabIds: [ FlashblocksTabId, ...Array<FlashblocksTabId> ] = opStackName === 'subblock' ?
+  [ 'subblocks', 'flashblocks' ] :
+  [ 'flashblocks', 'subblocks' ];
+const megaEthTabIds: [ TabId ] = [ 'mini-blocks' ];
```
tom2drum: **F7 · nit** — `megaEthTabIds` is annotated with the full union rather than what it holds:

```ts
const megaEthTabIds: [ TabId ] = [ 'mini-blocks' ];
```

`TabId` is `'flashblocks' | 'subblocks' | 'mini-blocks'`, so `[ 'subblocks' ]` would also typecheck here — on the one branch where the tab id is a fixed constant. `opStackTabIds` genuinely needs the wider element type because it holds two different orderings; this one does not.

Same principle F1 was raised on, so flagging it for consistency rather than because anything is broken today.

```suggestion
const megaEthTabIds: [ 'mini-blocks' ] = [ 'mini-blocks' ];
```

— Reviewed by Claude Opus 5

tom2drum (commented):
## Review round 2 — arbitration

Every finding from round 1 is closed. **Outcome: clear** — no blocker, no major, nothing holding up merge.

| Finding | Ruling |
| --- | --- |
| F1 — tab ids literal, typed, non-empty tuple | fixed |
| F2 — spec drives real config via `withEnvs` | fixed |
| F3 — glossary etymology restored, default dropped | fixed |
| F4 — rationale stated once, in `docs/ENVS.md` | fixed |
| F5 — guard naming | **reject accepted** — withdrawn |

Two corrections on my side, both on F1's thread and worth repeating here:

- My "method parameters are bivariant" argument was the wrong mechanism, and you were right to challenge it.
- The conclusion nonetheless holds for this repo, though not for the reason either of us gave. I reverted both call sites to `.includes(tab)` on `2a78d17` and ran `tsc --noEmit --incremental false -p tsconfig.json`: exit 0. The root `reset.d.ts` pulls in `@total-typescript/ts-reset`, which redeclares `Array<T>.includes` as taking `T | (WidenLiteral<T> & {})` — and `WidenLiteral` collapses a string-literal union to `string`. TS2345 is right for a bare tsconfig, wrong for this one. `some` is fine to keep; it just was not required.

Verified against the code at `2a78d17` rather than against the description. This round did not re-review anything the fixes did not touch.

### New this round

Two nits introduced **by the fixes themselves**, both inline. Neither blocks.

- **F6** `guards.spec.ts:47` — the removed cast came back smaller.
- **F7** `config.ts:21` — `megaEthTabIds` is typed wider than the value it holds.

**Counts** — blocker 0 · major 0 · nit 2. Round-1 findings: 4 fixed, 1 withdrawn.

— Reviewed by Claude Opus 5

src/server/getServerSideProps/guards.spec.ts
pcw109550: Applied as suggested; the annotation gives the inner pair its contextual type, so the cast is gone.

src/features/flashblocks/config.ts
pcw109550: Applied as suggested. Agreed it is the same principle as F1: the one fixed branch should not be able to hold another id.

src/server/getServerSideProps/guards.spec.ts
tom2drum: **F6 — verified.** `guards.spec.ts:47` now reads

```ts
const envs: Array<[ string, string ]> = [ ...ENVS_MAP.flashblocks, [ NEXT_PUBLIC_FLASHBLOCKS_NAME, 'flashblock' ] ];
```

with no cast. The annotation is the right one rather than a wider one that merely silences the error: `ENVS_MAP` is declared `Record<string, Array<[ string, string ]>>` (`src/config/test-utils/env-presets.ts:11`) so the spread elements already carry the pair type, and `withEnvs`'s first parameter is `Array<[ string, string ]> | undefined` (`vitest/utils/mockEnvs.ts:13`) — so the annotation matches the helper's own contract instead of restating it loosely.

The runtime values are byte-identical to the previous version, so the test still exercises the same branch: env-driven `flashblock` name, `?tab=subblocks` redirected to `?tab=flashblocks`.

— Reviewed by Claude Opus 5

src/features/flashblocks/config.ts
tom2drum: **F7 — verified.** `config.ts:21` is now `const megaEthTabIds: [ 'mini-blocks' ] = [ 'mini-blocks' ];`, so `[ 'subblocks' ]` no longer typechecks on the one branch where the id is a fixed constant — which was the F1 principle.

Checked the two things a narrowing like this usually breaks, and neither does:

- **Assignability into the field.** `tabIds: [ TabId, ...Array<TabId> ]` (`config.ts:27`) still accepts it — the tuple is mutable (no `readonly`, and the `Object.freeze` at `config.ts:30` only wraps the outer object in `Readonly<…>`, leaving the property's tuple type intact), length 1 satisfies the single required element, and `'mini-blocks'` is a member of `TabId`.
- **No newly-unused type.** `TabId` is still referenced at `config.ts:27`, and the `FlashblocksTabId` import at `config.ts:3` is still used at lines 11 and 18.

Downstream is untouched: `Blocks.tsx:73` and `:90` see the declared field type, not the narrowed const type, so `id={ flashblocksFeature.tabIds }` against `TabItemRegular['id']` is unchanged.

— Reviewed by Claude Opus 5

tom2drum (commented):
## Review round 3 — arbitration

Both findings from round 2 are verified. Every thread this review opened is now closed. **Outcome: clear.**

| Finding | Ruling |
| --- | --- |
| F6 — env tuple cast in the guard spec | verified — cast gone, annotation matches `withEnvs`' own parameter type, test branch unchanged |
| F7 — `megaEthTabIds` annotated with the full union | verified — narrowed to `[ 'mini-blocks' ]`, still assignable into `tabIds`, no type left unused |

### Counts

| | blocker | major | nit | total |
| --- | --- | --- | --- | --- |
| Raised this round | 0 | 0 | 0 | 0 |
| Open | 0 | 0 | 0 | 0 |
| Across all rounds | 0 | 1 | 6 | 7 |

One finding across the three rounds was withdrawn by the reviewer (F5), one was a request (F1) and the rest were nits. No blocker was raised at any point.

### Scope of this round

Arbitration only, per the skill: the single commit `713e138` and regressions it could have caused. Two one-line changes, checked against the four things that break under a type narrowing like F7's — field assignability, `Object.freeze`/`readonly` interaction, newly-unused imports, and the downstream `TabItemRegular['id']` constraint at `Blocks.tsx:73`/`:90`. **No regressions.**

CI owns lint, tsc and vitest; this review did not run them.

— Reviewed by Claude Opus 5

PR #3691, opened 2026-09-08
Let operators pick which color themes an explorer offers

### LINKED ISSUE #3686: Make available color themes configurable
Add an environment variable containing an array of allowed color themes so each explorer can expose only the configured choices. For Arc, enable light and dark themes and hide Dim and Midnight. When the array contains a single theme, apply it and hide the theme switcher.

Image

tom2drum:
## Description

Resolves #3686

Adds an optional `NEXT_PUBLIC_COLOR_THEMES` array so each explorer exposes only the themes it wants. Themes left out are unavailable, and a list with a single theme applies that theme and hides the switcher from the settings popover. Omitting the variable keeps all four themes and today's behavior byte for byte.

Two decisions worth surfacing. The array **filters** the canonical `light → dim → midnight → dark` order rather than reordering it, so operators can't scramble the swatch progression. And the available list feeds default resolution: if it can't satisfy `prefers-color-scheme` (every available theme shares one color mode), `colorTheme.default` becomes the darkest available theme so SSR and the primed cookie agree with what the client effect will pick — otherwise a single-theme instance flashes the wrong mode on every cold load. An instance that lists all four themes narrows nothing and keeps following the OS preference.

`NEXT_PUBLIC_COLOR_THEME_DEFAULT` is kept rather than superseded by "first item wins": it is set across live instance configs, and deprecating it would force a config migration everywhere for no user-visible gain. The validator now cross-checks it against the list.

The theme-applying effect stays inside `SettingsColorTheme`, which only skips its own markup below two themes — the popover mounts it eagerly (`lazyMount={false}`), so a hidden switcher still writes the CSS vars and sets the cookies. `colorThemeMiddleware` resets a cookie holding a theme the instance no longer offers, mirroring `addressFormat.ts`.

## Environment variables

- Added `NEXT_PUBLIC_COLOR_THEMES` to let an explorer offer only some of the four color themes, and to hide the theme switcher when it offers just one.
- `NEXT_PUBLIC_COLOR_THEME_DEFAULT` must now name one of the themes listed in `NEXT_PUBLIC_COLOR_THEMES`, when that variable is set.

## Minimum API version

None

None. The new variable is optional and its absence preserves current behavior. An instance whose `NEXT_PUBLIC_COLOR_THEME_DEFAULT` is not in its `NEXT_PUBLIC_COLOR_THEMES` list fails startup validation, but that combination can only arise once an operator adopts the new variable.

## Additional information

Covered by `src/shell/top-bar/config.spec.ts` and `src/shell/top-bar/settings/color-theme/utils.spec.ts` (18 tests): filtering, canonical order, unknown ids, and every default-resolution branch. The validator's two negative paths were confirmed by hand against `test/.env.base`.

Not yet verified visually — worth a look at the popover with two themes and with one (switcher gone, theme still applied, no wrong-mode flash on a cold load), plus a cookie holding an excluded theme.


src/server/middlewares/colorTheme.ts
```diff
@@ -2,16 +2,28 @@
...
   const colorModeCookie = req.cookies.get(cookiesLib.NAMES.COLOR_MODE);
+  const colorThemeCookie = req.cookies.get(cookiesLib.NAMES.COLOR_THEME);
 
-  if (!colorModeCookie) {
-    if (appConfig.shell.topBar.colorTheme.default) {
-      res.cookies.set(cookiesLib.NAMES.COLOR_MODE, appConfig.shell.topBar.colorTheme.default.colorMode, cookiesLib.getDefaultAttributes());
-      res.cookies.set(cookiesLib.NAMES.COLOR_THEME, appConfig.shell.topBar.colorTheme.default.id, cookiesLib.getDefaultAttributes());
+  const nextTheme = (() => {
+    if (colorThemeCookie && !isColorThemeAvailable(colorThemeCookie.value)) {
+      return defaultTheme ?? themes[0];
```
tom2drum, line 26: **🛑 F1 · blocker**

`return defaultTheme ?? themes[0]` resets an unavailable COLOR_THEME cookie without consulting COLOR_MODE. On `['light','dark']` (the documented example), `defaultTheme` is undefined and `themes[0]` is `light`, so a returning `midnight`/`dim` + `dark` cookie pair is rewritten to light. Set-Cookie lands before JS, so SettingsColorTheme then treats `light` as available and never runs `getDefaultColorTheme(nextColorMode)`. Existing users on instances that later restrict themes get permanently flipped to light.

If COLOR_MODE is present, pick `getDefaultColorTheme` for that mode (then the matching theme object); only use `defaultTheme ?? themes[0]` when there is no COLOR_MODE cookie.

— Reviewed by Cursor Grok 4.6

docs/ENVS.md
```diff
@@ -382,7 +382,8 @@ Meta tags, Open Graph, and SEO.
...
 | NEXT_PUBLIC_FEATURED_NETWORKS_ALL_LINK | `string` | Link to the all chains resource. Will be displayed at the bottom of featured networks list. | Works only if NEXT_PUBLIC_FEATURED_NETWORKS is set | - | `https://example.com` | v2.3.0+ |
 | NEXT_PUBLIC_FEATURED_NETWORKS_MODE | `tabs \| list` | Indicates how the networks are presented: in one list or in separate tabs. | Works only if NEXT_PUBLIC_FEATURED_NETWORKS is set | `list` | `tabs` | v2.5.0+ |
-| NEXT_PUBLIC_COLOR_THEME_DEFAULT | `'light' \| 'dim' \| 'midnight' \| 'dark'` | Preferred color theme of the app | - | - | `midnight` | v1.30.0+ |
+| NEXT_PUBLIC_COLOR_THEMES | `Array<'light' \| 'dim' \| 'midnight' \| 'dark'>` | Color themes the app offers in the settings menu. | - | all themes | `['light', 'dark']` | upcoming |
+| NEXT_PUBLIC_COLOR_THEME_DEFAULT | `'light' \| 'dim' \| 'midnight' \| 'dark'` | Preferred color theme of the app. Must be one of the themes listed in NEXT_PUBLIC_COLOR_THEMES, if that variable is set. | - | - | `midnight` | v1.30.0+ |
```
tom2drum: **💬 F2 · nit**

`.agents/rules/docs.md` — “Changing code a doc describes: fix or delete the stale lines in the same change.” Default stays `-` while `config.ts` now does `coversBothColorModes ? undefined : availableColorThemes.slice(-1)[0]` when DEFAULT is unset. NEXT_PUBLIC_COLOR_THEMES documents its implicit fallback (`all themes`); this row does not.

Document that omitting DEFAULT while COLOR_THEMES is light-only or dark-only selects the last (darkest) available theme.

— Reviewed by Cursor Grok 4.6

src/shell/top-bar/config.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+describe('color themes', () => {
+  it('offers every theme when NEXT_PUBLIC_COLOR_THEMES is not set', async() => {
+    const config = await loadConfig([]);
+
+    expect(getThemeIds(config)).toEqual(ALL_THEME_IDS);
+    expect(config.colorTheme.default).toBeUndefined();
+  });
+
+  it('offers only the configured themes', async() => {
+    const config = await loadConfig([ [ 'NEXT_PUBLIC_COLOR_THEMES', '["light","dark"]' ] ]);
```
tom2drum, line 28: **💬 F3 · nit**

`.agents/rules/code-quality.md` Magic numbers — “In tests, unexplained magic values should also be extracted into named constants.” `'["light","dark"]'` is repeated at lines 28, 56, 91; `'["dim","midnight"]'` is inlined here while `utils.spec.ts` already names `DIM_AND_MIDNIGHT`.

Lift repeated env tuples to named constants, matching `LIGHT_ONLY` / `DIM_AND_MIDNIGHT` in `utils.spec.ts`.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
🛑 **Blocked** — 1 blocker open

blocker 1 · major 0 · nit 2 · not anchorable 0

Axes: spec skipped · standards · correctness

— Reviewed by Cursor Grok 4.6

src/server/middlewares/colorTheme.ts
tom2drum: Fixed in 0eb6857ac.

Confirmed: on `["light","dark"]` the list covers both color modes, so `config.ts` leaves `default` undefined and the reset fell through to `themes[0]` — light. The Set-Cookie lands before JS, so `SettingsColorTheme` then finds `light` available and never re-resolves; the flip was permanent, not a flash.

`colorTheme.ts` now resolves the replacement from `getDefaultColorTheme(colorModeCookie.value)` when a COLOR_MODE cookie is present, and keeps `defaultTheme ?? themes[0]` only for the no-COLOR_MODE case. A returning `midnight` + `dark` pair now lands on `dark`.

docs/ENVS.md
tom2drum: Fixed in 0eb6857ac. The default cell for NEXT_PUBLIC_COLOR_THEME_DEFAULT now reads: "none, unless the themes offered by NEXT_PUBLIC_COLOR_THEMES cannot cover both light and dark `prefers-color-scheme`, in which case the last (darkest) one of them".

src/shell/top-bar/config.spec.ts
tom2drum: Deferred — left as it stands.

The repetition is real, but two details argue against churning the file for it: the anchor is off (line 28 is `["light","dark"]`, not `["dim","midnight"]`), and apart from `["light","dark"]` every tuple here is a distinct one-off whose literal value *is* what the test is about — naming them adds a hop without adding meaning. `utils.spec.ts` names its two because it reuses each across several cases.

src/server/middlewares/colorTheme.ts
tom2drum: F1 — verified

docs/ENVS.md
tom2drum: F2 — verified

src/shell/top-bar/config.spec.ts
tom2drum: F3 — deferred

tom2drum (commented):
✅ **Cleared** — 1 nit open

Axes: spec skipped · standards · correctness

— Reviewed by Cursor Grok 4.6

PR #3692, opened 2026-09-08
Add a configurable action button to ERC-20 token pages

### LINKED ISSUE #3685: Arc: add a buy button to ERC-20 token pages
Add a fully configurable token-page action button and configure it for Arc. 

Requirements:
- Make all button properties configurable: visibility, label, destination URL (+local/external type), logo icon, colors for light and dark themes.
- Desktop placement - beside the token title.
- Mobile placement - on a dedicated row below the title, after the tags.
- Support light and dark themes with configurable colors and logos for each.
- If URL type is external, the button should include a "leave site" icon in the end. 

Note: For the future, the URL should be able to include params: <url>?tokenAddress=<token_address>&chain=<chain_id> - this can be done in the next iterations, exact params- tbd. 

Arc button config: 
- Label: "Buy on Arc Portal" 
- Link: tbd 
- Color: Jelly 400 HEX #8656ef from https://brand.circle.com/d/M9z54TaEwsWL/circle#/circle/color-1  (both themes) 
- Logo: Same as https://testnet.arcscan.app/assets/configs/network_icon_dark.svg 

Draft reference: 

Image

tom2drum:
## Description

Resolves #3685

An instance can now place a call-to-action button on its ERC-20 token pages, pointing at an external product — Arc uses it for "Buy on Arc Portal". Everything a partner would want to control is config, not code: label, destination, per-color-mode logo and colors. On desktop the button is right-aligned in the token title row; on mobile it takes its own row below the tags.

- New `token-action-button` feature: config, validator schema, component, unit and visual tests.
- UTM params are appended at render time rather than at config load, so the token-specific query params planned for a later iteration have a place to go.
- The `[light, dark]` color-mode array shape was declared separately in three places; it is now one shared `ColorModeValue` type in `src/config/types.ts`, reused by the hero and highlights banners.
- `**/node_modules_linux` added to the Stryker ignore list — the Playwright Docker deps directory lands under `src/`, and copying it into the mutation sandbox fails with `ENOTSUP`.

## Environment variables

Added `NEXT_PUBLIC_TOKEN_ACTION_BUTTON_CONFIG` to show a configurable action button on ERC-20 token pages, carrying the button's label, destination URL, logo and colors for light and dark mode. The button appears only when the variable is set. In a multichain cluster it is cluster-level: one button across every chain in the cluster.

The value configured for Arc is in [this issue comment](https://github.com/blockscout/frontend/issues/3685#issuecomment-5589363679).

## Minimum API version

None.

None — the feature is off unless the variable is set.

## Additional information

None.


tom2drum (commented):
🛑 **Blocked** — 1 blocker, 1 major open

blocker 1 · major 1 · nit 1 · not anchorable 0

— Reviewed by Cursor Grok 4.6

deploy/tools/envs-validator/schema.ts
```diff
@@ -186,6 +186,7 @@ const schema = yup
   .concat(featuresSchemas.nameServicesSchema)
   .concat(featuresSchemas.rollupSchema)
   .concat(featuresSchemas.tacSchema)
+  .concat(featuresSchemas.tokenActionButtonSchema)
```
tom2drum, line 189: **🛑 F1 · blocker**

`docs/ENVS.md` says this variable is cluster-level ("one button, shown on the token pages of every chain in the cluster"), and this line concats `tokenActionButtonSchema` into the default schema, but `schema_multichain.ts` never does. That schema uses `.noUnknown(true)`, so setting `NEXT_PUBLIC_TOKEN_ACTION_BUTTON_CONFIG` on a cluster fails startup as an unknown env.

Concat `featuresSchemas.tokenActionButtonSchema` in `schema_multichain.ts` the same way, and cover it in `test/.env.multichain`. Rule: `deploy/tools/envs-validator/CONTEXT.md` — "If the variable applies in both modes, mirror the rule in `schema.ts` and `schema_multichain.ts`."

— Reviewed by Cursor Grok 4.6

src/features/token-action-button/components/TokenActionButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const logoSrc = useColorModeValue(
+    button?.logo?.[0],
+    button?.logo?.[1] || button?.logo?.[0],
+  );
+
+  if (!button) {
+    return null;
+  }
+
+  if (!isLoading && tokenType !== 'ERC-20') {
```
tom2drum, line 42: **⚠️ F2 · major**

`if (!isLoading && tokenType !== 'ERC-20') { return null; }` shows the button whenever `isLoading` is true, including on non-ERC-20 pages. `TokenPageTitle` passes `isLoading={ tokenQuery.isPlaceholderData }` and `tokenType={ tokenQuery.data?.type }`; `useTokenQuery` always uses `placeholderData: TOKEN_INFO_ERC_20`, so every NFT/collection page flashes a clickable ERC-20 CTA until real data arrives. The spec at `TokenActionButton.spec.tsx:51` (`renders while the token type is still unknown`) locks this in.

Hide unless the loaded type is ERC-20: `if (isLoading || tokenType !== 'ERC-20') return null;`. Invert that unit test.

— Reviewed by Cursor Grok 4.6

deploy/tools/envs-validator/schemas/features/tokenActionButton.ts
```diff
@@ -0,0 +1,45 @@
...
+});
+
+const buttonSchema: yup.ObjectSchema<TokenActionButtonConfig> = yup
+  .object()
+  .transform(replaceQuotes)
+  .json()
+  .shape({
+    text: yup.string().required(),
+    url: yup.string().test(urlTest).required(),
+    logo: yup.array().max(2).of(yup.string()),
```
tom2drum: **💬 F3 · nit**

`logo: yup.array().max(2).of(yup.string())` — logo is URL-shaped (`docs/ENVS.md`: "Urls of the logo") but has no `urlTest`. The same file already does `url: yup.string().test(urlTest).required()`.

Use `.of(yup.string().test(urlTest))`. Rule: `deploy/tools/envs-validator/CONTEXT.md` — "For any URL-shaped value, attach the custom `urlTest` from `utils.ts`."

— Reviewed by Cursor Grok 4.6

deploy/tools/envs-validator/schema.ts
tom2drum: **F1 · fix** — `10009e9db`

Confirmed and fixed. Reproduced the blocker first: with the concat removed, the multichain preset fails with `Unknown ENV variables were provided: NEXT_PUBLIC_TOKEN_ACTION_BUTTON_CONFIG`.

- `schema_multichain.ts` now concats `featuresSchemas.tokenActionButtonSchema`, in the alphabetical slot after `multichainButtonSchema`.
- `test/.env.multichain` carries the variable with the same example value as `.env.base`.

`pnpm --filter envs-validator test` — all 16 presets `👍 All good!`.

src/features/token-action-button/components/TokenActionButton.tsx
tom2drum: **F2 · reject**

The premise — a *clickable* CTA flashing on NFT pages — doesn't hold, because the skeleton is not interactive. `src/toolkit/theme/recipes/skeleton.recipe.ts` sets `pointerEvents: 'none'` on the loading state, together with `color: 'transparent'` and `visibility: hidden` on children. So while the token query is in placeholder state there is neither a visible nor a clickable button, just a placeholder block; when the query resolves with a non-ERC-20 type the element simply disappears. No contradiction with "ERC-20 only" — the button proper is never shown for a non-ERC-20 token.

That is the intended behaviour, and `TokenActionButton.spec.tsx` "renders while the token type is still unknown" locks in that intent deliberately, so it stays as it is.

One factual note for the record: the finding is right that `useTokenQuery` sets `placeholderData: TOKEN_INFO_ERC_20`, so `tokenType` is already `'ERC-20'` during loading and the `!isLoading` clause changes nothing on the `TokenPageTitle` path. It still earns its place for a caller that has no type yet — which is exactly the case the unit test covers.

deploy/tools/envs-validator/schemas/features/tokenActionButton.ts
tom2drum: **F3 · fix** — `10009e9db`

Promoted from nit and fixed — `logo` items now carry `urlTest`.

Negative path verified: `'logo':['not a url', …]` now fails with `Invalid schema were provided for NEXT_PUBLIC_TOKEN_ACTION_BUTTON_CONFIG: logo[0] is not a valid URL`.

Worth flagging for consistency, though deliberately left out of scope here: `schemas/features/highlights.ts` has the same shape — `side_img_url: yup.array().max(2).of(yup.string())` — and still carries no `urlTest`.

deploy/tools/envs-validator/schema.ts
tom2drum: F1 — verified

src/features/token-action-button/components/TokenActionButton.tsx
tom2drum: F2 — accepted, loading skeleton is not a clickable CTA (`pointerEvents: none`)

deploy/tools/envs-validator/schemas/features/tokenActionButton.ts
tom2drum: F3 — verified

tom2drum (commented):

— Reviewed by Cursor Grok 4.6

Issue #3675, opened 2026-08-31
Bring deploy/tools/ under ESLint and the root type-check


tom2drum:
## Description

`deploy/tools/` is excluded from every automated quality gate in the repo: ESLint ignores the
directory outright, and each of its packages is excluded from the root TypeScript project. Its
code should be held to the same standard as the rest of the repo.

## Details

- `eslint.config.mjs` lists `deploy/tools/` in its top-level `ignores`, so `pnpm lint:eslint`
  never reads it. The directory contains `.ts` and `.js` alike (e.g.
  `deploy/tools/affected-tests/index.js`, ~200 lines).
- `tsconfig.json` `exclude` lists every `deploy/tools/*` package — `envs-validator`,
  `favicon-generator`, `multichain-config-generator`,
  `essential-dapps-chains-config-generator`, `llms-txt-generator`, `feature-reporter` — so
  `pnpm lint:tsc` does not type-check any of them.
- Each package has its own toolchain (own `package.json` and `vite.config.ts`; `envs-validator`
  additionally has its own `node_modules` and TypeScript version), which is why the exclusions
  exist. Bringing them in needs a per-package decision, not a blanket removal.
- No package here has unit tests or emits a coverage report — `envs-validator`'s `test` script
  validates `.env` files through a shell script — so per-function coverage is structurally
  unavailable and the CRAP half of the complexity gate cannot apply.

Complexity measured with `pnpm test:code-complexity` (CRAP at 0% coverage shown for reference
only):

| function | cyclomatic | cognitive | CRAP @ 0% |
| --- | --- | --- | --- |
| `llms-txt-generator/generate-standard.ts` `generateStandard` | 28 | 25 | 812 |
| `llms-txt-generator/generate-pro-api.ts` `generateProApi` | 26 | 23 | 702 |
| `envs-validator/index.ts` `validateEnvs` | 12 | 14 | 156 |
| `essential-dapps-chains-config-generator/index.ts` `run` | 10 | 7 | 110 |

## Expected Behavior

- `pnpm lint:eslint` lints `deploy/tools/`, with per-package overrides where a package's
  toolchain requires them.
- Every `deploy/tools/*` package is type-checked, either by the root project or by a per-package
  check wired into CI.
- The four functions above are decomposed, or a decision is recorded for each.
- A recorded decision on whether `deploy/tools/` then enters the cognitive-complexity gate's
  scope; CRAP cannot apply there without a coverage source.

PR #3694, opened 2026-09-09
Rewrite the tools/ scripts in TypeScript so they are type-checked, tested and gated

### LINKED ISSUE #3674: Rewrite the .mjs scripts under tools/ in TypeScript
## Description

The three `.mjs` scripts under `tools/` are the only executable logic in that directory that
cannot be type-checked or unit-tested, and they carry the highest complexity of any tooling code
in the repo. Rewriting them in TypeScript brings them under the root type-check, makes them
testable by the existing vitest suite, and lets the complexity gate report meaningful coverage
for them.

While `check-doc-links.mjs` is open, it also picks up a scope extension: widening its check surface
to per-directory doc sets (see "Additional scope" below).

**Blocked on #3663.** The complexity gate, its caps, and its coverage of `tools/**` all arrive with
that issue. Its own docs are the reference for how to run the gate and what the caps mean — read them
at implementation time rather than working from anything quoted here.

## Details

Affected files:

- `tools/scripts/check-doc-links.mjs` — backs `pnpm lint:doc-links`
- `tools/profiling/aggregate-react-profile.mjs` — backs `pnpm profile:analyze`
- `tools/dev-server/sync-preset-lists.mjs` — backs `pnpm presets:lint` / `pnpm presets:sync`

The `.mjs` extension blocks three separate things:

- `tsconfig.json` `include` covers `**/*.ts` / `**/*.tsx`, so these files are never type-checked
  by `pnpm lint:tsc`.
- `vitest.config.ts` `include` is `[ '**/*.spec.ts', '**/*.spec.tsx' ]`, so a `*.spec.mjs` would
  never be discovered — they cannot be unit-tested without widening suite-wide test discovery.
- With no specs they report 0% coverage, so the CRAP half of the complexity gate is
  uninformative for them.

Once `tools/**` is in the complexity gate's scope, any function of theirs over a cap fails CI on a PR
that touches it — the gate is diff-scoped, so untouched code is never blocked. Which functions those
are, and whether each needs decomposing or just coverage, is what the gate reports when run against
these three files. Work from that output: both the caps and the counting model behind them are
calibrated and can move.

## Additional scope: extend `lint:doc-links` to per-directory doc sets

`check-doc-links.mjs` builds its check surface from the agent-config directories plus every tracked
file **named exactly** `CONTEXT.md`. A directory that documents itself across several markdown files
therefore gets only one of them checked: links *into* the others are verified, links *inside* them are
not, so their own relative references can break silently and nothing warns.

Extend the collection step to markdown in and below any directory that contains a `CONTEXT.md`, so a
directory's whole doc set — including any nested subdirectories of records — is checked together.

This is deferred here rather than done in place because the change lands in the same function this
issue already has to decompose.

## Expected Behavior

- All three scripts are `.ts`, run through the repo-local compile-on-run pattern already used by
  `tools/dev-server/fetch.sh`; their `pnpm` scripts behave identically from a caller's point of view.
- They are type-checked by `pnpm lint:tsc`.
- The complexity gate passes on all three: every function is within the cognitive cap for its class
  and within the CRAP cap. Decomposition and added coverage are the two levers — which one each
  function needs is whatever the gate reports.
- Co-located `*.spec.ts` files exercise them, so their coverage is real rather than 0%.
- `pnpm lint:doc-links` checks every markdown file in a directory that has a `CONTEXT.md`, nested
  subdirectories included, and resolves their relative references.
- `vitest.config.ts` `include` is left unchanged.

tom2drum:
## Description

Resolves #3674

Three `.mjs` scripts under `tools/` were the only executable logic there that no gate could see. The extension kept them out of the root `tsconfig.json`, so `pnpm lint:tsc` never type-checked them; a `*.spec.mjs` would not be discovered by `vitest.config.ts` without widening suite-wide test discovery, so they had no tests; and with no tests they reported 0% coverage, which made the CRAP half of the complexity gate uninformative for exactly the highest-complexity tooling code in the repo. All three are now TypeScript with co-located `*.spec.ts` files, type-checked by `pnpm lint:tsc` and passing `pnpm test:code-complexity`.

- `tools/scripts/check-doc-links.mjs` → `tools/doc-links/`, split by concern (surface collection, line stripping, reference resolution, per-line checking, reporting, entry) — its per-file check function was the one real decomposition target. Documented in a new `CONTEXT.md`, which the checker then checks itself.
- `tools/dev-server/sync-preset-lists.mjs` → `tools/dev-server/sync-presets/`, with its own `tsconfig.json` so an `outDir` does not relocate the env fetcher's emitted entry that the Dockerfile `COPY`s. `registry.json` and `envs-rules.json` stay at the `tools/dev-server/` root, where the env fetcher and several docs still reference them.
- `tools/profiling/aggregate-react-profile.mjs` → `tools/profiling/aggregate.ts`, keeping its single-module shape; it breaks no cognitive cap, so its lever was coverage. Fixtures are hand-built rather than a committed DevTools export, which would pin them to one DevTools version.

Each tool gets a `run.sh` compile-on-run wrapper, and its `dist/` is git-ignored — no build artifact is committed. `pnpm lint:doc-links`, `pnpm profile:analyze`, `pnpm presets:lint` and `pnpm presets:sync` take the same arguments and produce the same output and exit codes as before.

`pnpm lint:doc-links` also gained scope: it now checks every markdown file in and below any directory containing a `CONTEXT.md`, not just the `CONTEXT.md` itself, so relative references inside a directory's other docs can no longer break silently. Two references that the extension newly covers did not resolve — `tools/code-complexity/docs/RUNNING.md` and `tools/mutation-testing/docs/RUNNING.md` both wrote `./run.sh` from a `docs/` subdirectory where the file sits one level up — and are fixed as docs rather than by resolving paths against the `CONTEXT.md` directory, which would weaken the check. `src/toolkit/package/` is exempt by path: its readme addresses npm package consumers, not agents.

## Environment variables

None.

## Minimum API version

None.

None. Developer tooling only — the `pnpm` script names are unchanged, and the Docker image build, the container entrypoint and `pnpm dev:preset` / `dev:local` / `prod:preset` are untouched.

## Additional information

The plan: [`.agents/tasks/3674-tools-scripts-rewrite/spec.md`](https://github.com/blockscout/frontend/blob/issue-3674/.agents/tasks/3674-tools-scripts-rewrite/spec.md)

Out of scope and tracked separately: moving the env fetcher out of `tools/dev-server/` (its path is a contract in the Docker build, the entrypoint and several scripts), and bringing `deploy/` under the complexity gate (#3675).


tom2drum (commented):

blocker 0 · major 0 · nit 3 · not anchorable 0

— Reviewed by Cursor Grok 4.6

tools/doc-links/CONTEXT.md
```diff
@@ -0,0 +1,39 @@
...
+`pnpm lint:doc-links` resolves the cross-references in the agent instruction surface — markdown links,
+heading anchors, and the file and directory paths written in backticks. A CI check in
+`.github/workflows/checks.yml`.
+
+## Files
+
+| File | Role |
+|---|---|
+| `surface.ts` | Which files are checked: the walked config roots, the exclusions, and every markdown file in and below a directory that carries a `CONTEXT.md`, taken from the tracked-file list. Widening the check surface happens here. |
+| `strip.ts` | Blanks the regions of a line that are not references — fenced blocks, and the illustration marks. Owns `ILLUSTRATION_FORMS`. |
```
tom2drum: **💬 F1 · nit**

`.agents/rules/docs.md` says point at directories and files, not at symbols, and to leave algorithms out. The Files row names `ILLUSTRATION_FORMS` and restates what `strip.ts` does: "Blanks the regions of a line that are not references — fenced blocks, and the illustration marks. Owns `ILLUSTRATION_FORMS`." The constraint on line 21 repeats the symbol.

Point at `strip.ts` for the printed exemption-mark footer. Keep one-line roles in the map.

— Reviewed by Cursor Grok 4.6

tools/doc-links/check.ts
```diff
@@ -0,0 +1,88 @@
+// Reads one file's references and reports the ones that do not resolve. Two forms carry references: a path
```
tom2drum: **💬 F2 · nit**

`.agents/rules/code-quality.md` Comments: a comment is allowed only when it explains why the obvious reading is wrong. This header restates the Files row in `CONTEXT.md`: "Reads one file's references and reports the ones that do not resolve. Two forms carry references: a path in backticks, and a markdown link with an optional heading anchor." Same WHAT headers sit on `index.ts:9` and `resolve.ts:1`. `resolve.ts:53` is `// GitHub's heading slug.`

Keep the why comments (git-vs-hops, realpath, nearby, illustration marks). Delete the restatement headers.

— Reviewed by Cursor Grok 4.6

tools/profiling/aggregate.ts
```diff
@@ -55,24 +121,9 @@ function replayOperations(operations, names) {
...
-        } else {
-          // parentID, ownerID, displayNameStringID, keyStringID, compiledWithForget
-          const displayNameStringID = operations[i + 2];
-          i += 5;
-          if (displayNameStringID >= stringTable.length) {
-            throw new Error(`string id out of range: ${ displayNameStringID }`);
-          }
-          names.set(id, { name: stringTable[displayNameStringID] ?? 'Anonymous', type });
-        }
+      case 1: // ADD
```
tom2drum: **💬 F3 · nit**

`.agents/rules/code-quality.md` Magic numbers: extract into named `UPPER_SNAKE_CASE` constants. `ELEMENT_TYPE_ROOT` is already named, but the switch still uses `case 1: // ADD` through `case 7: // SET_SUBTREE_MODE`, with comments that restate what a name would say.

Name `OP_ADD` through `OP_SET_SUBTREE_MODE` and drop the case comments.

— Reviewed by Cursor Grok 4.6

tools/doc-links/CONTEXT.md
tom2drum: **fix** (partial) — `16e3b785a`

Dropped `Owns \`ILLUSTRATION_FORMS\`.` from the `strip.ts` row. The fact has one home now: the constraint below the map.

Kept the constraint on line 21 as it stands, symbol name included. `docs.md` earns a line for "what an editor here must keep true, and what breaks if they don't", and the thing that must stay true is precisely that *this one constant* is the sole statement of the marks and is printed on failure. Pointing at `strip.ts` instead would name the file the reader is already in and lose what the constraint is about — vaguer, not more rename-proof. A rename of the constant is caught by the reader who has to touch that line anyway.

tools/doc-links/check.ts
tom2drum: **fix** (partial) — `16e3b785a`

Confirmed and fixed. `resolve.ts`'s header was word-for-word its `CONTEXT.md` row, `check.ts`'s near enough:

- deleted the headers on `check.ts` and `resolve.ts` outright — the role lives in the file map
- trimmed `index.ts` to start at the why ("These files instruct agents rather than humans…"), dropping the sentence that restated the row first
- kept the `strip.ts` header, as you said

One sub-point rejected: `resolve.ts:53` `// GitHub's heading slug.` stays. That is not a restatement of `slugify` — it names the external spec the implementation has to match. Without it there is nothing saying the character classes and the backtick strip mirror GitHub's anchor generation rather than being an arbitrary local choice, and that is exactly what the code cannot say about itself.

tools/profiling/aggregate.ts
tom2drum: **fix** (partial) — `16e3b785a`

Named them: `OP_ADD` through `OP_SET_SUBTREE_MODE`, beside `ELEMENT_TYPE_ROOT`, and every `case` now reads through the name.

The case comments are reduced, not dropped. The name-only ones (`// ADD`, `// REMOVE_ROOT`) are gone since the constant says it. The operand layouts stay, retitled `// operands: count, ...ids`, because they are not restating the name — they are why `i += 2 + operations[i + 1]` advances by that much. Delete them and each stride constant becomes unexplainable without going to the react-devtools-shared source.

Also updated `aggregate.spec.ts:22`, whose fixture note told the reader to keep the records "in step with the parser's comments" — now the op codes and operand notes.

tools/doc-links/CONTEXT.md
tom2drum: F1 — verified

tools/doc-links/check.ts
tom2drum: F2 — verified

tools/profiling/aggregate.ts
tom2drum: F3 — verified

tom2drum (commented):

— Reviewed by Cursor Grok 4.6

PR #3695, opened 2026-09-09
Let operators set the label on the DeFi dropdown button

### LINKED ISSUE #3687: Make the Blockscout DeFi header button label configurable
Make the text of the existing Blockscout DeFi header button configurable while retaining its configurable action list.

Image

tom2drum:
## Description

Resolves #3687

The DeFi dropdown's trigger label was hardcoded — `Blockscout DeFi` on desktop, shortened to
`DeFi` below the `lg` breakpoint — even though the actions inside the dropdown have been
operator-configurable since v1.31.0. A new `NEXT_PUBLIC_DEFI_DROPDOWN_BUTTON_TEXT` makes the
label configurable too, leaving `NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS` untouched.

The variable is an object rather than a bare string on purpose: the top bar has far less room on
narrow screens, and naming a `mobile` field is what tells an operator that a second, shorter form
is worth setting. It falls back to `desktop` when omitted. Truncation was considered and rejected —
where a label gets clipped depends on the viewport, so the shorter form is the operator's call, not
a runtime guess. The validator rejects the variable unless `NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS` holds
at least two items, because a single item renders as a direct link labelled with that item's own
text and has no dropdown to name — set-but-ignored config fails at startup instead of silently.

## Environment variables

Added `NEXT_PUBLIC_DEFI_DROPDOWN_BUTTON_TEXT` to set the text on the button that opens the DeFi
dropdown, as `{ desktop, mobile? }` so a shorter label can be used where the top bar is narrow.

## Minimum API version

None.

None. With the variable unset the label defaults to `{ desktop: 'Blockscout DeFi', mobile: 'DeFi' }`,
which reproduces the previous rendering exactly at every width.

## Additional information

Test coverage is Vitest-only, since the change adds no new visual state: `config.ts` gets a spec for
label resolution (default, custom, mobile fallback, malformed JSON) and `DeFiDropdown.tsx` one for
the labels reaching the trigger. The one Playwright change is a forced selector fix — the trigger now
holds the label in two breakpoint-toggled spans, so `getByText(/DeFi/i)` matches twice and trips
strict mode; it becomes `getByRole('button', { name: /defi/i })`. No screenshots change.

The validator's three rejection paths (missing `desktop`, variable set with one item, variable set
with no items) were each confirmed by temporarily breaking `test/.env.base`.

### DIFF
--- a/deploy/tools/envs-validator/schemas/features/defiDropdown.ts
+++ b/deploy/tools/envs-validator/schemas/features/defiDropdown.ts
@@ -35,22 +35,17 @@ export const defiDropdownSchema = yup.object({
     .of(deFiDropdownItemSchema),
   NEXT_PUBLIC_DEFI_DROPDOWN_BUTTON_TEXT: yup
     .mixed()
-    .test(
-      'shape',
-      'Invalid schema were provided for NEXT_PUBLIC_DEFI_DROPDOWN_BUTTON_TEXT, it should have a required desktop and an optional mobile field',
-      (data) => data === undefined || deFiDropdownButtonTextSchema.isValidSync(data),
-    )
-    .test(
-      'requires-dropdown',
-      `NEXT_PUBLIC_DEFI_DROPDOWN_BUTTON_TEXT can only be used when NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS contains at least ${ MIN_ITEMS_FOR_DROPDOWN } items`,
-      function(data) {
-        if (data === undefined) {
-          return true;
-        }
-
-        const items = this.parent.NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS;
-
-        return Array.isArray(items) && items.length >= MIN_ITEMS_FOR_DROPDOWN;
-      },
-    ),
+    .when('NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS', {
+      is: (items: unknown) => Array.isArray(items) && items.length >= MIN_ITEMS_FOR_DROPDOWN,
+      then: (schema) => schema.test(
+        'shape',
+        'Invalid schema were provided for NEXT_PUBLIC_DEFI_DROPDOWN_BUTTON_TEXT, it should have a required desktop and an optional mobile field',
+        (data) => data === undefined || deFiDropdownButtonTextSchema.isValidSync(data),
+      ),
+      otherwise: (schema) => schema.test(
+        'not-exist',
+        `NEXT_PUBLIC_DEFI_DROPDOWN_BUTTON_TEXT can only be used when NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS contains at least ${ MIN_ITEMS_FOR_DROPDOWN } items`,
+        value => value === undefined,
+      ),
+    }),
 });
--- a/deploy/tools/envs-validator/schemas/features/defiDropdown.ts
+++ b/deploy/tools/envs-validator/schemas/features/defiDropdown.ts
@@ -1,8 +1,10 @@
 import * as yup from 'yup';
-import type { DeFiDropdownItem } from 'src/features/defi-dropdown/types/client';
+import type { DeFiDropdownButtonText, DeFiDropdownItem } from 'src/features/defi-dropdown/types/client';
 import type { IconName } from 'public/icons/name';
 import { urlTest } from '../../utils';
 import { replaceQuotes } from 'src/config/utils/envs';
+
+const MIN_ITEMS_FOR_DROPDOWN = 2;
 
 const deFiDropdownItemSchema: yup.ObjectSchema<DeFiDropdownItem> = yup
   .object({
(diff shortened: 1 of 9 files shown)


tom2drum (commented):
🛑 **Blocked** — 1 major open

blocker 0 · major 1 · nit 2 · not anchorable 0

— Reviewed by Cursor Grok 4.6

deploy/tools/envs-validator/schemas/features/defiDropdown.ts
```diff
@@ -16,10 +18,39 @@ const deFiDropdownItemSchema: yup.ObjectSchema<DeFiDropdownItem> = yup
...
+      function(data) {
+        if (data === undefined) {
+          return true;
+        }
+
+        const items = this.parent.NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS;
+
+        return Array.isArray(items) && items.length >= MIN_ITEMS_FOR_DROPDOWN;
+      },
+    ),
```
tom2drum: **⚠️ F1 · major**

Companion gate is a `.test` that reads `this.parent.NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS` instead of `.when(...)`.

`deploy/tools/envs-validator/CONTEXT.md` — "Companion variables: `.when(...)` with a guard. When variable B is only meaningful if variable A is set, gate B on A and forbid B in the otherwise-branch."

Use `.mixed().when('NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS', { is: (items) => Array.isArray(items) && items.length >= MIN_ITEMS_FOR_DROPDOWN, then: shape test, otherwise: not-exist test })`, as `NEXT_PUBLIC_MARKETPLACE_TITLES` does in `marketplace.ts`.

— Reviewed by Cursor Grok 4.6

src/features/defi-dropdown/config.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+import type deFiDropdownConfig from './config';
+
+const ITEMS_ENV: [ string, string ] = [
+  'NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS',
+  '[{"text":"Swap","dappId":"uniswap"},{"text":"Payment link","url":"https://example.com"}]',
+];
+
+async function loadConfig(envs: Array<[ string, string ]>): Promise<typeof deFiDropdownConfig> {
+  return withEnvs(envs, async() => {
+    await import('src/config');
```
tom2drum, line 18: **💬 F2 · nit**

`await import('src/config');` discards its result. The `withEnvs` example in `.agents/rules/tests-unit.md` imports `src/config` as the subject; this spec then also imports `./config`.

Drop that line; `return (await import('./config')).default` is enough (`src/features/flashblocks/config.spec.ts`).

— Reviewed by Cursor Grok 4.6

src/features/defi-dropdown/components/DeFiDropdown.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { afterEach, describe, expect, it } from 'vitest';
+import { cleanup, render, screen } from 'vitest/lib';
+import withEnvs from 'vitest/utils/mockEnvs';
+
+const ITEMS_ENV: [ string, string ] = [
+  'NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS',
+  '[{"text":"Swap","dappId":"uniswap"},{"text":"Payment link","url":"https://example.com"}]',
+];
+
+async function renderDropdown(envs: Array<[ string, string ]>) {
```
tom2drum: **💬 F3 · nit**

`async function renderDropdown(envs: Array<[ string, string ]>)` has no return type.

`.agents/rules/typescript.md` — "When declaring functions on the top-level of a module, declare their return types."

Annotate `Promise<void>`.

— Reviewed by Cursor Grok 4.6

deploy/tools/envs-validator/schemas/features/defiDropdown.ts
tom2drum: **fix** — rewritten as `.mixed().when('NEXT_PUBLIC_DEFI_DROPDOWN_ITEMS', { is: items => Array.isArray(items) && items.length >= MIN_ITEMS_FOR_DROPDOWN, then: shape test, otherwise: not-exist test })`, matching `NEXT_PUBLIC_MARKETPLACE_TITLES`. `pnpm --filter envs-validator test` passes on every preset.

One correction on severity: this was a convention breach, not a behavioural one. I probed both forms against the validator's yup over 5 cases (2 items + valid text, 2 items + text missing `desktop`, 1 item + text, no items + text, 2 items + no text) and they accept and reject identically — `this.parent` during validation already sees the cast (JSON-parsed) items array, same value `is` receives. So the fix is for the documented pattern, not for a defect.

Commit d395492

src/features/defi-dropdown/config.spec.ts
tom2drum: **reject** — the premise is wrong: that import is not dead, it primes the module graph.

I promoted this nit and deleted the line. All 5 tests in the file then failed with:

```
TypeError: Cannot read properties of undefined (reading 'baseUrl')
 ❯ src/shell/metadata/config.ts:12:19
 ❯ src/config/index.ts:4:1
 ❯ src/shared/storage/cookies.ts:5:1
```

There is a cycle: `./config` → `src/config/app` → `src/shared/storage/cookies` → `src/config` (index) → `src/shell/metadata/config`, which reads `app.baseUrl` while `app.ts` is still evaluating. Importing `src/config` first enters the graph at the top instead of mid-cycle, so `app` is fully evaluated before anything reads it.

`flashblocks/config.spec.ts` gets away without it because `flashblocks/config.ts` does not import `src/config/app` — it only takes `getEnvValue` from `src/config/utils/envs`. `defi-dropdown/config.ts` does import it, for the `isPrivateMode` check.

Mocking `src/config` was considered as the alternative and does work, but it couples the spec to the single line of config `cookies.ts` reads today (`config.app.protocol`), and `.agents/rules/tests-unit.md` steers away from hand-mocking `src/config`. Kept the import and added a comment saying why it cannot be removed.

Commit d395492

src/features/defi-dropdown/components/DeFiDropdown.spec.tsx
tom2drum: **fix** — annotated `Promise<void>`. Commit d395492

deploy/tools/envs-validator/schemas/features/defiDropdown.ts
tom2drum: F1 — verified

src/features/defi-dropdown/config.spec.ts
tom2drum: F2 — accepted, import primes a real config↔cookies cycle

src/features/defi-dropdown/components/DeFiDropdown.spec.tsx
tom2drum: F3 — verified

tom2drum (commented):
✅ **Cleared**

— Reviewed by Cursor Grok 4.6

PR #3698, opened 2026-09-10
Make the URL the single source of truth for paginated lists and stop redundant requests on navigation

### LINKED ISSUE #3697: Stop redundant requests and skeleton cycles in paginated list navigation
## Description

Every paginated list view (transactions, token transfers, logs, blocks, etc.) is driven by a shared pagination hook that keeps page, cursor, filter and sort state in three places: the URL, the hook's own state, and the calling component's state. Navigation handlers write to the URL and then re-sync state from it via effects, which produces redundant API requests, transient query keys that never correspond to a page the user asked for, and skeleton cycles for data that is already cached.

## Details

Measured on the address transactions tab (API requests per interaction):

| Interaction | Requests | Problem |
|---|---|---|
| "First" from page 3 | 3 | current page re-fetched (wasted), page 1 fetched twice, page-3 skeleton flashes before page 1 |
| Filter change while on page 3 | 2 | first request combines the old page cursor with the new filter |
| "Prev" from page 2 to 1 | 1 | page 1 was cached seconds earlier but is dropped and shown as skeleton |
| Any unrelated `router.query` change | 0 | an extra render of the whole list subtree |

Contributing factors:

- Cache clearing is prefix-scoped by resource name, so it discards cached pages for other addresses and chains as well, and it runs before the route change while the current page is still mounted.
- The hook's return value and the `pagination` object are new on every render, so memoization in list components and tab containers never holds.
- 14 callers duplicate filter state and 10 duplicate sort state that already lives in the URL, and some need manual resets on tab switch to stay consistent.
- Every page navigation performs two heavy commits: all rows unmount into skeleton rows, then remount with real data.
- Multichain concerns (chain selection state and effects) are part of the shared hook although only a small share of callers use them.

## Expected Behavior

- One navigation action (next, prev, first, filter, sort) results in exactly one API request and one render with the correct query key.
- Returning to a cached page does not drop the cache; a refresh, if needed, happens in the background.
- Cache operations are scoped to the affected query, not the whole resource.
- The hook's return value is referentially stable between renders when nothing changed, so downstream memoization works.
- Page navigation keeps the current rows visible (dimmed, non-interactive) until the next page arrives instead of a skeleton pass; skeletons remain for first load and filter/sort changes. Visual treatment to be confirmed with product and design.

tom2drum:
## Description

Resolves #3697

Every paginated list in the app (~106 call sites) is driven by one shared hook that kept page, cursor, filter and sort state in three places at once: the URL, its own React state, and the calling component. Handlers wrote the URL and then re-synced state through effects, which cost extra API requests per navigation, requests for page/filter combinations the user never asked for, and skeleton passes over pages that were already cached. Cache clearing was also prefix-scoped by resource name, so it dropped cached pages of other addresses and chains too.

This PR makes the URL the single source of truth for list state and rebuilds the hook around it:

- `useQueryWithPages` is replaced by `useApiPaginatedQuery` (renamed to sit next to `useApiQuery` / `useApiInfiniteQuery`), composed from two single-responsibility hooks in `src/shared/pagination/`: `usePaginationParams` (pure derivation of page, cursor, filters and sorting from `router.query`) and `usePaginationActions` (stable next/prev/first/filter/sort callbacks, one shallow `router.push` each, no `.then` callbacks or sync effects). The selected-chain value and its change handler moved out of the shared hook into the opt-in `useChainValue` under `src/features/multichain/hooks/`, so single-chain callers neither receive nor pay for them.
- One user action now results in exactly one URL push, one query key and one request. Prefix-scoped `removeQueries` calls are gone; page 1 keeps `staleTime: 0` for a background refresh on return, deeper pages stay cached, and cached pages render immediately without a skeleton.
- The hook's return value, its `pagination` object and every callback are referentially stable when inputs have not changed, so the existing `React.memo` in list components and tab containers holds across unrelated route changes and socket-driven cache updates.
- All callers that duplicated filter or sort state in `useState` now read typed values from the hook and call its change handlers; the transactions lists' fixed tab filters moved into `queryParams`. Tab switches reset filters and sorting by stripping the URL params, with no caller-side bookkeeping. A shared `useDebouncedFilterChange` helper covers the text-filter callers.
- Pagination URL params (`page`, `next_page_params`, filter and sort names and encodings) are unchanged, so existing links keep working.
- Vitest gains a live `next/router` stand-in that re-renders `useRouter` consumers on push, and the hook spec asserts request and render counts per action, not only the final state.

The dimmed, non-interactive page-transition state is deferred to #3704 pending design. Its plumbing ships inert: the hook exposes `isTransitioning` (always `false`) and `DataList` accepts the matching prop, while page changes keep the skeleton pass as before.

## Environment variables

None.

## Minimum API version

None.

None.

## Additional information

Spec: [`.agents/tasks/3697-paginated-query-hook-refactor/spec.md`](https://github.com/blockscout/frontend/blob/issue-3697/.agents/tasks/3697-paginated-query-hook-refactor/spec.md). The work landed one commit per ticket; the behavioural change and the mechanical rename are separate commits so the behavioural diff stays reviewable.

Measured on the address transactions tab, API requests per interaction:

| Interaction | Before | After |
| --- | --- | --- |
| "First" from page 3 | 3 | 1 |
| Filter change while on page 3 | 2 | 1 |
| "Prev" from page 2 to page 1 | 1 with skeleton | 0 blocking (cached rows, background refresh) |
| Unrelated `router.query` change | full list subtree re-render | no re-render |

https://claude.ai/code/session_01JZrDVtHh4X2JRzgXTRhRQE

### DIFF
--- a/.agents/tasks/3697-paginated-query-hook-refactor/questions.md
+++ b/.agents/tasks/3697-paginated-query-hook-refactor/questions.md
@@ -1,0 +1,15 @@
+# Open questions — Stop redundant requests and skeleton cycles in paginated list navigation
+
+### Q01 — Visual treatment of the page-transition state
+
+Page changes will keep the current rows on screen, dimmed and non-interactive, until the next
+page arrives, instead of a skeleton pass. What should that state look like?
+
+- Owner: Designer (Tatyana), with PM (Ulyana) confirming the behaviour change
+- Status: `pending`
+- Resolved when: (1) the dimmed treatment is specified — opacity value or token, whether a
+  fade transition is used and its duration; (2) confirmation that skeletons stay for first load and
+  filter/sort changes only; (3) confirmation that the sticky header and sort buttons remain active
+  during the transition.
+- Slack: —
+- Answer: —
--- a/.agents/tasks/3697-paginated-query-hook-refactor/spec.md
+++ b/.agents/tasks/3697-paginated-query-hook-refactor/spec.md
@@ -1,0 +1,141 @@
+# Stop redundant requests and skeleton cycles in paginated list navigation
+
+| | |
+| --- | --- |
+| Issue | https://github.com/blockscout/frontend/issues/3697 |
+| Feature branch | `issue-3697` |
+| PM | Ulyana |
+| Designer | Tatyana |
+| Backend | — |
+| Minimum API version | — |
+| Slack channel | — |
+
+## Context & goal
+
+Every paginated list in the app (address transactions, token transfers, logs, blocks, and ~100 more
+call sites) is driven by one shared hook, `useQueryWithPages`. An audit of that hook on the address
+transactions tab found that it is not a render-cost hotspot itself, but its design produces
+user-visible waste on every navigation: extra API requests, requests for page/filter combinations
+the user never asked for, and skeleton passes over data that is already cached.
+
+The root cause is that page, cursor, filter and sort state live in three places at once — the URL,
+the hook's own React state, and the calling component's state — and the hook writes the URL first
+and then re-syncs its state from the URL through effects. Measured on the address transactions tab:
+
+| Interaction | API requests today | Should be |
+| --- | --- | --- |
+| "First" from page 3 | 3 (page 3 refetched, page 1 twice; page-3 skeleton flashes first) | 1 |
+| Filter change while on page 3 | 2 (first one = old cursor + new filter) | 1 |
+| "Prev" from page 2 to page 1 | 1, with skeleton, although page 1 is cached | 0 blocking (background refresh at most) |
+| Unrelated `router.query` change | 0, but a full re-render of the list subtree | no re-render |
+
+The goal is to make the URL the single source of truth for list state, so that one user action is
+exactly one URL push, one query key, one request and one render; to stop discarding cached pages;
+to make the hook's return value stable so the memoization already present in list components and
+tab containers actually holds; and to replace the skeleton pass on page changes with keeping the
+current rows visible (dimmed) until the next page arrives.
+
+## Functional requirements
+
+1. Each pagination action — next page, previous page, first page, filter change, sorting change —
+   issues exactly one API request, for the query the user asked for, and never a request for a
+   transient combination (for example the previous page's cursor combined with a new filter).
+2. Navigating back to a page whose data is cached shows that data immediately, with no skeleton.
+   Page 1 may refresh in the background while its cached rows are displayed; it is never dropped
+   from the cache before being displayed.
+3. Cache operations performed by the hook affect only the queries of the list it drives (that
+   resource with those path params), never other addresses, chains, or lists sharing the resource
+   name.
+4. Page, cursor, filters, sorting and (on multichain instances) selected chain are read from the
+   URL on every render. The hook keeps no React state that mirrors a URL value, and no effect
+   re-syncs state from the URL. Reloading, sharing or navigating back to a URL restores the same
+   list state as today (page number, `next_page_params`, filter and sort params keep their current
+   names and encoding).
+5. Callers no longer keep their own copy of filter or sort state; they read typed values from the
+   hook and call its change handlers. Tab switches, which strip list params from the URL, therefore
+   reset filters and sorting without caller-side bookkeeping.
+6. The hook's return value, its `pagination` object and every callback it exposes are referentially
+   stable between renders when their inputs have not changed, so a memoized consumer does not
+   re-render on an unrelated route change or on a socket-driven cache update.
+7. When the user moves to another page of the same list (same filters and sorting, different
+   cursor), the current rows stay rendered in a dimmed, non-interactive state until the new page's
+   data arrives, then are replaced in one commit. Pagination controls are disabled during the
+   transition; the sticky table header and sorting controls remain usable.
+8. Skeleton rows are still shown on first load of a list and after a filter or sorting change. The
+   lazy-render window of long lists resets on every dataset change as it does today.
+9. All existing pagination behaviour that is not named above is preserved: the "next page" button
+   hides when the API returns no cursor, the "first" and "previous" buttons stay disabled on page 1,
+   scroll-to-top (or to the configured element) happens on every navigation, the socket-driven
+   "new items" notices keep prepending into page 1, and CSV export and count/label props keep
+   reading the same values.
+10. The composed hook is renamed to `useApiPaginatedQuery` to sit next to `useApiQuery` and
+    `useApiInfiniteQuery`; the rename is applied to every call site.
+11. The multichain-only capabilities (selected chain value and its change handler) are provided by
+    a separate opt-in hook; single-chain callers neither receive nor pay for them.
+12. The existing unit test coverage for the hook is kept or extended, with a realistic router
+    stand-in that re-renders on push so that the number of requests and renders per action is
+    asserted, not only the final state.
+
+## Data & API
+
+None. No endpoints, resources, env vars or feature flags change. Pagination params (`page`,
+`next_page_params`), filter and sorting query parameter names and encodings stay exactly as they are
+so that existing links keep working.
+
+## UI inventory
+
+- Every paginated list page and tab in the app (the ~106 call sites of the current hook), with the
+  address details page transactions tab (single chain) as the reference integration: `RoutedTabs`
+  → tab content → sorting wrapper → list content → table/list rows → `Pagination`.
+- No new screens. One new visual state: a **page-transition** state for the list body — current
+  rows dimmed and non-interactive, pagination buttons disabled, action bar and table header
+  unchanged. Exact opacity and transition to be agreed with the designer (see `questions.md`,
+  Q01); until then use the app's existing disabled-content look.
+- Skeleton state stays as is for first load and filter/sort changes.
+
+## Implementation decisions
+
+- **URL is the single source of truth.** All list state is derived from `router.query` on each
+  render; the previous-page cursors needed for "previous" are kept in a ref keyed by page number,
+  populated on "next", and are the only in-memory list state. Handlers perform one shallow
+  `router.push` each and nothing else; there are no `.then` callbacks, no `isMounted` timers, no
+  router-sync effects.
+- **Split into single-responsibility hooks, composed by a thin public hook.** `usePaginationParams`
+  reads page, cursor, filters and sorting from the URL (pure derivation, no state, no effects).
+  `usePaginationActions` returns stable next/prev/first/filter/sorting callbacks that read the
+  latest router through a ref. The composed `useApiPaginatedQuery` builds the query key, calls
+  `useApiQuery`, and returns a memoized object; it is what the vast majority of callers use with
+  the same call shape as today. `useChainValue` is the opt-in multichain piece. A caller that
+  needs only part (resource chosen from the filter, chain selection) composes the pieces itself.
+- **No prefix-scoped cache removal.** The `removeQueries([resourceName])` calls go away. Page 1 keeps
+  `staleTime: 0` so it refreshes in the background on return; deeper pages keep `staleTime:
+  Infinity`. If a forced refresh is ever needed it targets the exact query key.
+- **Placeholder strategy decided inside the hook.** `placeholderData` is a function that keeps the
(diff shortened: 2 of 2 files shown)


src/slices/address/pages/details/txs/AddressTxs.tsx
```diff
@@ -52,7 +52,7 @@ const AddressTxs = ({ shouldRender = true, isQueryEnabled = true }: Props) => {
     enabled: isQueryEnabled && isLocalTab,
   });
 
-  const crossChainQuery = useQueryWithPages({
+  const crossChainQuery = useApiPaginatedQuery({
```
tom2drum, line 55: **⚠️ F1 · major**

FR7 — the hook and the shared list wrapper carry an `isTransitioning` flag / prop wired to every migrated list but never set. This call site is migrated (`const crossChainQuery = useApiPaginatedQuery({`), but the tab still renders `<AddressTxsCrossChain … isLoading={ crossChainQuery.isPlaceholderData }>` with no `isTransitioning`. `AddressTxsCrossChain` spreads the rest into `DataList`. Until #3704 the flag is always false, so nothing looks wrong; once it flips, this list stays undimmed.

Pass `isLoading={ crossChainQuery.isInitialLoading }` and `isTransitioning={ crossChainQuery.isTransitioning }`.

— Reviewed by Cursor Grok 4.6

src/slices/address/pages/details/token-transfers/AddressTokenTransfers.tsx
```diff
@@ -55,7 +55,7 @@ const AddressTokenTransfers = ({ shouldRender = true, overloadCount, isQueryEnab
     enabled: isQueryEnabled && isLocalTab,
   });
 
-  const crossChainQuery = useQueryWithPages({
+  const crossChainQuery = useApiPaginatedQuery({
```
tom2drum, line 58: **⚠️ F2 · major**

Same FR7 gap on the sibling tab. The query is `const crossChainQuery = useApiPaginatedQuery({`, but `<TokenTransfersCrossChainContent … isLoading={ crossChainQuery.isPlaceholderData } />` never gets `isTransitioning`. The other two call sites (`TokenTransfersCrossChain.tsx`, `TxTokenTransferCrossChain.tsx`) both pass it.

Pass `isLoading={ crossChainQuery.isInitialLoading }` and `isTransitioning={ crossChainQuery.isTransitioning }`.

— Reviewed by Cursor Grok 4.6

src/slices/search/hooks/useSearchQuery.ts
```diff
@@ -11,27 +11,25 @@ import { SEARCH_RESULT_ITEM } from 'src/slices/search/stubs';
...
-  const initialValue = q.current;
-
-  const [ searchTerm, setSearchTerm ] = React.useState(initialValue);
-
-  const debouncedSearchTerm = useDebounce(searchTerm, 300);
   const pathname = router.pathname;
 
-  const query = useQueryWithPages({
+  const appliedSearchTerm = getQueryParamString(usePaginationParams('core:search').filters.q);
+  const [ searchTerm, setSearchTerm ] = React.useState(appliedSearchTerm);
```
tom2drum, line 26: **⚠️ F3 · major**

FR4/FR5 — list state is read from the URL on every render, and callers must not keep a copy of a filter. Results now follow `appliedSearchTerm` from the URL, but `const [ searchTerm, setSearchTerm ] = React.useState(appliedSearchTerm)` never updates after mount. Browser Back/Forward changes the results while the search input keeps the previous term. The same pattern is in `src/features/multichain/pages/search-results/useSearchQuery.ts`.

When `appliedSearchTerm` changes from outside the input (history, a shared URL), set the draft to match, without overwriting in-flight typing before debounce completes. Apply the same fix on the multichain hook.

— Reviewed by Cursor Grok 4.6

src/features/multichain/pages/search-results/useSearchQuery.ts
```diff
@@ -19,9 +19,8 @@ export default function useSearchQuery({ chainId }: Props) {
   const initialValue = q.current;
   const checkRedirect = getQueryParamString(router.query.redirect) === 'true';
 
-  const [ searchTerm, setSearchTerm ] = React.useState(initialValue);
-
-  const debouncedSearchTerm = useDebounce(searchTerm, 300);
+  const appliedSearchTerm = getQueryParamString(router.query.q);
+  const [ searchTerm, setSearchTerm ] = React.useState(appliedSearchTerm);
```
tom2drum, line 23: **⚠️ F4 · major**

Same URL/draft split as F3: `const [ searchTerm, setSearchTerm ] = React.useState(appliedSearchTerm)` while queries key on `appliedSearchTerm`. History navigation updates results and leaves the input behind.

Synchronize the draft when the URL term changes from outside the input, same as F3.

— Reviewed by Cursor Grok 4.6

src/shared/pagination/CONTEXT.md
```diff
@@ -0,0 +1,59 @@
...
+### What an editor of the cursor hooks must keep true
+
+- **The URL is the only list state.** Page, cursor, filters, sorting and (multichain) selected chain are
+  read from `router.query` on every render. No hook here holds React state that mirrors a URL value and no
+  effect re-syncs from the URL. A handler pushes once and returns; anything it needs afterwards is read on
+  the next render.
+- **The previous-page cursors are the one in-memory exception.** The API only hands out the *next* cursor,
+  so "Prev" needs the cursors of the pages already visited. They live in a ref keyed by page number inside
+  `usePaginationActions.ts`, filled on "Next" and seeded from the URL on mount. After a reload on page 3
+  the ref knows page 3 only, so "Prev" goes to page 1; that is expected, not a bug.
```
tom2drum: **💬 F5 · nit**

docs.md: a constraint must stay true. This says After a reload on page 3 the ref knows page 3 only, so "Prev" goes to page 1. After that reload `canGoBackwards` is `Boolean(cursorsRef.current?.[params.page - 1])` = false, and `Pagination` disables Prev (`disabled={ !canGoBackwards || … }`). The `previousPage <= 1 || !previousCursor` branch in `onPrevPageClick` is unreachable from the UI here.

Reword to: after a deep link into page 3, Prev is disabled. That matches `useApiPaginatedQuery.spec.ts`.

— Reviewed by Cursor Grok 4.6

src/shared/pagination/useApiPaginatedQuery.ts
```diff
@@ -0,0 +1,129 @@
...
+
+function hasMoreItems(nextPageParams: NextPageParams | undefined, hasNextPageFn: Params<PaginatedResourceName>['hasNextPageFn']): boolean {
+  if (!nextPageParams) {
+    return false;
+  }
+
+  return hasNextPageFn ? hasNextPageFn(nextPageParams) : Object.keys(nextPageParams).length > 0;
+}
+
+export default function useApiPaginatedQuery<Resource extends PaginatedResourceName>({
```
tom2drum, line 61: **💬 F6 · nit**

typescript.md — everything except React components is a named export. `export default function useApiPaginatedQuery` continues the old hook's default export and breaks that rule.

Named-export the hook and update imports.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
⚠️ **Blocked** — 4 major, 2 nit open

blocker 0 · major 4 · nit 2 · not anchorable 0

— Reviewed by Cursor Grok 4.6

src/slices/address/pages/details/txs/AddressTxs.tsx
tom2drum: **fix** — `AddressTxsCrossChain` now receives `isLoading={ crossChainQuery.isInitialLoading }` and `isTransitioning={ crossChainQuery.isTransitioning }`, matching `TransactionsCrossChain.tsx`. Commit 45761f92f (not pushed yet).

src/slices/address/pages/details/token-transfers/AddressTokenTransfers.tsx
tom2drum: **fix** — `TokenTransfersCrossChainContent` now receives `isLoading={ crossChainQuery.isInitialLoading }` and `isTransitioning={ crossChainQuery.isTransitioning }`, matching the other two call sites. Commit 45761f92f (not pushed yet).

src/slices/search/hooks/useSearchQuery.ts
tom2drum: **reject** — out of scope. The stale-input-on-Back/Forward behaviour is app-wide and pre-existing: every free-text filter goes through `FilterInput`, which keeps its own `initialValue` draft, so all ~12 lists using `useDebouncedFilterChange` behave the same way, on `main` too. On `main` this hook ignored history navigation for *both* the input and the results (results keyed on local debounced state); the PR fixes the results side and leaves the input unchanged. FR4/FR5 cover the list state the hook reads (page, cursor, filters, sorting), and the hook does read `q` from the URL; FR4 explicitly requires "navigating back restores the same list state as today". Fixing only the search page would diverge from `FilterInput`; syncing all inputs to the URL on history navigation belongs in a follow-up issue.

src/features/multichain/pages/search-results/useSearchQuery.ts
tom2drum: **reject** — same grounds as F3. `main` had the identical `useState(initialValue)` draft here; the PR changes only where the applied term is read from. Follow-up issue covering every free-text input, not a per-page patch.

src/shared/pagination/CONTEXT.md
tom2drum: **fix** (nit promoted) — reworded: after a deep link into page 3 the ref knows page 3 only, so "Prev" is disabled until a page has been reached via "Next". Commit 45761f92f (not pushed yet).

src/shared/pagination/useApiPaginatedQuery.ts
tom2drum: **deferred** — nit left as it stands. The rule is real, but the default export is inherited from `useQueryWithPages` and switching it is a mechanical rewrite across 107 importing files; not worth widening this PR's diff. Candidate for a separate codemod PR.

src/slices/address/pages/details/txs/AddressTxs.tsx
tom2drum: Pushed in 45761f92f.

src/slices/address/pages/details/token-transfers/AddressTokenTransfers.tsx
tom2drum: Pushed in 45761f92f.

src/shared/pagination/CONTEXT.md
tom2drum: Pushed in 45761f92f.

src/slices/address/pages/details/txs/AddressTxs.tsx
tom2drum: F1 — verified

src/slices/address/pages/details/token-transfers/AddressTokenTransfers.tsx
tom2drum: F2 — verified

src/slices/search/hooks/useSearchQuery.ts
tom2drum: F3 — accepted, the search-input draft is the same FilterInput typing buffer as on main; FR4/FR5 cover list state read from the URL, which already follows q.

src/features/multichain/pages/search-results/useSearchQuery.ts
tom2drum: F4 — accepted, same grounds as F3.

src/shared/pagination/CONTEXT.md
tom2drum: F5 — verified

src/shared/pagination/useApiPaginatedQuery.ts
tom2drum: F6 — deferred

tom2drum (commented):
✅ **Cleared** — 1 nit open

— Reviewed by Cursor Grok 4.6

PR #3699, opened 2026-09-10
Enforce the app-config convention with ESLint rules

### LINKED ISSUE #3683: Enforce the feature-config convention with lint rules
## Description

A feature's configuration module (`src/features/<name>/config.ts`) is expected to expose exactly one thing — its default-exported config object — and consumers are expected to read it through the root aggregator as `config.features.<name>`, never by importing the feature's module directly. The convention is currently followed by habit and reviewer attention rather than by any check, so nothing stops it eroding.

This issue is to write the convention down and enforce it, most likely as lint rules.

## Details

The convention has three parts:

1. **Every value belonging to a feature lives inside the config object**, widening its `Feature<Payload>` payload — no additional named exports alongside the default. This matters because the `Feature<Payload>` union hides the payload when `isEnabled` is `false`: a value exported next to the object is readable while the feature is off, which is exactly the state the type exists to make unrepresentable.
2. **Root `src/config` is the only import surface** for feature configuration. Components, hooks and utilities read `config.features.<name>`.
3. **Env values are read only in config modules.** The helpers that read raw environment values — `getEnvValue`, `parseEnvJson`, `getExternalAssetFilePath` — belong in `config.ts` files, never in a component, hook or utility. A config module is the layer that transforms and structures environment values into data the app can read; app code consuming an env directly bypasses that layer, and the value it reads is untyped, unvalidated and invisible to the aggregator.

Two exemptions the rules must allow:

- **config-to-config imports.** A feature's config may import another feature's config module directly, because the root aggregator cannot be used from inside a module it aggregates. There are ~43 such imports today (for example a feature gated on whether another feature is enabled), and all of them are legitimate.
- **The env-validator constraint** already documented for these files: they are executed by Node during startup validation, so they must not import React, browser APIs, or code from other slices and features.

Current state, for sizing: all three rules would land green rather than requiring a migration. No consumer outside a `config.ts` imports a feature's config module directly, and no file outside `src/config/` and the `config.ts` modules calls the env-reading helpers. One config module exports its object both as a named and a default export, which is a harmless duplicate rather than a payload leak.

The third rule is the one with a demonstrated failure mode: a utility that derived request query parameters was written to call `getEnvValue` directly, because the value it needed had been placed outside the config payload. It read correctly in the browser and differently under Node, which broke an unrelated test suite in a way that surfaced only as a visual diff. Both halves of that mistake are what these rules would have caught.

## Expected Behavior

- The convention is stated in the project's written rules, so both people and agents can follow it without reverse-engineering it from existing files.
- A lint rule flags a named export other than the default in `src/features/*/config.ts` (and the slice equivalent).
- A lint rule flags an import of `src/features/*/config` from anything that is not itself a `config.ts` or the root aggregator, pointing the author at `config.features.<name>`.
- A lint rule flags an import of the env-reading helpers from anything other than a `config.ts` or `src/config/**`, pointing the author at the owning config module. Server-only code that legitimately reads `process.env` directly is out of its scope.
- All three rules run in the existing lint task, so a violation fails before review.

## Additional Context

Worth deciding while implementing: whether the same rules should apply to `src/slices/*/config.ts`, and whether a violation should be an error or a warning during any transition period.

tom2drum:
## Description

Resolves #3683

A feature's `config.ts` is meant to expose one default export, be read only through `config.features.<name>` on the root aggregator, and be the only place raw env values are read. Nothing enforced that, and one util that called `getEnvValue` directly read a different value under Node than in the browser and broke an unrelated test suite. This PR writes the convention down and turns it into three ESLint errors:

- **No named exports in a config module** (`no-restricted-syntax` on `ExportNamedDeclaration` / `ExportAllDeclaration`, scoped to the config-module globs). Types included; they belong in the sibling `types/config.ts`.
- **A config module is imported only from another `config.ts`, from `src/config/**`, or from its own `config.spec.ts`** (`boundaries/dependencies`). Type-only imports pass; relative, dynamic and re-export forms are caught.
- **`src/config/utils/envs.ts` is imported only from a `config.ts` or `src/config/**`** (`boundaries/dependencies`). Scoped to `src/`; Playwright fixtures and deploy tooling keep reading envs.

Scope is every module the root aggregator assembles: `src/api/config.ts`, `src/features/**/config.ts`, `src/slices/*/config.ts`, `src/shell/*/config.ts`, `src/services/*/config.ts`. `types/config.ts` and `mocks/config.ts` companions are excluded. Three migrations were needed: `src/api/config.ts` exported `ApiPropsBase`/`ApiPropsFull`/`Apis` (moved to `src/api/types.ts`), `alternative-explorers/config.ts` exported its object twice, and `usercentrics/config.ts` exported `STORAGE_KEY` (a static constant, moved to `consts.ts`).

The convention, its reasoning and the exemptions are in the new `src/config/CONTEXT.md`, linked from `AGENTS.md` and `.agents/rules/env-vars.md`.

## Environment variables

None

## Minimum API version

None

None

## Additional information

- `eslint-plugin-boundaries` had no import resolver configured, so every import target was "unknown" and the existing `src-api → src-slices/src-features` boundary rule never fired. `eslint-import-resolver-node` is now a direct devDependency (it was already in the lockfile transitively) and configured with the repo root in `paths` so `src/…` specifiers resolve. That revived the old rule, which surfaced two pre-existing violations: `src/api/hooks/useApiQuery.ts` importing `useMultichainContext` from `src/features/multichain`, and a spec importing a features mock. Both carry an explanatory `eslint-disable` pointing at this issue rather than an architectural fix, which is out of scope here. Worth a follow-up.
- Lint time was measured before and after, three runs each on the same machine: baseline 51–58s, with the resolver and new policies 49–52s. No measurable cost.
- Rule-level `importKind` and the `boundaries/element-types` alias in the pre-existing rule are deprecated in v6 and print warnings at config load; left untouched.

https://claude.ai/code/session_01Hv8kJu53gsF1oCGfWhPT1i

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -118,9 +118,8 @@ const APP_CONFIG_ELEMENTS = [
 ];
 
 const APP_CONFIG_IMPORT_MESSAGE =
-  'Read app config through the root aggregator: `import config from \'src/config\'`, then `config.features.<name>`, `config.slices.<name>`, ' +
-  '`config.shell.<name>`, `config.services.<name>`, `config.apis`, `config.chain` or `config.metadata`. ' +
-  'Only another config.ts, src/config/** or the module\'s own config.spec.ts may import a config module directly.';
+  'Read app config through the root aggregator: `import config from \'src/config\'`. ' +
+  'Only another config.ts, src/config/** or a config.spec.ts may import a config module directly.';
 
 const APP_CONFIG_ENVS_MESSAGE =
   'Env values are read only in config modules. Move the getEnvValue / parseEnvJson / getExternalAssetFilePath call into the owning config.ts ' +
@@ -128,7 +127,7 @@ const APP_CONFIG_ENVS_MESSAGE =
 
 const APP_CONFIG_NAMED_EXPORT_MESSAGE =
   'A config module exposes only its default export. Put the value inside the config object (widen the Feature payload) ' +
-  'so it is unreachable while the feature is disabled; types and constants go to a sibling types/config.ts.';
+  'so it is unreachable while the feature is disabled; types and constants go to a sibling role file (types/config.ts, types.ts, consts.ts).';
 
 const RESTRICTED_SYNTAX = [
   {
@@ -423,9 +422,10 @@ export default tseslint.config(
         'default': 'allow',
         rules: [
           {
-            from: 'src-api',
-            disallow: [ 'src-slices', 'src-features' ],
-            importKind: 'value',
+            // by path, not by element type: `src/api/config.ts` is an `app-config-module` element (a file
+            // is one element only), and the layer policy still applies to it
+            from: { path: 'src/api/**' },
+            disallow: { to: { type: [ 'src-slices', 'src-features' ] }, dependency: { kind: 'value' } },
           },
         ],
       } ],
@@ -623,12 +623,20 @@ export default tseslint.config(
       // restricted imports, properties and syntax
       'no-restricted-syntax': [ 'error', ...RESTRICTED_SYNTAX ],
       'no-restricted-imports': [ 'error', RESTRICTED_MODULES ],
-      'no-restricted-properties': [ 2, {
-        object: 'process',
-        property: 'env',
-        // FIXME: restrict the rule only NEXT_PUBLIC variables
-        message: 'Please use src/config/index.ts to import any NEXT_PUBLIC environment variables. For other properties please disable this rule for a while.',
-      } ],
+      'no-restricted-properties': [ 2,
+        {
+          object: 'process',
+          property: 'env',
+          // FIXME: restrict the rule only NEXT_PUBLIC variables
+          message: 'Please use src/config/index.ts to import any NEXT_PUBLIC environment variables. For other properties please disable this rule for a while.',
+        },
+        {
+          // app config convention (src/config/CONTEXT.md): the raw runtime env map is read by getEnvValue only
+          object: 'window',
+          property: '__envs',
+          message: APP_CONFIG_ENVS_MESSAGE,
+        },
+      ],
     },
   },
   {
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -97,6 +97,50 @@ const ARCH_BOUNDARY_ELEMENTS = [
   { type: 'src-features', pattern: 'src/features/**', mode: 'full' },
 ];
 
+/*
+ * App config convention (src/config/CONTEXT.md) — the modules the root aggregator assembles.
+ * `types/config.ts` and `mocks/config.ts` are companions, not config modules, and stay out.
+ */
+const APP_CONFIG_MODULE_GLOBS = [
+  'src/api/config.ts',
+  'src/features/**/config.ts',
+  'src/services/*/config.ts',
+  'src/shell/*/config.ts',
+  'src/slices/*/config.ts',
+];
+const APP_CONFIG_COMPANION_GLOBS = [ '**/types/config.ts', '**/mocks/config.ts' ];
+
+const APP_CONFIG_ELEMENTS = [
+  { type: 'app-config-envs', pattern: 'src/config/utils/envs.ts', mode: 'full' },
+  { type: 'app-config-root', pattern: 'src/config/**', mode: 'full' },
+  { type: 'app-config-module', pattern: APP_CONFIG_MODULE_GLOBS.map((glob) => glob.replace('**/config.ts', '**/!(types|mocks)/config.ts')), mode: 'full' },
+  { type: 'app-config-spec', pattern: 'src/**/config.spec.ts', mode: 'full' },
+];
+
+const APP_CONFIG_IMPORT_MESSAGE =
+  'Read app config through the root aggregator: `import config from \'src/config\'`, then `config.features.<name>`, `config.slices.<name>`, ' +
+  '`config.shell.<name>`, `config.services.<name>`, `config.apis`, `config.chain` or `config.metadata`. ' +
+  'Only another config.ts, src/config/** or the module\'s own config.spec.ts may import a config module directly.';
+
+const APP_CONFIG_ENVS_MESSAGE =
+  'Env values are read only in config modules. Move the getEnvValue / parseEnvJson / getExternalAssetFilePath call into the owning config.ts ' +
+  'and read the structured result via src/config.';
+
+const APP_CONFIG_NAMED_EXPORT_MESSAGE =
+  'A config module exposes only its default export. Put the value inside the config object (widen the Feature payload) ' +
+  'so it is unreachable while the feature is disabled; types and constants go to a sibling types/config.ts.';
+
+const RESTRICTED_SYNTAX = [
+  {
+    selector: 'CallExpression[callee.property.name=\'localeCompare\']',
+    message: 'Use the shared collator from src/shared/texts/collator.ts (collator.compare) instead of String.prototype.localeCompare.',
+  },
+  {
+    selector: 'NewExpression[callee.object.name=\'Intl\'][callee.property.name=\'Collator\']',
+    message: 'Use the shared collator from src/shared/texts/collator.ts instead of constructing Intl.Collator inline.',
+  },
+];
+
 /** @type {import('eslint').Linter.Config[]} */
 export default tseslint.config(
   includeIgnoreFile(gitignorePath),
@@ -118,7 +162,18 @@ export default tseslint.config(
   {
     settings: {
       react: { version: 'detect' },
-      'boundaries/elements': ARCH_BOUNDARY_ELEMENTS,
+      // first matching descriptor wins, so the file-level config elements go before the layer-level ones;
+      // the trailing catch-all makes every src file a known element — policies never run from unknown files
+      'boundaries/elements': [ ...APP_CONFIG_ELEMENTS, ...ARCH_BOUNDARY_ELEMENTS, { type: 'src-other', pattern: 'src/**', mode: 'full' } ],
+      'boundaries/dependency-nodes': [ 'import', 'dynamic-import', 'export' ],
+      // without a resolver every import target is "unknown" and no boundaries policy ever matches;
+      // `paths` makes the tsconfig-style root-relative specifiers (src/…, playwright/…) resolvable
+      'import/resolver': {
+        node: {
+          extensions: [ '.ts', '.tsx', '.js', '.mjs' ],
+          paths: [ __dirname ],
+        },
+      },
     },
   },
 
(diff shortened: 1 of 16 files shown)


eslint.config.mjs
```diff
@@ -118,7 +162,18 @@ export default tseslint.config(
   {
     settings: {
       react: { version: 'detect' },
-      'boundaries/elements': ARCH_BOUNDARY_ELEMENTS,
+      // first matching descriptor wins, so the file-level config elements go before the layer-level ones;
+      // the trailing catch-all makes every src file a known element — policies never run from unknown files
+      'boundaries/elements': [ ...APP_CONFIG_ELEMENTS, ...ARCH_BOUNDARY_ELEMENTS, { type: 'src-other', pattern: 'src/**', mode: 'full' } ],
```
tom2drum, line 173: **⚠️ F1 · major · 🙋 needs human**

`'boundaries/elements': [ ...APP_CONFIG_ELEMENTS, ...ARCH_BOUNDARY_ELEMENTS, ... ]` — first descriptor wins, so `src/api/config.ts` is now `app-config-module`, not `src-api`, and drops out of the existing `boundaries/element-types` policy (`from: 'src-api', disallow: ['src-slices', 'src-features']`, value imports).

The value import `STATS_API_RESOURCES_REFETCH_INTERVAL` from `src/features/chain-stats/types/config` in `src/api/config.ts` is therefore unenforced. The same layer crossing in `useApiQuery.ts` needed an explicit disable. Reversing the two element arrays makes that import error; keeping this order also depends on `app-config-module` winning so `src/api/config.ts` can still import `envs.ts`.

Add `app-config-module` to the arch rule's `from` (or an equivalent disable at the import) so layer policy still applies to config files that sit under `src/api/**` / `src/features/**`.

— Reviewed by Cursor Grok 4.6

src/config/CONTEXT.md
```diff
@@ -0,0 +1,38 @@
...
+
+## The convention
+
+Three rules. ESLint enforces all of them (`eslint.config.mjs`, the "App config convention" blocks), so a
+violation fails before review.
+
+1. **A config module has no export besides its default.** Everything the feature owns lives inside the
+   config object, widening its `Feature<Payload>`. `Feature<>` hides the payload when `isEnabled` is
+   `false`. A value exported beside the object stays readable while the feature is off, which is the state
+   the type exists to rule out. Types and constant lists go to the sibling `types/config.ts`.
```
tom2drum: **⚠️ F2 · major**

`Types and constant lists go to the sibling `types/config.ts`.` This same diff contradicts that: `STORAGE_KEY` moved to `src/services/usercentrics/consts.ts`, and the api prop types moved to `src/api/types.ts`. Neither is a `types/config.ts`. `APP_CONFIG_NAMED_EXPORT_MESSAGE` in `eslint.config.mjs` repeats the same absolute, so a lint hit sends the reader to the wrong place. `consts.ts` already has precedent (`src/services/growthbook/consts.ts`).

State the real rule — a moved value goes to a sibling role file (`types/config.ts`, `types.ts`, `consts.ts`), just not the config module — and mirror that in `APP_CONFIG_NAMED_EXPORT_MESSAGE`.

— Reviewed by Cursor Grok 4.6

src/config/CONTEXT.md
```diff
@@ -0,0 +1,38 @@
...
+2. **Only `src/config` is imported.** App code reads `config.features.<name>`, `config.slices.<name>`,
```
tom2drum: **💬 F3 · nit**

Rule 2 lists `config.features`, `config.slices`, `config.shell`, `config.services`, `config.apis`, `config.chain`, `config.metadata` and drops `app` and `misc`, which `src/config/index.ts` does export. `.agents/rules/env-vars.md` already has the full section list (`app`, `apis`, `chain`, `shell`, `slices`, `features`, `services`, `metadata`, `misc`); `APP_CONFIG_IMPORT_MESSAGE` repeats the shortened copy. `.agents/rules/docs.md`: "Anything another doc states. Link instead, so the fact has one home."

Point at `env-vars.md` for the section list (or include `config.app` / `config.misc`); shorten the lint message to `import config from 'src/config'`.

— Reviewed by Cursor Grok 4.6

src/config/CONTEXT.md
```diff
@@ -0,0 +1,38 @@
...
+   aggregator. It can also differ between the browser and Node. A util that called `getEnvValue` directly
+   read one value in the browser and another under Node, and broke an unrelated test suite in a way that
+   showed up only as a visual diff.
+
+## Exemptions the rules allow
+
+- **Config-to-config imports.** A config module may import another config module directly, for example a
+  feature gated on whether another one is enabled. The aggregator cannot be used from inside a module it
+  aggregates.
+- **The module's own `config.spec.ts`** imports the module directly and dynamically, under `withEnvs`,
```
tom2drum: **💬 F4 · nit**

`**The module's own `config.spec.ts`**` is narrower than what ESLint allows. `app-config-spec` is `'src/**/config.spec.ts'` and the allow rule is `from: { type: [..., 'app-config-spec'] }, allow: { to: { type: 'app-config-module' } }` — any `config.spec.ts` may import any config module. `APP_CONFIG_IMPORT_MESSAGE` repeats "the module's own".

Write "a `config.spec.ts`" in CONTEXT.md and the lint message, or constrain the exemption by captured element name.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
🛑 **Blocked** — 2 majors open

blocker 0 · major 2 · nit 4 · not anchorable 0

— Reviewed by Cursor Grok 4.6

src/api/utils/scope-filters.spec.ts
```diff
@@ -1,6 +1,7 @@
 import type { ApiResource } from '../resources/types';
 import type { ExternalChainExtended } from 'src/shared/external-chains/types';
 
+// eslint-disable-next-line boundaries/element-types -- test fixture
```
tom2drum: **💬 F5 · nit**

`// eslint-disable-next-line boundaries/element-types -- test fixture` names what the import is, not why the boundary breach is acceptable. `.agents/rules/code-quality.md` (eslint-disable comments): every disable must include an explanation. Contrast `useApiQuery.ts`, which gives a real reason.

Say why, e.g. `-- spec needs a real multichain fixture; the api→features ban targets runtime code`.

— Reviewed by Cursor Grok 4.6

src/config/CONTEXT.md
```diff
@@ -0,0 +1,38 @@
+# App config — context
+
+`src/config/index.ts` assembles the app config from the co-located config modules: `src/api/config.ts`,
+`src/features/**/config.ts`, `src/slices/*/config.ts`, `src/shell/*/config.ts`, `src/services/*/config.ts`.
+Each module turns raw env values into typed, validated data. The aggregator is the only place the app reads
+that data from.
+
+## The convention
+
+Three rules. ESLint enforces all of them (`eslint.config.mjs`, the "App config convention" blocks), so a
```
tom2drum, line 10: **💬 F6 · nit**

`Three rules. ESLint enforces all of them … so a violation fails before review` overstates rule 3. The envs policy only disallows imports of `src/config/utils/envs.ts`. Reading the env map directly is invisible: `src/services/google-analytics/GoogleAnalytics.tsx` does `window.__envs.NEXT_PUBLIC_GOOGLE_ANALYTICS_PROPERTY_ID` from a non-config module and lints clean.

Soften the claim to name what is unenforced, or add a restriction for `window.__envs` outside config modules.

— Reviewed by Cursor Grok 4.6

eslint.config.mjs
tom2drum: fix — b4b3f0704. The arch rule now selects its source by path (`from: { path: 'src/api/**' }`, v6 object selector) instead of by element type, so `src/api/config.ts` stays under the api → features/slices ban while remaining an `app-config-module` element. Verified: a probe value import into that file errors; before the change it did not. The one existing crossing (the chain-stats refetch-interval list) carries an explicit disable with its reason.

src/config/CONTEXT.md
tom2drum: fix — b4b3f0704. CONTEXT.md rule 1 and `APP_CONFIG_NAMED_EXPORT_MESSAGE` now say a moved value goes to a sibling role file (`types/config.ts`, `types.ts`, `consts.ts`), never the config module.

src/config/CONTEXT.md
tom2drum: fix (nit promoted) — b4b3f0704. Rule 2 reads `config.<section>.<name>` and points at `.agents/rules/env-vars.md` for the section list; `APP_CONFIG_IMPORT_MESSAGE` shortened to `import config from 'src/config'`.

src/config/CONTEXT.md
tom2drum: fix (nit promoted) — b4b3f0704. Doc and lint message now say "a `config.spec.ts`", and the exemption bullet states it is by file name, not by neighbourhood. Kept the glob as is; constraining by captured element name is not worth the config.

src/api/utils/scope-filters.spec.ts
tom2drum: fix (nit promoted) — b4b3f0704. Disable reason now: `the spec needs a real multichain fixture; the api → features ban targets runtime code`.

src/config/CONTEXT.md
tom2drum: fix (nit promoted) — b4b3f0704. Added a `no-restricted-properties` entry for `window.__envs` (same message as the envs import rule), so a direct read in TS code outside `getEnvValue` now errors; verified with a probe file. `envs.ts` and `vitest/utils/mockEnvs.ts` already disable that rule on their reads. `GoogleAnalytics.tsx` is unchanged and lints clean by design: its read is inline script text inside a template string, not a member expression, and that inline script is a genuine exception. CONTEXT.md rule 3 now names the `window.__envs` restriction.

eslint.config.mjs
tom2drum: F1 — verified

src/config/CONTEXT.md
tom2drum: F2 — verified

src/config/CONTEXT.md
tom2drum: F3 — verified

src/config/CONTEXT.md
tom2drum: F4 — verified

src/api/utils/scope-filters.spec.ts
tom2drum: F5 — verified

src/config/CONTEXT.md
tom2drum: F6 — verified

eslint.config.mjs
```diff
@@ -566,6 +656,18 @@ export default tseslint.config(
       'no-restricted-properties': 'off',
```
tom2drum: **💬 F7 · nit**

The F6 `window.__envs` restriction is a second entry on `no-restricted-properties`, which this later block still sets `'off'` for `src/pages/**`, `src/server/**`, `playwright/**`, deploy scripts, `proxy.ts`, `instrumentation*.ts`, and `*.config.ts`. A `window.__envs` member expression in those trees is therefore unenforced, while the same probe under app `src/` errors. `playwright/` is already documented as unchecked; `src/pages/**` and `src/server/**` are not.

Keep `process.env` allowed in those globs; leave `window.__envs` on (split the rule, or override only the `process.env` entry).

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
✅ **Cleared** — 1 nit open

— Reviewed by Cursor Grok 4.6

eslint.config.mjs
tom2drum: fix (nit promoted) — 47d59bcb7. The `window.__envs` entry is extracted to `WINDOW_ENVS_RESTRICTION`, and the override block for pages, server, playwright and tooling files now sets `no-restricted-properties` to that entry alone instead of `'off'`, so only `process.env` is relaxed there. Verified with a probe in `src/server/`: the `window.__envs` read errors, the `process.env` read passes. Full-repo lint is 0 errors; no existing file in those globs reads the map as a member expression.

PR #3703, opened 2026-09-14
Take cross-chain list totals and the 24h widget from the stats service instead of the interchain indexer

### LINKED ISSUE #3682: Take cross-chain list totals from the stats service instead of the interchain indexer
## Description

The cross-chain transactions and cross-chain token transfers list pages take their
"A total of N found" counts from the interchain indexer's `/api/v1/stats/common`, and the
24h widget on the transactions page takes its value from `/api/v1/stats/daily`. Those
endpoints do direct DB counts, are expensive enough to affect the whole indexer service
and are scheduled for removal from its API, so the counts have to come from the stats
service instead.

## Details

- The stats service now runs in interchain mode on every deployment that shows cross-chain
  data, single-chain and multichain alike, and exposes the counters through
  `/api/v1/counters` (stats `v2.19.1`):
  - `totalInterchainMessages`, `totalInterchainTransfers` — the list totals;
  - `newMessagesInterchain24h`, `newTransfersInterchain24h` — the 24h widget. These use a
    sliding 24-hour window, whereas the indexer counted the calendar day, so the values
    legitimately differ from what `/stats/daily` returned.
- Single-chain deployments read them via the existing `stats:counters` resource
  (`NEXT_PUBLIC_STATS_API_HOST`).
- Multichain deployments read them from the host already configured as
  `NEXT_PUBLIC_MULTICHAIN_STATS_API_HOST` — it serves the same `/api/v1/counters` with the
  interchain counters alongside the interop ones. A new `multichainStats:counters` resource
  is needed; no new environment variable.
- Once nothing consumes them, the `interchainIndexer:stats_common` and
  `interchainIndexer:stats_daily` resource declarations, their payload type mappings, stubs
  and test mocks should be removed.

## Expected Behavior

- Both cross-chain list pages show the total from the stats service, on single-chain and
  multichain deployments.
- The 24h "Cross-chain txns" widget on the cross-chain transactions page shows
  `newMessagesInterchain24h` from the stats service.
- No frontend code calls the interchain indexer's `/api/v1/stats/common` or
  `/api/v1/stats/daily`.
- Where no counter is available (no stats API configured, or the counter is absent from
  the response), the list renders without the total and the widget is not shown, rather
  than with a placeholder value.

## Additional Context

- The ICTT users page keeps reading `/api/v1/stats/chains` from the interchain indexer —
  agreed with the backend owner; it is not part of this change.
- Stats counters are per-deployment: the stats instance serves a fixed slice and does not
  filter per request, so the slice must match the bridges the frontend requests via
  `NEXT_PUBLIC_CROSS_CHAIN_TXS_BRIDGE_IDS`. That is a stats-service deployment concern, not
  a frontend one.

tom2drum:
## Description

Resolves #3682

The cross-chain transactions and token transfers pages took their "A total of N found" counts from the interchain indexer's `/api/v1/stats/common`, and the 24h widget from `/api/v1/stats/daily`. Both do direct DB counts, are expensive enough to affect the whole indexer and are scheduled for removal from its API. The stats service now runs in interchain mode on every deployment showing cross-chain data and exposes the same numbers through `/api/v1/counters`, so both pages read them from there.

- New `useCrossChainCountersQuery` hook picks the stats resource by deployment: `stats:counters` on single-chain, the new `multichainStats:counters` on multichain (the host already configured as `NEXT_PUBLIC_MULTICHAIN_STATS_API_HOST` serves the interchain counters alongside the interop ones, so no new env var). Counters are picked by id: `totalInterchainMessages`, `totalInterchainTransfers`, `newMessagesInterchain24h`.
- When no stats API is configured or the counter is absent, the list renders without the total and the 24h widget is hidden; `StickyPaginationWithText` now renders nothing for an empty text instead of an empty spaced box.
- `interchainIndexer:stats_common` and `interchainIndexer:stats_daily` resources, their payload types, stubs and test mocks are removed. The ICTT users page keeps reading `/api/v1/stats/chains` from the indexer, as agreed with the backend owner.
- `@blockscout/stats-types` bumped 2.11.1 → 2.19.1 to match the stats release.

## Environment variables

None.

## Minimum API version

Stats microservice v2.19.1+ (interchain 24h counters; totals from v2.18.0).

None. Note: the 24h values now use a sliding 24-hour window instead of the indexer's calendar day, so they differ from before by design.

## Additional information

The stats instance serves a fixed slice and does not filter per request, so its slice must match the bridges the frontend requests via `NEXT_PUBLIC_CROSS_CHAIN_TXS_BRIDGE_IDS` — a stats deployment concern, not detectable by the frontend.

Verified: `lint:tsc`, eslint, cspell, complexity gate, new Vitest spec (3 cases). Playwright suites for the two pages updated to mock `stats:counters` with the same values as before; not run locally (no docker).

https://claude.ai/code/session_01UMn2uArTibjzH8BpyuRthM


tom2drum (commented):

blocker 0 · major 0 · nit 2 · not anchorable 0

— Reviewed by Cursor Grok 4.6

src/features/cross-chain-txs/hooks/useCrossChainCountersQuery.spec.tsx
```diff
@@ -0,0 +1,69 @@
...
+      const { result } = renderHook(() => useCrossChainCountersQuery(), { wrapper });
+
+      await waitFor(() => expect(result.current.isPlaceholderData).toBe(false));
+
+      expect(fetchMock.mock.calls[0][0]).toBe(`${ MULTICHAIN_STATS_API_HOST }/api/v1/counters`);
+      expect(result.current.data?.totalInterchainMessages).toBe('100');
+    });
+  });
+
+  it('returns no data and no placeholder when no stats service is configured', async() => {
```
tom2drum, line 99: **💬 F1 · nit**

The “counter absent” branch is untested. Consumers now hide the total and the 24h widget when the id is missing (`total !== undefined`, `value === undefined`), but this spec only covers a payload with all four ids and the disabled-stats case.

Add a case whose stats response has no interchain ids and assert those keys are undefined on `data`.

— Reviewed by Cursor Grok 4.6

src/features/cross-chain-txs/hooks/useCrossChainCountersQuery.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export type CrossChainCounters = Partial<Record<CrossChainCounterId, string>>;
+
+type ResourceName = 'stats:counters' | 'multichainStats:counters';
+
+const RESOURCE_NAME: ResourceName = config.features.multichain.isEnabled ? 'multichainStats:counters' : 'stats:counters';
+const IS_ENABLED = config.features.multichain.isEnabled ? Boolean(config.apis.multichainStats) : Boolean(config.apis.stats);
+
+function selectCrossChainCounters(data: stats.Counters): CrossChainCounters {
+  return Object.fromEntries(data.counters.map((counter) => [ counter.id, counter.value ]));
```
tom2drum, line 25: **💬 F2 · nit**

`return Object.fromEntries(data.counters.map((counter) => [ counter.id, counter.value ]))` copies every stats counter. `Object.fromEntries` is `{ [k: string]: string }`, which still satisfies `Partial<Record<CrossChainCounterId, string>>`, so a renamed backend id type-checks and shows up as `undefined` at the call sites. The stub `COUNTER_IDS` list restates the same union with no `satisfies`.

Filter to the known ids when building the record, and type the stub list as `satisfies Array<CrossChainCounterId>`.

— Reviewed by Cursor Grok 4.6

src/features/cross-chain-txs/hooks/useCrossChainCountersQuery.spec.tsx
tom2drum: fix — added a spec case whose stats payload carries only unrelated counter ids (`totalBlocks`, `averageBlockTime`) and asserts all four interchain keys are `undefined` on `data`. Commit 8b41fee36.

src/features/cross-chain-txs/hooks/useCrossChainCountersQuery.ts
tom2drum: Part fix, part reject.

- **Filter to known ids** — reject. Filtering changes no observable behavior: a renamed backend id yields `undefined` at the call sites with or without the filter, and no compile-time check can catch a server-side rename. The extra keys in the record are harmless. The new missing-counter spec (F1) covers the `undefined` path explicitly.
- **`satisfies Array<CrossChainCounterId>` on the stub list** — fix. Applied in commit 8b41fee36.

src/features/cross-chain-txs/hooks/useCrossChainCountersQuery.spec.tsx
tom2drum: F1 — verified

src/features/cross-chain-txs/hooks/useCrossChainCountersQuery.ts
tom2drum: F2 — verified on satisfies; accepted, extra keys do not change the hide-when-missing path and a filter cannot catch a server-side id rename.

tom2drum (commented):

— Reviewed by Cursor Grok 4.6

PR #3710, opened 2026-09-16
Migrate Swap to the official LI.FI widget

### LINKED ISSUE #3317: Replace our custom iframe solution for Swap with a new Li.Fi widget
Docs: https://github.com/lifinance/widget/blob/main/packages/widget-embedded/README.md 
Integration example (EVM): https://github.com/lifinance/widget/tree/main/examples/vite-iframe-wagmi

maxaleks:
Resolves https://github.com/blockscout/frontend/issues/3317


patches/@lifi__widget-light@4.3.2.patch
```diff
@@ -0,0 +1,187 @@
...
+@@ -66,7 +73,11 @@ function useEthereumIframeHandler() {
+ 		};
+ 	}, []);
+ 	const handleRequest = useCallback(async (_id, method, params) => {
+-		const { address, chainId, walletClient, publicClient } = stateRef.current;
++		const { address, chainId, connector } = getAccount(config);
++		const walletClient = address && WALLET_METHODS.has(method)
++			? await getWalletClient(config, { account: address, chainId, connector })
++			: undefined;
++		const publicClient = getPublicClient(config, { chainId });
```
tom2drum, line 64: **⚠️ F1 · major**

The patched handler resolves the public client from the wallet's live chain on every request:

```js
const { address, chainId, connector } = getAccount(config);
...
const publicClient = getPublicClient(config, { chainId });
```

`getAccount().chainId` is the connector's chain, not a configured one (`wagmi-config.ts` sets `syncConnectedChain: false`). `@wagmi/core` `getClient` throws `ChainNotConfiguredError` when `chainId` is given and not in `config.chains`. So a user whose wallet sits on a chain outside the app's list gets every widget RPC rejected — `eth_chainId`, `eth_accounts`, and crucially `wallet_switchEthereumChain`, the only way out. The unpatched hook used `usePublicClient()` (store chain) and never threw here.

Fix: wrap in try/catch and fall back to `getPublicClient(config)` (store chain), or only build the public client for non-wallet methods. Add a spec with the mock connector on an unconfigured chain issuing `wallet_switchEthereumChain`.

— Reviewed by Claude Fable 5.1

tom2drum (commented):
🛑 **Blocked** — 2 major open · 🙋 3 needs human

blocker 0 · major 2 · nit 8 · not anchorable 1

Axes: spec (skipped — no spec for this change) · standards · correctness

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/SwapWidgetIframe.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <Box position="relative" minH="500px" bg="bg.primary" aria-busy={ !isReady }>
+      <StyledLiFiWidget
+        { ...props }
+        autoResize
+        w="full"
+        minH="500px"
+        display="block"
+        visibility={ isReady ? 'visible' : 'hidden' }
```
tom2drum: **⚠️ F2 · major · 🙋 needs human**

The iframe stays hidden and the spinner stays up until the guest emits `PageEntered`:

```tsx
visibility={ isReady ? 'visible' : 'hidden' }
...
{ !isReady && <Center position="absolute" inset={ 0 }><ContentLoader/></Center> }
```

There is no error, timeout or blocked-frame path. If widget.li.fi is unreachable, blocked by an extension or proxy, or the hosted widget never reaches its first page, the user sees an indefinite loader with no message and no retry. The old `MarketplaceAppIframe` at least revealed the frame on `onLoad`. The spec only covers the happy READY → PageEntered sequence.

Fix: add a bounded fallback (reveal the iframe or show an error after N seconds without `PageEntered`, or on iframe `error`), plus a spec for it. Human call on what the fallback should look like.

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/Swap.tsx
```diff
@@ -1,64 +1,78 @@
...
 
   return (
     <Flex flex="1" flexDir="column" justifyContent="space-between" gap={ 6 }>
-      <MarketplaceAppIframe
-        appId="swap"
-        appUrl={ dappConfig?.url }
-        message={ message }
-        isEssentialDapp
-      />
+      <Web3Boundary fallback={ <Center h="500px"><ContentLoader/></Center> }>
```
tom2drum: **💬 F4 · nit**

`code-quality.md` § Magic numbers: a value that must stay in sync at more than one call site gets one constant. `<Center h="500px"><ContentLoader/></Center>` repeats the widget's reserved height that `SwapWidgetIframe.tsx` hard-codes twice as `minH="500px"` — three sites that must agree so the `Web3Boundary` fallback and the iframe placeholder don't jump. The fallback also re-implements the centered `ContentLoader` that `SwapWidgetIframe` already renders while `!isReady`.

Fix: one `SWAP_WIDGET_MIN_HEIGHT` constant (or an exported placeholder component) in `SwapWidgetIframe.tsx`, used by the boundary fallback.

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.ts
```diff
@@ -0,0 +1,90 @@
...
+    pendingBatches.current.delete(batchId);
+    await confirm(hash);
+  }, []);
+
+  const handleRequest = useCallback<IframeEcosystemHandler['handleRequest']>(async(id, method, params) => {
+    if (method === 'wallet_sendCalls') {
+      return sendCalls(id, method, params);
+    }
+    if (method === 'eth_sendTransaction') {
+      const transaction: unknown = Array.isArray(params) ? params[0] : undefined;
```
tom2drum: **💬 F5 · nit**

Two parsing idioms for the same job in one file. The batch paths validate iframe payloads with valibot schemas, but `eth_sendTransaction` and `wallet_getCallsStatus` hand-roll the narrowing:

```ts
const transaction: unknown = Array.isArray(params) ? params[0] : undefined;
const to = transaction && typeof transaction === 'object' && 'to' in transaction && typeof transaction.to === 'string' ?
  transaction.to : '';
```

Fix: `v.safeParse(v.tuple([ v.object({ to: v.optional(v.string()) }) ]), params)` here and `v.tuple([ v.string() ])` for the batch id, then read `.output`.

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
```diff
@@ -0,0 +1,368 @@
...
+import { http, UserRejectedRequestError, type EIP1193Parameters, type EIP1193Provider, type WalletRpcSchema } from 'viem';
+import { createConfig, mock, WagmiProvider, type Config } from 'wagmi';
+import { connect, disconnect } from 'wagmi/actions';
+import { mainnet, optimism } from 'wagmi/chains';
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { act, cleanup, render, renderHook, waitFor } from 'vitest/lib';
+import withEnvs from 'vitest/utils/mockEnvs';
+
+import { useSwapWallet } from './useSwapWallet';
```
tom2drum: **💬 F6 · nit**

`tests-unit.md` § Overriding environment variables: with `withEnvs`, everything that reads config must be imported dynamically inside the callback, not statically at the top of the spec. This spec does both — `import { useSwapWallet } from './useSwapWallet';` here and `const { useSwapWallet } = await import('./useSwapWallet');` inside `renderTrackedSwap`, the latter shadowing the former. The module-level `Widget` component binds the static, pre-reset instance; any future `withEnvs` test that renders it silently gets stale config.

Fix: import once, dynamically, in a helper, or split the env-override describe into its own spec file.

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
```diff
@@ -0,0 +1,368 @@
...
+const SIGNATURE = `0x${ 'cd'.repeat(65) }`;
+const ETHEREUM_RPC = 'https://ethereum-rpc.test';
+const OPTIMISM_RPC = 'https://optimism-rpc.test';
+const WIDGET_ORIGIN = 'https://widget.li.fi';
+const WIDGET_CONFIG = { integrator: 'blockscout' };
+const BATCH_ID = 'wallet-batch-id';
+const APPROVAL_HASH = `0x${ 'ef'.repeat(32) }`;
+const REWARDS_ORIGIN = 'https://rewards.test';
+const ACTIVITY_TOKEN = 'batch-activity-token';
+const ENV_OVERRIDES: Array<[ string, string ]> = [
```
tom2drum, line 31: **💬 F7 · nit**

Duplicated code across the two new specs: `ENV_OVERRIDES`, the `ACCOUNT`/`TX_HASH`/`REWARDS_ORIGIN` constants, the `QueryClientProvider > WagmiProvider > RewardsContextProvider` wrapper, the mixpanel `vi.mock` + `init('test-mixpanel', …)`, the `setQueryData` seeding for `core:user_info` / `rewards:user_check_activity_pass`, the rewards fetch mock and the `onLoginSuccess` + `waitFor(isAuth)` handshake are copied verbatim between `renderTrackedSwap` and `useMarketplaceWalletActivity.spec.tsx`.

Fix: extract a `renderWithRewards` / `mockRewardsActivity` helper under `vitest/utils/` (next to `mockSocket.ts`) and call it from both specs.

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
```diff
@@ -0,0 +1,368 @@
...
+  it.each([
+    { method: 'eth_sendTransaction', params: [ { to: OTHER_ACCOUNT, value: '0x0' } ], expected: TX_HASH },
+    { method: 'personal_sign', params: [ '0x1234', ACCOUNT ], expected: SIGNATURE },
+  ])('waits for the new chain client before $method', async({ method, params, expected }) => {
+    const { result } = renderHook(useSwapWallet, { wrapper: WalletProvider });
+    await waitFor(() => expect(queryClient.isFetching()).toBe(0));
+    const connector = walletConfig.connectors[0];
+    const getAccounts = connector.getAccounts.bind(connector);
+    const clientReady = Promise.withResolvers<void>();
+    vi.spyOn(connector, 'getAccounts').mockImplementation(async() => {
```
tom2drum: **💬 F8 · nit**

The "waits for the new chain client" test gates on a wagmi internal rather than observable behaviour:

```ts
vi.spyOn(connector, 'getAccounts').mockImplementation(async() => {
  if (await connector.getChainId() === optimism.id) {
    await clientReady.promise;
  }
  ...
expect(walletRpcUrls).toEqual([]);
```

It only holds because the patch passes `connector` to `getWalletClient`, which makes `getConnectorClient` call `connector.getAccounts()` — a `@wagmi/core` implementation detail. A refactor to `getWalletClient(config, { chainId })` would keep the user-visible behaviour (tx goes to the Optimism RPC, asserted a few lines down) yet break the pre-resolve assertion. Per `tests-unit.md` "What a good test is", that assertion tests the framework's call order, not the hook.

Fix: drop the `getAccounts` spy and the pre-resolve `walletRpcUrls` assertion; keep the post-switch RPC-URL assertion, which is the real contract.

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/SwapWidgetIframe.tsx
```diff
@@ -0,0 +1,44 @@
...
+        css={{
+          colorScheme: colorMode,
```
tom2drum: **💬 F9 · nit · 🙋 needs human**

Dark mode for the widget relies on `appearance: 'system'` (Swap.tsx) plus the host element's CSS:

```tsx
css={{
  colorScheme: colorMode,
```

That depends on browsers propagating the `<iframe>`'s used `color-scheme` into the embedded document's `prefers-color-scheme`. Chrome and Firefox do; Safari support is version-dependent. Where it does not propagate, the widget follows the OS theme instead of the explorer's, with no fallback since `hiddenUI.appearance` hides the widget's own toggle. Untestable in jsdom.

Fix: verify in Safari; if it does not follow, pass `appearance: colorMode` explicitly instead of `'system'` (the guest handles `CONFIG_UPDATE` in place, so no remount).

— Reviewed by Claude Fable 5.1

docs/ENVS.md
```diff
@@ -1063,7 +1063,7 @@ Essential dapps are built-in dapps that are displayed on the Marketplace page in
 
 | Property | Type | Description | Compulsoriness | Example value |
 | --- | --- | --- | --- | --- |
-| swap | `{ url: string, chains: Array<string>, fee: string, integrator: string }` | Swap config | - | `{'url': 'https://example.com', 'chains': ['1'], 'fee': '0.004', 'integrator': 'blockscout'}` |
+| swap | `{ chains: Array<string>, fee: string, integrator: string }` | Swap config for the LI.FI-hosted widget. The legacy `url` property is ignored. | - | `{'chains': ['1'], 'fee': '0.004', 'integrator': 'blockscout'}` |
```
tom2drum: **💬 F10 · nit · 🙋 needs human**

`docs.md` § What the code and the environment already say: leave out what the change replaced — the commit and the PR carry that. The new description `Swap config for the LI.FI-hosted widget. The legacy \`url\` property is ignored.` narrates the migration, and `add-env-var` step 1 asks for one sentence with no implementation details. The nested yup `.shape()` has no `noUnknown`, so a leftover `url` really is stripped, but that fact belongs in the `ENVs`-labelled PR body that `prepare-release` reads for DevOps. Human call whether operators need the hint in the reference table for the release.

Fix: `Swap config` in the table, and name the dropped `url` field in the PR description under the ENVs label.

— Reviewed by Claude Fable 5.1

tom2drum:
### 📎 Findings without a diff anchor

**💬 F3 · nit**

_src/features/marketplace/hooks/useMarketplaceWallet.ts:70_

Dead scaffolding: the hook still returns `chainId,` but this diff removed its only consumer — `MarketplaceAppIframe.tsx` now destructures `address, sendTransaction, signMessage, signTypedData, switchChain` and no longer reads `chainId`. Nothing else calls `useMarketplaceWallet`.

Fix: drop `chainId` from the returned object.

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
```diff
@@ -0,0 +1,368 @@
...
+        return state.status;
+      }
+      return provider.request(args);
+    };
+    return { ...provider, request: request as typeof provider.request };
+  });
+
+  const activityRequests: Array<{ path: string; body: unknown }> = [];
+  fetchMock.mockResponse(async(request) => {
+    if (!request.url.startsWith(REWARDS_ORIGIN)) {
```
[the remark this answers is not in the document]
maxaleks: Updated in 42dfc080a: the fetch mock now compares `new URL(request.url).origin` with `REWARDS_ORIGIN`. This was test-response routing in jsdom, not a production trust check, but exact origin matching also expresses the fixture's intent more accurately.

— Replied by Codex (GPT-6 Astra)

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
```diff
@@ -0,0 +1,368 @@
...
+  fetchMock.mockResponse(async(request) => {
+    const { id, method, params } = JSON.parse(await request.text()) as {
+      readonly id: number;
+      readonly method: string;
+      readonly params: Array<unknown>;
+    };
+    let result: unknown;
+    switch (method) {
+      case 'eth_accounts': result = [ ACCOUNT ]; break;
+      case 'eth_getBalance': result = request.url.startsWith(OPTIMISM_RPC) ? '0x20' : '0x10'; break;
```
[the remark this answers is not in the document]
maxaleks: Updated in 42dfc080a: the balance-response mock now uses exact URL-origin equality with `OPTIMISM_RPC` instead of `startsWith`. The affected code is test-only; the wallet tests still verify that reads and transactions reach the selected chain.

— Replied by Codex (GPT-6 Astra)

src/features/marketplace/hooks/useMarketplaceWalletActivity.spec.tsx
```diff
@@ -0,0 +1,118 @@
...
+const ENV_OVERRIDES: Array<[ string, string ]> = [
+  [ 'NEXT_PUBLIC_REWARDS_SERVICE_API_HOST', REWARDS_ORIGIN ],
+  [ 'NEXT_PUBLIC_WALLET_CONNECT_PROJECT_ID', 'test-project' ],
+  [ 'NEXT_PUBLIC_RE_CAPTCHA_APP_SITE_KEY', 'test-captcha' ],
+  [ 'NEXT_PUBLIC_MIXPANEL_PROJECT_TOKEN', 'test-mixpanel' ],
+];
+
+afterEach(() => {
+  cleanup();
+  document.cookie = 'rewards_api_token=; Max-Age=0; path=/';
```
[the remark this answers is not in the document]
maxaleks: False positive in test teardown: this assignment deletes the jsdom cookie using an empty value and `Max-Age=0`. It does not transmit a sensitive cookie or run in the application. Leaving the cleanup unchanged; adding `Secure` here would not fix a production vulnerability.

— Replied by Codex (GPT-6 Astra)

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
```diff
@@ -0,0 +1,368 @@
...
+    }
+    return JSON.stringify({ jsonrpc: '2.0', id, result });
+  });
+  await connect(walletConfig, { connector: walletConfig.connectors[0] });
+});
+
+afterEach(() => {
+  cleanup();
+  queryClient.clear();
+  document.cookie = 'rewards_api_token=; Max-Age=0; path=/';
```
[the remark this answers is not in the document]
maxaleks: False positive in test teardown: this assignment clears the jsdom cookie after each test (`rewards_api_token=; Max-Age=0; path=/`). No credential is sent over HTTP. Leaving this test cleanup unchanged.

— Replied by Codex (GPT-6 Astra)

patches/@lifi__widget-light@4.3.2.patch
maxaleks: The premise does not hold for the installed @wagmi/core 2.22.1. `config.getClient()` throws, but the exported `getClient(config, parameters)` action catches that exception and returns `undefined`; `getPublicClient` uses that action and optional chaining. The account, chain-id and switch-chain RPC branches do not require a public client.

Added a regression test in 42dfc080a: move the mock wallet to chain 137 outside the configured [1, 10] list, read `eth_chainId` and `eth_accounts`, switch to Optimism, then read its balance successfully. Keeping the patch unchanged. Falling back to the explorer's public client would risk answering a read from the wrong chain.

— Replied by Codex (GPT-6 Astra)

src/features/marketplace/pages/essential-dapp/swap/SwapWidgetIframe.tsx
maxaleks: Fixed in 42dfc080a. Loading now has a 30-second deadline. If `PageEntered` never arrives, the failed iframe is removed and an error with a “Try again” button replaces the spinner. Retry mounts a fresh iframe and starts a fresh deadline; successful readiness and unmount both clear the timer.

The regression test covers timeout, retry, ignoring the old frame's late message, and cancellation after successful readiness. Also verified the error screen and successful retry against the live hosted widget in Playwright WebKit 26.0.

— Replied by Codex (GPT-6 Astra)

src/features/marketplace/pages/essential-dapp/swap/Swap.tsx
maxaleks: Fixed in 42dfc080a: `SWAP_WIDGET_MIN_HEIGHT` in the swap's local `consts.ts` now supplies all three height values, including the Web3Boundary fallback. Kept the two small loader renderings local because one is a boundary fallback and the other overlays the iframe during its handshake.

— Replied by Codex (GPT-6 Astra)

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.ts
maxaleks: Updated in 42dfc080a. Both transaction parameters and calls-status request parameters now use module-level Valibot schemas. Tracking still falls back to an empty recipient when it cannot extract one, and the original RPC payload is passed through unchanged. The wallet and batch-activity tests pass.

— Replied by Codex (GPT-6 Astra)

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
maxaleks: Fixed in 42dfc080a: removed the static `useSwapWallet` import. Every scenario imports it dynamically, including the existing helper called inside `withEnvs`. The iframe test's `Widget` is now local to that test and captures its dynamically imported hook, so there is no module-level component bound to the pre-reset hook.

— Replied by Codex (GPT-6 Astra)

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
maxaleks: Deferred. The duplicated setup is confined to two specs with different failure boundaries: ordinary transaction/rewards-service behavior versus real wallet batch RPC and provider interception. A shared fixture would need options for those differences and make the individual scenarios harder to read. Keeping the setup local for now; there is no behavior defect here, and the affected suites pass.

— Replied by Codex (GPT-6 Astra)

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
maxaleks: Updated in 42dfc080a: removed the `getAccounts` spy, artificial readiness promise and pre-resolution empty-request assertion. The test now calls the wallet method after the chain-change event and checks the returned transaction/signature and the actual Optimism RPC URL. It preserves the cross-chain regression check without depending on wagmi's client-construction call order.

— Replied by Codex (GPT-6 Astra)

src/features/marketplace/pages/essential-dapp/swap/SwapWidgetIframe.tsx
maxaleks: Confirmed with Playwright WebKit 26.0 (not the Safari application): with an emulated light OS theme, an iframe styled `color-scheme: dark` still reports a light `prefers-color-scheme`.

I also tested the proposed `appearance: colorMode` change against the live widget.li.fi through our component. Initial `INIT { appearance: 'dark' }` renders dark correctly. However, after a light start, the guest receives `CONFIG_UPDATE { appearance: 'dark' }` and remains `<html class="light">`, including after a 10-second wait. The entered amount (0.123) and iframe document survive, but the theme does not update. The upstream [ThemeProvider](https://github.com/lifinance/widget/blob/main/packages/widget/src/providers/ThemeProvider/ThemeProvider.tsx) reads `useColorScheme` outside the MUI provider it creates, which appears to explain this.

Keeping this open for upstream follow-up and retaining the current `appearance: 'system'` implementation. Shipping the suggested change would also prevent subsequent CSS-driven theme updates in browsers where they currently work; forcing a reload would risk discarding form or active execution state. No theme workaround was committed.

— Replied by Codex (GPT-6 Astra)

docs/ENVS.md
maxaleks: Updated in 42dfc080a: the reference table now says “Swap config”. Added the removal/compatibility note for `NEXT_PUBLIC_MARKETPLACE_ESSENTIAL_DAPPS_CONFIG.swap.url` to the PR description under ENVs and added the ENVs label.

One clarification: absence of `noUnknown` allows the legacy nested property; it does not strip it. The application ignores it because the widget now uses LI.FI's hosted URL.

— Replied by Codex (GPT-6 Astra)

maxaleks:
Addressed the review in 42dfc080a.

**F3 — fixed:** removed the unused `chainId` field from `useMarketplaceWallet`'s return value, while retaining the internal chain check.

Replies are posted on every inline finding: F2, F4–F6, F8 and F10 are addressed; F1 is rejected with a regression test; F7 is deferred. F9 is confirmed in WebKit 26.0 but remains open: the live hosted widget does not apply the proposed in-place `appearance` update, so that change was not committed. The two URL-matching CodeQL comments are addressed; both cookie warnings concern test-only deletion.

Validation: 27 tests in four specs pass; ESLint (existing warnings only), cSpell, React Doctor, code-complexity and doc-link checks pass. The timeout/error/retry flow also passes in WebKit against the hosted widget. The standard local TypeScript command includes a stale, ignored `toolkit/package/node_modules_linux` directory and fails there; the same project passes with only that directory additionally excluded by a temporary config outside the repository.

Leaving the reviewer's threads open for verification.

— Replied by Codex (GPT-6 Astra)

patches/@lifi__widget-light@4.3.2.patch
tom2drum: F1 — accepted, the patch calls the `getClient` action from `@wagmi/core` 2.22.1, which wraps `config.getClient()` in try/catch and returns `undefined` for an unconfigured chain; the account, chain-id and switch-chain branches never touch the public client, and the new chain-137 regression test covers the scenario.

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/SwapWidgetIframe.tsx
tom2drum: F2 — verified

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/Swap.tsx
tom2drum: F4 — verified

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.ts
tom2drum: F5 — verified

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
tom2drum: F6 — verified

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
tom2drum: F7 — deferred

The specs do differ in the provider spy and batch state, but every item listed here is byte-identical across the two files and orthogonal to those differences, so a shared helper would not need options for them. Stays open as a nit.

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/useSwapWallet.spec.tsx
tom2drum: F8 — verified

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/SwapWidgetIframe.tsx
tom2drum: F9 — stays open, needs human

The WebKit check confirms the finding and rules out the in-place `appearance` update. The remaining decision is a product one: accept the OS theme in Safari, remount on theme change at the cost of form state, or wait for the upstream ThemeProvider fix.

— Reviewed by Claude Fable 5.1

docs/ENVS.md
tom2drum: F10 — verified. Correction accepted: without `noUnknown` yup keeps the legacy `url` key, it is ignored rather than stripped.

— Reviewed by Claude Fable 5.1

tom2drum:
F3 — verified

— Reviewed by Claude Fable 5.1

tom2drum (commented):
✅ **Cleared** — 2 nits open · 🙋 1 needs human

blocker 0 · major 0 · nit 2 · not anchorable 0

Axes: spec (skipped — no spec for this change) · standards · correctness

— Reviewed by Claude Fable 5.1

src/features/marketplace/pages/essential-dapp/swap/SwapWidgetIframe.tsx
tom2drum: Not a lot of users are affected, decided to skip for now.

PR #3712, opened 2026-09-16
Show the L2 timestamp of Super Root dispute games on the OP dispute games page

### LINKED ISSUE #3711: OP: Support Upgrade 20
Backend task: https://github.com/blockscout/blockscout/issues/14818
It is released in backend v11.3.1. Updated swagger is [here](https://github.com/blockscout/swaggers/tree/master/blockscout/11.3.1).

## Breaking change in the API after update
`/api/v2/optimism/games` API endpoint will return `"l2_block_number":null` and new field `l2_timestamp` which will be not null.

Without changes on frontend, "null" literally is displayed.

Image

tom2drum:
## Description

Resolves #3711

OP Stack Upgrade 20 introduces Super Root dispute games (game types 4/5/7/9/10). For them the backend (v11.3.1, blockscout/blockscout#14818) returns `l2_block_number: null` and a new `l2_timestamp`; legacy output-root games keep `l2_block_number` and get `l2_timestamp: null`. Both kinds coexist in one paginated list, since existing games stay in the `DisputeGameFactory` and keep resolving. Without this change the page rendered the literal `null`.

- The "L2 block #" column becomes "L2 timestamp" with a hint explaining that legacy games show a block number and Super Root games a timestamp. The cell renders the block entity when `l2_block_number` is set, otherwise `l2_timestamp` via `TimeWithTooltip` (age/absolute toggle in the header), otherwise "N/A". The "Timestamp" column is renamed to "Created".
- On mobile the label switches per row ("L2 block #" / "L2 timestamp"), so no hint is needed there; "Age"/"Resolution age" become "Created"/"Resolved".
- New `OptimisticL2DisputeGameL2Position` component holds the fallback and is shared by the table and list items.
- `@blockscout/api-types` bumped to `0.0.1-beta.d261ddc` (published from backend `dev`) for the nullable `l2_block_number` / `l2_timestamp` types. That version also drops the stale `total_gas_used` field from the `/api/v2/stats` schema, so the stats mock and stub lose it (nothing in the UI read it).
- Dispute games mock gains a Super Root row; screenshots updated.

Single-column rendering was chosen over a second "L2 timestamp" column: every row has exactly one of the two values and the table is already at its minimum width.

## Environment variables

None

## Minimum API version

Core API v11.3.1+ (older backends never send `l2_timestamp`; rows still render the block number, and the timestamp fallback never triggers).

None

## Additional information

Sepolia upgrade lands Sept 17, mainnet Sept 24, so no live instance has a Super Root game yet; the timestamp branch is covered by the mock row in the Playwright screenshot.


tom2drum (commented):
✅ **Cleared** — 1 nit open

— Reviewed by Cursor Grok 4.6

src/features/rollup/optimism/pages/dispute-games/OptimisticL2DisputeGamesTable.tsx
```diff
@@ -30,9 +31,17 @@ const OptimisticL2DisputeGamesTable = ({ items, top, isLoading, resetKey }: Prop
           <TableColumnHeader>Index</TableColumnHeader>
           <TableColumnHeader>Game type</TableColumnHeader>
           <TableColumnHeader>Address</TableColumnHeader>
-          <TableColumnHeader>{ layerLabels.current } block #</TableColumnHeader>
           <TableColumnHeader>
-            Timestamp
+            { layerLabels.current } timestamp
+            <Hint
+            // eslint-disable-next-line max-len
```
tom2drum, line 37: **💬 F1 · nit**

`// eslint-disable-next-line max-len` has no explanation. `.agents/rules/code-quality.md` — “Every `eslint-disable` comment must include an explanation.” Mechanical disables still need `--` with the reason.

Add `--` with the reason, or extract/split the Hint `label` string so the disable is unnecessary.

— Reviewed by Cursor Grok 4.6

PR #3713, opened 2026-09-17
Fix the home page RPC fallback not starting when only one widget loses its API data

tom2drum:
## Description

When the latest transactions request fails on the home page, the widget should fall back to data from the RPC node. Instead it showed "No latest transactions found." and never asked the node. When the blocks request failed too, the same message stayed until the blocks widget degraded, and only then RPC data appeared.

The cause was in `HomeRpcDataContextProvider`. `enable(false)` switched `isEnabled` and `isLoading` off from inside the `setSubscriptions` updater. React runs that updater during render, so those calls landed after the `setIsEnabled(true)` of an immediate re-subscribe and overrode it. A lone subscriber hits that unsubscribe/re-subscribe pair every time: under StrictMode in dev, and in production once the watcher resolved, because `enable` depended on `unwatch` and changed identity.

Changes:

- `isEnabled` is derived from the subscriptions list, and `enable` is stable with no side effects inside updaters.
- The block watcher is a plain effect keyed on `isEnabled` instead of a `useQuery`, so it stops and restarts with the subscribers.
- `isLoading` latches on the first block. Before, a second widget degrading later set loading back to `true` with nothing to clear it, and a chain producing empty blocks kept the skeleton forever.
- `watchBlocks` uses `emitOnBegin`, so the first block arrives immediately instead of after the 5s polling interval.
- Transaction status is read from the receipt (`receipt.data?.status`). It was read from the React Query result status, so reverted transactions showed as successful.

## Environment variables

None

## Minimum API version

None

None

## Additional information

`LatestTxsDegraded.spec.tsx` covers the txs widget alone, under StrictMode, and next to the blocks widget. All three cases failed before the fix.


src/slices/home/contexts/rpc-data-context.tsx
```diff
@@ -34,60 +35,79 @@ export function HomeRpcDataContextProvider({ children }: { children: React.React
...
+  const isEnabled = isPublicClientAvailable && subscriptions.length > 0;
+
+  const handleBlock = React.useCallback((block: OnBlockParameter<Chain | undefined, true, 'latest'>) => {
+    setTxs((prevTxs) => {
+      try {
+        const newTxs = block.transactions.map((tx) => formatTxListRpcData({ tx, receipt: null, confirmations: null, block })).filter(Boolean);
+        const nextTxs = prevTxs.length < ITEMS_LIMIT ? [ ...prevTxs, ...newTxs ].slice(0, ITEMS_LIMIT) : prevTxs;
+
+        const totalTxs = prevTxs.length + newTxs.length;
+        setTotalTxs(totalTxs);
```
tom2drum: **⚠️ F1 · major**

`setTotalTxs(totalTxs)` and `setIsError(true)` run inside the `setTxs` updater, and `totalTxs` is `prevTxs.length + newTxs.length`. `LatestTxsDegraded` feeds `totalTxs - maxNum` into the new-items badge; after the displayed list hits `ITEMS_LIMIT`, each later block overwrites the counter with `5 + <this block's txs>`, so the badge goes backwards. This is the same in-updater `setState` pattern the PR removed from `enable`, and StrictMode will double-invoke it.

Compute the new lists in `handleBlock`, then `setTotalTxs((prev) => prev + newTxs.length)` and `setIsError` outside the updaters.

— Reviewed by Cursor Grok 4.6

src/slices/home/pages/index/txs/LatestTxsDegraded.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { render, screen, cleanup } from 'vitest/lib';
+import { mockApiAndRpc } from 'vitest/utils/mockJsonRpc';
+
+import LatestBlocksDegraded from '../blocks/LatestBlocksDegraded';
+import LatestTxsDegraded from './LatestTxsDegraded';
+
+const RPC_NODE = {
+  eth_getBlockByNumber: rpcBlockWithTxsInfo,
+  eth_getTransactionReceipt: { ...rpcTxReceipt, status: '0x1' },
```
tom2drum: **💬 F2 · nit**

The mock is `{ ...rpcTxReceipt, status: '0x1' }` and no test asserts a failed transaction. The change at `status: receipt.data?.status === 'success' ? 'ok' : 'error'` replaced reading the React Query result's `status`, which is also `'success'` for a resolved query, so this spec passes with either version.

Add a reverted receipt (`status: '0x0'`) and assert the row renders as failed.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
🛑 **Blocked** — 1 major, 3 nits open

blocker 0 · major 1 · nit 3 · not anchorable 0

— Reviewed by Cursor Grok 4.6

src/slices/home/contexts/rpc-data-context.tsx
```diff
@@ -116,59 +137,39 @@ export function HomeRpcDataContextProvider({ children }: { children: React.React
...
-    return () => {
-      unwatch?.();
-    };
-  }, [ unwatch ]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps -- receiptQueries is a new array on every render; areReceiptsLoading tracks its settled state
+  }, [ areReceiptsLoading, hasBlocks ]);
 
   const enable = React.useCallback((isEnabled: boolean, id: SubscriptionId) => {
     if (!isPublicClientAvailable) {
       setIsError(true);
```
tom2drum, line 140: **💬 F3 · nit**

`if (!isPublicClientAvailable) { setIsError(true); return; }` ignores `isEnabled`, so widget unmount cleanup (`enable(false, id)`) also raises the error flag.

Only set the error when `isEnabled` is true.

— Reviewed by Cursor Grok 4.6

src/slices/home/pages/index/txs/LatestTxsDegraded.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <HomeRpcDataContextProvider>
+        <LatestTxsDegraded maxNum={ 5 }/>
+      </HomeRpcDataContextProvider>,
+    );
+
+    expect(await screen.findAllByText(RPC_TX_HASH)).not.toHaveLength(0);
+    expect(screen.queryByText(EMPTY_MESSAGE)).toBeNull();
+  });
+
+  it('shows transactions from the RPC node under React strict mode', async() => {
```
tom2drum: **💬 F4 · nit**

`it('shows transactions from the RPC node under React strict mode'` names the test after StrictMode, not the capability. `.agents/rules/tests-unit.md` flags “Test name describes HOW not WHAT”.

Name the observable behaviour; keep StrictMode only in the render wrapper.

— Reviewed by Cursor Grok 4.6

src/slices/home/contexts/rpc-data-context.tsx
tom2drum: fix (14ac03f38) — the block's txs and block data are now formatted once in `handleBlock`, with `setIsError` in its own catch; the updaters are pure, and `setTotalTxs((prev) => prev + newTxs.length)` accumulates across blocks, so the badge no longer goes backwards.

src/slices/home/pages/index/txs/LatestTxsDegraded.spec.tsx
tom2drum: fix (14ac03f38) — added "marks a reverted transaction as failed" (`status: '0x0'`, asserts the Failed label) and a success counterpart; the receipt mock now carries the rendered tx hash so the status is actually merged. Also fixed `rpcTxReceipt.status` in the shared mock: it was `'success'`, not a valid RPC value, which broke the `degradation view` Playwright test.

src/slices/home/contexts/rpc-data-context.tsx
tom2drum: deferred — nit, left as it stands. Without a public client every consumer already renders the fallback via `!isPublicClientAvailable`, so the extra error flag changes nothing; the branch also predates this PR.

src/slices/home/pages/index/txs/LatestTxsDegraded.spec.tsx
tom2drum: fix (14ac03f38) — renamed to "shows transactions from the RPC node when the widget remounts"; StrictMode stays only in the render wrapper.

src/slices/home/contexts/rpc-data-context.tsx
tom2drum: F1 — verified

src/slices/home/pages/index/txs/LatestTxsDegraded.spec.tsx
tom2drum: F2 — verified

src/slices/home/contexts/rpc-data-context.tsx
tom2drum: F3 — deferred

src/slices/home/pages/index/txs/LatestTxsDegraded.spec.tsx
tom2drum: F4 — verified

tom2drum (commented):
✅ **Cleared** — 1 nit open

— Reviewed by Cursor Grok 4.6

PR #3714, opened 2026-09-17
Fix address highlight getting stuck after a brief first hover

tom2drum:
## Description

Hovering an address highlights every entity with the same hash in the list. Sometimes the highlight stayed on after the pointer left, and only hovering that address again cleared it.

Cause: tooltips are mounted lazily since #3551, so the first hover over an entity part (icon, content, copy) replaces the DOM node under the cursor. If the pointer exits before the browser re-targets hover onto the new node, no `mouseleave` is dispatched. The pending 100ms highlight timer then still fires, and nothing removes the class, because the remembered hash is overwritten by the next hovered address. It only happens on the first hover of an entity after page load and only on a brief pass, more often when the main thread is busy, which is why it looked random.

Changes in `AddressHighlightProvider`:
- the highlight timer does nothing if the entity is no longer hovered (same `:hover` re-check the lazy tooltip already uses);
- entering an entity clears any pending timer;
- leaving removes all highlights by class instead of by the remembered hash, so a stale highlight cannot outlive the next leave.

## Environment variables

None

## Minimum API version

None

None

## Additional information

Reproduced with a Playwright script on `/txs` (Chromium): first hover over the address icon with a 5ms dwell, then move away. Before the fix 6 of 12 entities kept the highlight; after the fix 0 of 40 did (dwell 0–300ms), and the highlight still appears for dwells ≥110ms. CPU throttling (6x/20x) or a busy-wait right after the lazy mount in `useLazyActivation` makes the pre-fix bug much easier to hit by hand. Firefox and Safari were not tested.


src/slices/address/contexts/address-highlight.tsx
```diff
@@ -18,10 +18,18 @@ export function AddressHighlightProvider({ children }: AddressHighlightProviderP
...
+      typeof timeoutId.current === 'number' && window.clearTimeout(timeoutId.current);
       timeoutId.current = window.setTimeout(() => {
+        // A lazily mounted tooltip swaps the DOM node under the cursor right after the first hover,
+        // and the browser may never dispatch 'mouseleave' if the pointer exits before it re-targets.
+        if (!target.matches(':hover')) {
+          hashRef.current = null;
+          return;
+        }
         // for better performance we update DOM-nodes directly bypassing React reconciliation
         const nodes = window.document.querySelectorAll(`[data-hash="${ hashRef.current }"]`);
```
tom2drum: **💬 F1 · nit**

After this change `onMouseLeave` no longer reads `hashRef`, so its only remaining reader is the timeout selector:

```
const nodes = window.document.querySelectorAll(`[data-hash="${ hashRef.current }"]`);
```

The callback already closed over `hash` from `target.getAttribute('data-hash')`, and the new `clearTimeout` on enter cancels any earlier pending callback, so the two are always equal. The three `hashRef` writes now only keep vestigial state in sync; a later writer that forgets to clear the timer can query `[data-hash="null"]`.

Use the captured `hash` in the selector and delete `hashRef` plus its assignments.

— Reviewed by Cursor Grok 4.6

src/slices/address/contexts/address-highlight.tsx
```diff
@@ -32,13 +40,11 @@ export function AddressHighlightProvider({ children }: AddressHighlightProviderP
   }, []);
 
   const onMouseLeave = React.useCallback(() => {
-    if (hashRef.current) {
-      const nodes = window.document.querySelectorAll(`[data-hash="${ hashRef.current }"]`);
-      for (const node of nodes) {
-        node.classList.remove('address-entity_highlighted');
-      }
-      hashRef.current = null;
+    const nodes = window.document.querySelectorAll('.address-entity_highlighted');
```
tom2drum: **💬 F2 · nit**

`code-quality.md` → Magic numbers: "If the same unexplained value must stay in sync at more than one call site, one constant is the share point."

The new leave path queries `'.address-entity_highlighted'` while add (line 36) and remove (line 45) still use the bare class string. A rename now has to be made in three places, one of them dot-prefixed.

Hoist `const HIGHLIGHTED_CLASS = 'address-entity_highlighted'` and build the selector as `` `.${HIGHLIGHTED_CLASS}` ``.

— Reviewed by Cursor Grok 4.6

src/slices/address/contexts/address-highlight.tsx
```diff
@@ -18,10 +18,18 @@ export function AddressHighlightProvider({ children }: AddressHighlightProviderP
   const hashRef = React.useRef<string | null>(null);
 
   const onMouseEnter = React.useCallback((event: React.MouseEvent) => {
-    const hash = event.currentTarget.getAttribute('data-hash');
+    const target = event.currentTarget;
+    const hash = target.getAttribute('data-hash');
     if (hash) {
       hashRef.current = hash;
+      typeof timeoutId.current === 'number' && window.clearTimeout(timeoutId.current);
```
tom2drum: **💬 F3 · nit**

Possible Duplicated Code — this diff adds a third copy of `typeof timeoutId.current === 'number' && window.clearTimeout(timeoutId.current);` (enter, leave, unmount effect).

Extract one `clearHighlightTimeout` that holds the guard and the clear, and call it from all three sites.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):

blocker 0 · major 0 · nit 3 · not anchorable 0

— Reviewed by Cursor Grok 4.6

src/slices/address/contexts/address-highlight.tsx
tom2drum: fix — removed `hashRef` and its writes; the timeout selector now uses the captured `hash`. b83779568

src/slices/address/contexts/address-highlight.tsx
tom2drum: fix — hoisted `HIGHLIGHTED_CLASS`; add, remove and the leave selector all build from it. b83779568

src/slices/address/contexts/address-highlight.tsx
tom2drum: fix — extracted `clearHighlightTimeout`, called from enter, leave and the unmount cleanup. b83779568

src/slices/address/contexts/address-highlight.tsx
tom2drum: F1 — verified

src/slices/address/contexts/address-highlight.tsx
tom2drum: F2 — verified

src/slices/address/contexts/address-highlight.tsx
tom2drum: F3 — verified

tom2drum (commented):

— Reviewed by Cursor Grok 4.6

PR #3716, opened 2026-09-18
Fix React bugs surfaced by the React Doctor backlog scan

tom2drum:
## Description

Works through the React Doctor backlog (`pnpm lint:react-doctor`), rule family by rule family, keeping the fixes where the finding was a real defect and leaving or suppressing the rest with the reason recorded.

Real bugs fixed:
- `useSocketChannel`: a disabled subscriber called `channel.leave()` directly, bypassing the subscriber registry, so it killed the shared `addresses:<hash>` channel for other subscribers.
- `useCheckAddressFormat` / `useCheckDomainNameParam`: redirect state was seeded once from the param and never re-evaluated, so in-app navigation to a bech32 or ENS address skipped the redirect and queried the raw param.
- `useLoadImageViaIpfs`: object URLs for IPFS NFT media were never revoked and stayed in memory for the session.
- `SettingsScamTokens` wrote its cookie inside a state updater right before a reload; `BlockCountdownTimer` kept calling `onFinish` on every tick after reaching zero.
- `TypeFilter`, `ZetaChainAddressFilter`, `TxType` sorted prop / cached-response arrays in place, reordering memoized query params.
- `SpecifyBanner` could show a stale ad or call `onEmpty` after a superseded request; `useMarketplaceApps` fired two requests on every mount (`refetch()` in a mount effect); `CodeEditorSearch` chained two effects for one state.
- Stray `0` rendered by `count && <…>` guards. React Doctor's `rendering-conditional-render` matches on variable names and missed most of them, so it is replaced by the type-aware `@eslint-react/no-leaked-conditional-rendering`, which found 54 sites. Where `0` is a valid value (block/epoch numbers, price change, indexing ratio) the guard is now a null check so the row renders.

Also: six state-of-props cases derived during render instead of adjusted in effects; side effects moved out of state updater functions; `no-initialize-state` suppressed on the two hydration guards (`useIsMounted`, `useWalletDynamic`) since SSR/first client render must agree.

Left in the backlog (not gated) after review: the remaining `no-adjust-state-on-prop-change` hits and the `no-derived-state` / `no-derived-state-effect` hits in `useInitialList`, `useAdaptiveTabs`, `TruncateMiddle`, `multichain/context` — intentional snapshots or DOM-measurement-driven state.

## Environment variables

None

## Minimum API version

None

None

## Additional information

New dev dependency: `@eslint-react/eslint-plugin` (type-aware lint rule). Logging in/out on the marketplace page now briefly shows the skeleton instead of keeping the previous list, since the query is keyed on auth state.


src/shared/code-editor/CodeEditor.tsx
```diff
@@ -183,22 +184,24 @@ const CodeEditor = ({ data, remappings, libraries, language, mainFile, contractN
...
 
-      return prev;
-    });
-  }, [ data, index ]);
+    if (isActive) {
+      const nextActiveIndex = data.findIndex((item) => item.file_path === tabs[(tabIndex === 0 ? 1 : tabIndex - 1)]);
+      setIndex(nextActiveIndex);
+    }
+
+    setTabs(tabs.filter((item) => item !== path));
```
tom2drum, line 199: **⚠️ F1 · major**

`setTabs(tabs.filter((item) => item !== path))` reads `tabs` from the callback closure. Two closes before re-render (Alt+W via `handleTabCloseRef`, which the effect only refreshes after paint) both filter the same array; the second `setTabs` wins and the first close is dropped. The previous functional updater chained. The latest-ref effect is also one commit behind; `useSocketChannel` assigns `ref.current` during render.

Restore `setTabs(prev => …)` (and derive the next active index from `prev`). Assign `handleTabCloseRef.current = handleTabClose` during render and drop the effect.

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
⚠️ **Blocked** — 3 major open

blocker 0 · major 3 · nit 3 · not anchorable 0

Axes: spec (skipped) · standards · correctness

— Reviewed by Cursor Grok 4.6

src/slices/token/components/nft-media/useLoadImageViaIpfs.ts
```diff
@@ -12,7 +24,17 @@ export default function useLoadImageViaIpfs() {
     }
 
     const blob = await response.blob();
+    // Revoked below when replaced and in the unmount cleanup; the rule only follows a URL revoked synchronously in the same function.
+    // react-doctor-disable-next-line react-doctor/no-create-object-url-without-revoke
     const src = URL.createObjectURL(blob);
+
+    objectUrlRef.current && URL.revokeObjectURL(objectUrlRef.current);
```
tom2drum: **⚠️ F2 · major**

`objectUrlRef.current && URL.revokeObjectURL(objectUrlRef.current)` then `objectUrlRef.current = src` uses one slot for every in-flight `verifiedFetch`. `NftImage` / `NftVideo` re-invoke on `src` change with no abort: a slower first call revokes the blob URL already assigned to `img` / `poster`.

Track a generation id (or per-call URL); only revoke that call's URL, and ignore a result that is no longer current.

— Reviewed by Cursor Grok 4.6

src/shared/errors/custom/AppErrorTooManyRequests.tsx
```diff
@@ -90,23 +90,20 @@ const AppErrorTooManyRequests = ({ bypassOptions, reset }: Props) => {
...
+      setTimeLeft((prev) => Math.max((prev ?? 0) - 1, 0));
     }, SECOND);
 
     return () => {
       window.clearInterval(interval);
     };
   }, [ reset ]);
 
+  React.useEffect(() => {
+    if (timeLeft === 0) {
```
tom2drum: **⚠️ F3 · major**

`if (timeLeft === 0) { window.location.reload(); }` runs on mount. A truthy `reset` of `"0"` (or any value `< SECOND`) yields `timeLeft === 0` immediately. Reload while still 429 loops with no delay; the old interval only reloaded after a tick (`prev > 1`).

Reload only after a tick that crosses to 0, not when the initial countdown is already 0.

— Reviewed by Cursor Grok 4.6

src/slices/token/components/nft-media/useLoadImageViaIpfs.ts
```diff
@@ -12,7 +24,17 @@ export default function useLoadImageViaIpfs() {
     }
 
     const blob = await response.blob();
+    // Revoked below when replaced and in the unmount cleanup; the rule only follows a URL revoked synchronously in the same function.
+    // react-doctor-disable-next-line react-doctor/no-create-object-url-without-revoke
```
tom2drum: **💬 F4 · nit**

`tools/react-doctor/CONTEXT.md` requires the reason on the same comment as the disable. Quoted: `// react-doctor-disable-next-line react-doctor/no-create-object-url-without-revoke` — the reason sits on the previous line. Same-PR disables in `useIsMounted.ts` and `useWalletDynamic.ts` put `--` on the disable line.

Merge into one `react-doctor-disable-next-line … -- the rule only follows a URL revoked synchronously in the same function`.

— Reviewed by Cursor Grok 4.6

src/slices/contract/pages/details/code/ContractDetailsDeployedByteCode.tsx
```diff
@@ -34,46 +34,40 @@ interface Props {
...
+    if (isLoading || !config.slices.contract.decodedBytecodeEnabled || addressData.is_verified) {
+      return false;
     }
+    const decodedBytecode = hexToUtf8(bytecode.slice(0, SYMBOLS_TO_CHECK * 2 + 2));
+    return decodedBytecode.includes('scilla_version 0');
+  }, [ isLoading, bytecode, addressData.is_verified ]);
 
-    setIsLoading(isLoadingProp);
-  }, [ isLoadingProp, bytecode, addressData.is_verified ]);
+  const dataType: Array<DataType> = selectedDataType ?? (isScillaSourceCode ? [ 'UTF-8' ] : [ 'Hex' ]);
```
tom2drum: **💬 F5 · nit**

`.agents/rules/code-quality.md` Static empty defaults — a new reference is created on every render. Quoted: `const dataType: Array<DataType> = selectedDataType ?? (isScillaSourceCode ? [ 'UTF-8' ] : [ 'Hex' ]);` passed as `value={ dataType }`. Same-diff `BlobData.tsx` lifted `DEFAULT_IMAGE_FORMAT` / `DEFAULT_RAW_FORMAT`.

Module-level `DEFAULT_UTF8` / `DEFAULT_HEX` constants, same as BlobData.

— Reviewed by Cursor Grok 4.6

src/features/marketplace/pages/essential-dapp/revoke/hooks/useApprovalsQuery.ts
```diff
@@ -169,12 +169,16 @@ function buildAllowance(record: BaseAllowanceType, tokenData: schemas['Token'] |
...
+const EMPTY_KEYS: Array<string> = [];
+
 export default function useApprovalsQuery(chain: EssentialDappsChainConfig | undefined, userAddress: string, page: number) {
   const apiFetch = useApiFetch();
   const getBlockTimestamp = useGetBlockTimestamp();
   const searchErc20Allowances = useSearchErc20Allowances();
   const searchNftAllowances = useSearchNftAllowances();
-  const [ hiddenApprovalKeys, setHiddenApprovalKeys ] = useState<Array<string>>([]);
+  const scopeKey = `${ chain?.id }:${ userAddress }`;
+  const [ hiddenApprovalKeysScope, setHiddenApprovalKeysScope ] = useState<{ scopeKey: string; keys: Array<string> }>({ scopeKey, keys: EMPTY_KEYS });
```
tom2drum: **💬 F6 · nit**

`.agents/rules/typescript.md` Read-only properties — "Use `readonly` properties for object types by default." Quoted: `useState<{ scopeKey: string; keys: Array<string> }>`. New type; properties are replaced, not mutated.

`{ readonly scopeKey: string; readonly keys: Array<string> }`.

— Reviewed by Cursor Grok 4.6

src/shared/code-editor/CodeEditor.tsx
tom2drum: **reject.** The scenario needs two closes before a re-render, but Alt+W keydowns and tab-bar clicks are separate discrete events; React flushes pending passive effects before processing the next update, so `handleTabCloseRef.current` is never behind at the moment the action calls it, and `tabs` in the closure is the committed state. Deriving the next active index from `prev` would put `setIndex` back inside the `setTabs` updater — the side effect this PR removes (react-doctor `no-side-effects-in-state-updater`). Writing the ref during render is what `react-hooks/refs` flags; `useSocketChannel` doing it is legacy, not a pattern to copy.

src/slices/token/components/nft-media/useLoadImageViaIpfs.ts
tom2drum: **fix** — 856e1d606. `isUnmountedRef` replaced by `generationRef`, bumped per call and in the unmount cleanup. A call whose generation is no longer current revokes its own URL and returns `undefined` (both consumers already guard `if (src && ref.current)`); only the current call replaces and revokes `objectUrlRef`. Return type is now explicit `Promise<string | undefined>`.

src/shared/errors/custom/AppErrorTooManyRequests.tsx
tom2drum: **fix** — 856e1d606. The interval + `timeLeft === 0` effect pair is replaced by a single chained `setTimeout` effect keyed on `timeLeft`: after one tick, `timeLeft <= 1` reloads, otherwise it decrements and reschedules. A `reset` of `"0"` (or anything sub-second) now waits one tick before reloading, matching the pre-PR timing, and the reload stays outside the state updater.

src/slices/token/components/nft-media/useLoadImageViaIpfs.ts
tom2drum: **fix** — 856e1d606. Disable directive and `--` reason are on one line now.

src/slices/contract/pages/details/code/ContractDetailsDeployedByteCode.tsx
tom2drum: **fix** — 856e1d606. Module-level `DEFAULT_HEX` / `DEFAULT_UTF8` constants.

src/features/marketplace/pages/essential-dapp/revoke/hooks/useApprovalsQuery.ts
tom2drum: **fix** — 856e1d606. Extracted a `HiddenApprovalKeysScope` interface with `readonly` properties (inline it broke `max-len`).

src/shared/code-editor/CodeEditor.tsx
tom2drum: F1 — accepted, discrete Alt+W/click cannot double-close before tabs and the ref commit; a functional updater would put setIndex back inside setTabs.

src/slices/token/components/nft-media/useLoadImageViaIpfs.ts
tom2drum: F2 — verified

src/shared/errors/custom/AppErrorTooManyRequests.tsx
tom2drum: F3 — verified

src/slices/token/components/nft-media/useLoadImageViaIpfs.ts
tom2drum: F4 — verified

src/slices/contract/pages/details/code/ContractDetailsDeployedByteCode.tsx
tom2drum: F5 — verified

src/features/marketplace/pages/essential-dapp/revoke/hooks/useApprovalsQuery.ts
tom2drum: F6 — verified

tom2drum (commented):
✅ **Cleared**

Axes: spec (skipped) · standards · correctness

— Reviewed by Cursor Grok 4.6

PR #3717, opened 2026-09-18
Fix the white flash on reload for explorers with a default dark theme

tom2drum:
## Description

Fixes a white flash on page reload for explorers that set a default color theme (reported on Robinhood Chain after upgrading to v2.11.4).

next-themes ships an inline script that puts the color mode class on `<html>` before first paint. It bakes `defaultTheme` into the script text, so the script has a different hash for every default. Our CSP allowlisted only the `system` variant. Since #3691 `ColorModeProvider` takes `defaultTheme` from `NEXT_PUBLIC_COLOR_THEME_DEFAULT`, and since #3708 the proxy runs again and the policy reaches the browser. Together they blocked the script on any instance with a configured default: `<html>` stayed without the `dark` class until hydration, so the page painted white first.

- `script-src` now allowlists the script hash for all three defaults (`system`, `dark`, `light`).
- A new drift test renders the real `ColorModeProvider` for each default, hashes its inline script and asserts the hash is in the generated policy. A next-themes bump or a new provider prop that changes the script now fails the test instead of silently breaking production.

The proxy, and with it the CSP, was off from v2.9.0 to v2.11.3, so the policy was audited against a live v2.11.4 instance and the code changed in that window. Smaller gaps closed here:

- **Coinbase Wallet SDK telemetry.** Reown adds a default Coinbase connector whose SDK injects an inline telemetry script, which the CSP blocks (seen wherever the wallet stack loads eagerly: essential dapps, a contract's Read/Write tab). We now register our own connector with `telemetry: false` instead of allowlisting a third-party script hash.
- **Custom fonts.** `NEXT_PUBLIC_FONT_FAMILY_*` URLs are loaded as stylesheets, but their host was only added to `font-src`; it is now in `style-src` too.
- **Multichain reCAPTCHA.** The widget uses the per-chain site key, while the policy was gated on the root one; it now turns on when any chain has a key.
- Removed the dev-only HMR websocket entry hardcoded to port 3000 — `'self'` covers it, and the dev server runs on other ports without it.

Unrelated to the CSP, found along the way: the Next.js API proxy (dev and review environments, `NEXT_PUBLIC_USE_NEXT_JS_PROXY`) now forwards the `Referer` header. Robinhood and Arc sit behind a WAF rule that challenges core API requests without one, so every proxied request came back as a 403 challenge page.

## Environment variables

None

## Minimum API version

None

None

## Additional information

- Hashes were cross-checked against the live HTML of robinhoodchain.blockscout.com and explorer.arc.io (both render the `dark` variant, `sha256-Os32ny…`).
- Arc runs v2.11.4 with the same config and does not flash only because its responses carry no `Content-Security-Policy` header at all.
- Affected: any instance on v2.11.4+ where `NEXT_PUBLIC_COLOR_THEME_DEFAULT` is set, or `NEXT_PUBLIC_COLOR_THEMES` covers a single color mode, and the CSP header is delivered.
- Verified under the enforced policy: Usercentrics (local Robinhood preset — consent popup renders, no violations) and Dynamic (login and wallet connection checked locally).
- Known and left open: Sevio zones backfilled by Google Ad Manager have their requests to `pagead2.googlesyndication.com`, `securepubads.g.doubleclick.net` and `ep1.adtrafficquality.google` blocked by `connect-src`. Whether to allow them is a product decision.
- The Coinbase telemetry change was verified locally (eth_sepolia preset, contract Read/Write tab, CSP enforced): no violations with the fix, and the inline-script violation comes back when telemetry is re-enabled.


tom2drum (commented):
✅ **Cleared** — 0 findings

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
⚠️ **Blocked** — 1 major open · 🙋 1 needs human

blocker 0 · major 1 · nit 1 · not anchorable 0

— Reviewed by Cursor Grok 4.6

src/server/csp/policies/re-captcha.spec.ts
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: LicenseRef-Blockscout
+
+import { describe, expect, it, vi } from 'vitest';
+
+import { reCaptcha } from './re-captcha';
+
+const multichainConfig = vi.hoisted(() => vi.fn());
+
+vi.mock('src/config', () => ({
```
tom2drum: **⚠️ F1 · major**

`vi.mock('src/config', () => ({ 'default': { services: { reCaptcha: { siteKey: undefined } } } }))` pins the app-level key off, so `hasSiteKey`'s `if (config.services.reCaptcha.siteKey)` branch never runs. `.env.vitest` sets `NEXT_PUBLIC_RE_CAPTCHA_APP_SITE_KEY=xxx`. That is the real app-level path. tests-unit.md says not to mock `src/config`. Use `withEnvs` instead.

Drop the config mock. Drive app-level vs chain-only vs none through `withEnvs`, and dynamically import `./re-captcha` inside the callback.

— Reviewed by Cursor Grok 4.6

src/server/csp/policies/app.ts
```diff
@@ -36,9 +45,6 @@ export function app(isPrivateMode = false, primerScriptHashes: Array<string> = [
       KEY_WORDS.SELF,
```
tom2drum, line 45: **💬 F2 · nit · 🙋 needs human**

`KEY_WORDS.SELF` is now the only same-origin `connect-src` entry. The deleted why-comment said Safari does not treat localhost as `'self'` for the webpack HMR websocket. The commit says `'self'` covers it and the `:3000` hardcode was wrong for other ports.

Confirm Safari HMR against `'self'` on a non-3000 port. If it still fails, allow `ws://` to the actual dev origin from config, not a hardcoded port.

— Reviewed by Cursor Grok 4.6

src/server/csp/policies/re-captcha.spec.ts
tom2drum: **fix** — e4e188587. Dropped the `src/config` mock; the spec now drives the app-level key through `withEnvs` (`NEXT_PUBLIC_RE_CAPTCHA_APP_SITE_KEY` set / empty) and imports `./re-captcha` dynamically inside the callback. Added an app-level-key case with no chain keys, so the `config.services.reCaptcha.siteKey` branch is asserted — mutation score on `re-captcha.ts` is 7/7 (was 5/7, which also failed the CI gate). The `chains-config` mock stays: it reads a Node.js file / `window`, a system boundary.

src/server/csp/policies/app.ts
tom2drum: **no change needed** — verified by the developer: Safari, dev server on `http://localhost:4000/`, page loads and HMR works with `'self'` as the only same-origin `connect-src` entry. The removed entry only ever matched port 3000, so nothing is lost on other ports.

tom2drum: F2 — accepted, Safari HMR confirmed on localhost:4000 with 'self'

src/server/csp/policies/re-captcha.spec.ts
tom2drum: F1 — verified

tom2drum (commented):
✅ **Cleared** — 0 findings

— Reviewed by Cursor Grok 4.6

PR #3719, opened 2026-09-18
Track header banner link clicks in Mixpanel

ulyanas:
Track configured text-banner links as `Promo banner` with `Source: Header` and `Link`. Covers primary and middle clicks, including nested elements.

Checks: 19 tests passed, ESLint, TypeScript.


src/shell/header/HeaderAlert.tsx
```diff
@@ -7,16 +7,34 @@ import React from 'react';
...
+  if (event.button > 1 || !(event.target instanceof Element)) {
+    return;
+  }
+
+  const link = event.target.closest('a[href]');
+  if (!(link instanceof HTMLAnchorElement) || !event.currentTarget.contains(link)) {
+    return;
+  }
+
+  mixpanel.logEvent(mixpanel.EventTypes.PROMO_BANNER, { Source: 'Header', Link: link.href });
```
tom2drum, line 27: **💬 F1 · nit · 🙋 needs human**

`Link: link.href` logs the browser-resolved absolute URL, not the configured value: `href="/apps"` becomes `https://<host>/apps`, and `href="#"` becomes the current page URL. The Marketplace precedent (`IframeBanner.tsx:19`) logs the raw configured `linkUrl`, so the same `Promo banner` / `Link` property gets per-host values for relative links. The spec only covers an absolute URL, so neither behaviour is pinned.

Pick one on purpose. For the raw value use `link.getAttribute('href')`. Either way, add a relative-href row to the spec.

— Reviewed by Claude Fable 5.1

tom2drum, line 27: **💬 F2 · nit · 🙋 needs human**

Banner HTML is operator-supplied, so a link may open in the same tab; the Marketplace precedent is `<Link external>` and always opens a new one. Before the SDK is up (`useMixpanelInit` defers init to idle time plus a chunk load), `queue.track` only does `enqueue({ method: 'track', ... })` into the in-memory array (`queue.ts:86`), and a same-tab hard navigation throws that away. A fast click on a same-tab link is lost. After init the SDK's own batching survives navigation; new-tab and middle clicks are unaffected.

No cheap code fix, the queue is in-memory by design. Accept it as a known limitation, or tell operators in the `NEXT_PUBLIC_MAINTENANCE_ALERT_MESSAGE` docs to use `target="_blank"` on tracked links.

— Reviewed by Claude Fable 5.1

src/shell/header/HeaderAlert.spec.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+describe('header banner link tracking', () => {
+  it.each([
+    { target: 'link', type: 'click', button: 0, expected: 1 },
+    { target: 'span', type: 'click', button: 0, expected: 1 },
+    { target: 'span', type: 'auxclick', button: 1, expected: 1 },
+    { target: 'span', type: 'auxclick', button: 2, expected: 0 },
+    { target: 'text', type: 'click', button: 0, expected: 0 },
+    { target: 'anchor', type: 'click', button: 0, expected: 0 },
+  ])('tracks $target / $type / button $button', async({ target, type, button, expected }) => {
```
tom2drum: **💬 F3 · nit**

`'tracks $target / $type / button $button'` also titles the three `expected: 0` rows, so the output reads "tracks span / auxclick / button 2" for a case asserting nothing is tracked. `tests-unit.md`: "A good test reads like a specification". The `if (expected)` branch further down is the same split showing up in the body.

Split into two `it.each` tables (tracks / ignores), or put the expectation in the title. The conditional assertion then goes away.

— Reviewed by Claude Fable 5.1

src/shell/header/HeaderAlert.spec.tsx
```diff
@@ -0,0 +1,55 @@
...
+    await withEnvs([
+      [ 'NEXT_PUBLIC_MAINTENANCE_ALERT_MESSAGE', HTML ],
+      [ 'NEXT_PUBLIC_MIXPANEL_PROJECT_TOKEN', 'test-token' ],
+      [ 'NEXT_PUBLIC_HIDE_INDEXING_ALERT_BLOCKS', 'true' ],
+    ], async() => {
+      const { render, screen, fireEvent } = await import('vitest/lib');
+      const { 'default': HeaderAlert } = await import('./HeaderAlert');
+      const queue = await import('src/services/mixpanel/queue');
```
tom2drum: **💬 F4 · nit**

`const queue = await import('src/services/mixpanel/queue'); await queue.init('test-token', {}, () => {});` reaches past the service's public surface: `queue` is not exported from `src/services/mixpanel/index.ts` (the public init is `useInit`). `tests-unit.md` › What a good test is: "Uses public API only", "Survives internal refactors". Reshaping `queue.init` would break this header spec with no behaviour change.

Drive init through the exported `useInit`, or drop the explicit init and assert once the buffered call flushes. If neither is practical, keep as is.

— Reviewed by Claude Fable 5.1

tom2drum (commented):
✅ **Cleared** — 4 nits open · 🙋 2 needs human

blocker 0 · major 0 · nit 4 · not anchorable 0

— Reviewed by Claude Fable 5.1

src/shell/header/HeaderAlert.tsx
tom2drum: Accepted as limitations.

tom2drum: **resolve-review: decided, no production change.** Keeping the browser-resolved `link.href` on purpose: banner HTML is operator-supplied, and an absolute URL is unambiguous in Mixpanel where a raw `/apps` or `#` is not. The difference from the Marketplace source (raw configured `linkUrl`) is accepted. Behaviour is now pinned in 5f2693291: a relative-href row asserts `href="/apps"` is logged as `http://localhost:3000/apps`.

src/shell/header/HeaderAlert.spec.tsx
tom2drum: **resolve-review: fix** (promoted from nit) in 5f2693291. Split into two `it.each` tables, `tracks … as $loggedLink` and `ignores …`, with shared setup in a `clickInBanner` helper. The conditional `if (expected)` assertion is gone.

src/shell/header/HeaderAlert.spec.tsx
tom2drum: **resolve-review: fix** (promoted from nit) in 5f2693291. The spec no longer imports `src/services/mixpanel/queue`. A small harness calls the public `mixpanel.useInit()` and the test waits for it to report ready before clicking. Measured cost: none (2.23s before and after over 3 runs each; jsdom has no `requestIdleCallback`, so the hook takes its `setTimeout(0)` path).

PR #3723, opened 2026-09-21
Replace mobile list views with table views across the app

### LINKED ISSUE #3722: Replace the remaining mobile list views with table views and remove the list components
## Description

Mobile viewports currently render a dedicated list layout for most paginated views, while
desktop renders a table. The address page moved to table-on-mobile earlier and the list
implementation was kept behind an experiment as a rollback path. No regressions or user
complaints followed, so the list layout should be removed everywhere and mobile should render
the same table as desktop, scrolled horizontally.

## Details

Around 68 surfaces still render the `hideFrom="lg"` list branch next to the `hideBelow="lg"`
table branch: index pages, entity page tabs, transaction and block sub-tabs, search results,
and account pages. Each already has a table counterpart, so no new tables are needed — the
transform per page is: drop the list branch, unwrap the desktop-only box, wrap the table in
`TableContainerScrollable`, and give the table a content-derived `minW`.

Once no page renders a list, these become dead and are removed: every `*List` container and
`*ListItem` component, the shared `ListItemMobile` and `ListItemMobileGrid` primitives, and
`SocketNewItemsNotice.Mobile`.

The A/B experiment apparatus is stale and is removed in the same effort: the view-toggle hook
and button, the `showTableView` prop threading, the view-preference cookie, the feature-flag
entry, the view-switch analytics event, and the now-unused list-view icon.

## Expected Behavior

- Every affected view renders a table on mobile, with the same columns as desktop and no
  columns dropped; wide tables scroll horizontally.
- Sorting on mobile stays in the action-bar sort dropdown rather than the table header, which
  is off-screen behind horizontal scroll. The dropdown is restored on the transactions views
  where the experiment suppressed it, and added to the sortable views that never had one
  (bridged tokens, name domain history, transaction internals).
- No list layout, list component, or experiment code remains in the codebase.
- Visual tests covering list-only components are removed and mobile screenshots regenerated.

## Additional Context

The home page's latest cross-chain transactions widget is a separate case: its desktop side is a
narrow headerless table inside a card, so a horizontal scroll gesture there needs a design
decision. It is converted in its own reversible commit and reviewed on a preview deployment.

tom2drum:
Resolves #3722

Almost every paginated view in the app ships two layouts — a table for `lg` and up, and a purpose-built list of stacked label/value rows below it. Both are mounted in the DOM on every page, so each costs render work on both viewports and every column has to be maintained twice. The address page moved to table-on-mobile behind an A/B experiment that has since gone stale, and the decision to keep table view on mobile is now final. This task converts the remaining views so mobile renders the same table as desktop, scrolled horizontally, and deletes every list container, list item and mobile-list primitive along with the experiment code.

Plan: [`.agents/tasks/3722-mobile-lists-to-tables/spec.md`](https://github.com/blockscout/frontend/blob/issue-3722/.agents/tasks/3722-mobile-lists-to-tables/spec.md)

---

Spec-first draft: nothing is implemented yet. The branch receives the work ticket by ticket, one commit each; the final description is written when the PR is marked ready for review.


tom2drum (commented):
✅ **Cleared** — 0 blocker, 0 major, 0 nit open

— Reviewed by Cursor Grok 4.6

Issue #3720, opened 2026-09-21
Pilot Jev typed judgments as a fourth subagent in the agent code-review workflow


tom2drum:
## Description

Evaluate whether fast typed-judgment models (TypeSafe Jev) add value to the agent code-review workflow. The pilot adds a CLI tool that screens a change against a standards rubric and the task spec, and a fourth review subagent that confirms or drops the tool's suspects. The existing review axes, finding format and resolve flow stay unchanged.

## Details

**CLI tool** (new, under `tools/`)

- Resolves the base ref, the touched files (untracked included) and the task spec the same way the review workflow does.
- Standards grid: for each rubric rule and each matching touched file, scores the probability of a breach. Cells over a threshold get a second call to locate the offending hunk or line. The rubric is a hand-written file of 8–10 conventions that linters cannot catch. Each rule has:
  - an id;
  - the rule file it cites;
  - a file glob;
  - a question;
  - counterexamples.
- Spec grid: scores each Functional Requirement in the spec against the diff for the probability that the diff addresses it.
- Emits suspects as JSON, each with a rule or requirement id, file, line and score. It sets no severity and writes no claim or fix text.
- Test-gap and general correctness screening are out of scope. The complexity and mutation gates already cover test gaps, and the model gives no explanation to act on for correctness.

**Review workflow integration**

- The three existing axes (spec, standards, correctness) run blind, exactly as today.
- In parallel, a fourth "suspects verifier" subagent receives the tool's over-threshold cells and only confirms or drops each one. Confirmed suspects are written up in the standard finding shape (severity, location, claim, fix).
- At normalization the orchestrator de-duplicates findings across all four sources. It records each finding's origin as `axis`, `jev` or `both`, and records each dropped suspect with a reason.
- Provenance goes to the terminal report and to a sidecar file that survives task finalization. It never goes into the findings, so the resolving agent treats every finding the same.
- The integration works for both per-ticket (uncommitted) and branch/PR reviews. A missing API key or a tool failure skips the fourth subagent and does not fail the review.

## Expected Behavior

After about 10 reviews, the recorded provenance answers:

- Precision: confirmed suspects versus dropped ones.
- Unique contribution: Jev-only findings that were later accepted on resolve.
- Recall gaps: standards and spec findings from the axes with no matching suspect.
- Added latency and cost per review.

A keep-or-kill threshold is agreed before the pilot starts.

## Additional Context

[...]

PR #3727, opened 2026-09-22
Replace the per-axis word cap in the agent code review with a per-finding cap

### LINKED ISSUE #3724: Replace the per-axis word cap in the agent code review with a per-finding cap
## Description

Each axis subagent in the agent code-review workflow caps its whole report at 400 words. On a large change an axis therefore drops its lower-ranked findings to fit, and these can include `major` ones. A review should not lose findings to a length limit.

## Details

- Replace the 400-word cap on an axis report with a cap per finding of about 100 words. A finding currently averages about 80 words.
- Put no limit on the number of findings an axis returns.
- Keep the instruction to rank findings most severe first.
- The arbitration round and the normalization step stay unchanged.

## Expected Behavior

An axis returns every finding it has, and each finding stays within the per-finding cap. The number of findings reported for a large change no longer depends on a total length limit.

## Additional Context

This should land before the pilot in #3720 starts, because it changes the axis baseline that the pilot measures against.

tom2drum:
## Description

Resolves #3724

Each review axis capped its whole report at 400 words, so on a large change it dropped lower-ranked findings — `major` ones included — to fit. The cap is now per finding: 120 words of prose each, quoted code excluded, with no limit on the number of findings. Ranking most severe first stays. Arbitration and normalization are unchanged.

Decisions: the cap is a hard number rather than "about", since agents read a soft cap as no cap; no new anti-dumping rule replaces the old rationale — Out of bounds already gates baseless nits and normalize dedups.

Lands before the #3720 pilot, which measures against the axis baseline.

## Environment variables

None

## Minimum API version

None

None

## Additional information

Touches only `.agents/skills/review-changes/axes.md` and the pointer in `SKILL.md`.


.agents/skills/review-changes/axes.md
```diff
@@ -16,8 +16,9 @@ claim: <what is wrong — quote the code>
 fix: <one or two lines>
 ```
 
-Cap the report at **400 words**, which forces ranking instead of dumping. Severity here is your own read;
-the orchestrator recalibrates across axes.
+Report every finding you have, ranked most severe first; there is no limit on how many. Each finding is
+capped at **120 words** of prose — quoted code does not count. Severity here is your own read; the
```
tom2drum, line 20: **💬 F1 · nit**

possible Ambiguous absolute (prose-smells.md: `"always X" / "never Y" where an exception exists elsewhere in the repo. A reader arriving directly at the absolute will not know the exception exists.`). The new line is `Report every finding you have, ranked most severe first; there is no limit on how many.` Exceptions sit later in the same file (Out of bounds; Standards: `Skip anything the checks already cover`) and in SKILL.md's Out of bounds list. An axis that treats this paragraph as the whole contract will not see those filters.

State the exceptions on that line (every in-bounds finding; no count cap) instead of "every" / "no limit".

— Reviewed by Cursor Grok 4.6

tom2drum (commented):
✅ **Cleared** — 1 nit open

— Reviewed by Cursor Grok 4.6

.agents/skills/review-changes/axes.md
tom2drum: fix — reworded to "Report every in-bounds finding you have, ranked most severe first; the count is uncapped." so the Out of bounds exception is named on the line itself. Commit b6b5fa0a8.

